PR #50, opened 2026-04-20
ci: fix MD-only path filter with shell diff

codemonkeychris:
## Summary
- The `dorny/paths-filter` rule `['**', '!**/*.md']` did not actually skip markdown-only PRs (see #49) — each array entry is compiled as an independent matcher and OR'd, so `**` always matched and the `!`-negation never subtracted.
- Prior fix (fb3bf0d) swapped one broken picomatch pattern for another without addressing the underlying semantics.
- Replaces the filter with a shell-based diff that **defaults to running CI** and only skips when every changed file ends in `.md`. Empty diff, missing SHAs, or any non-md path all fall through to `non-md=true`, so we fail safe toward running the test suite.

## Test plan
- [ ] This PR itself touches `.github/workflows/ci.yml` (non-md) → Unit Tests **should run**.
- [ ] Follow-up: the next markdown-only PR against `main` should have Unit Tests **skipped**, and the `Detect changes` step log should show only `.md` files and `non-md=false`.
- [ ] Non-md changes on `push` to `main` continue to trigger Unit Tests (falls through `HEAD^..HEAD` branch).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,11 +24,27 @@ jobs:
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          PUSH_BEFORE: ${{ github.event.before }}
+          PUSH_SHA: ${{ github.sha }}
         run: |
+          set +e
+          ZERO_SHA="0000000000000000000000000000000000000000"
           if [ -n "$BASE_SHA" ] && [ -n "$HEAD_SHA" ]; then
-            files=$(git diff --name-only "$BASE_SHA" "$HEAD_SHA")
+            # Three-dot mirrors GitHub's PR diff (merge-base..head).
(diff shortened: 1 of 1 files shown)


codemonkeychris:
Addressed all three Copilot comments in 3af58a5:

1. **Three-dot PR diff** — switched to `git diff --name-only "$BASE_SHA...$HEAD_SHA"` to match GitHub's "Files changed" semantics (merge-base → head). Prevents base-branch noise from forcing CI runs on PRs whose head has merged base.
2. **Push range** — replaced `HEAD^..HEAD` with `github.event.before`..`github.sha`, so multi-commit pushes are fully covered. Zero-SHA (first push on a new branch) or any missing range now falls through to `non-md=true` rather than erroring or inspecting only the last commit.
3. **Fail-safe on diff failure** — any `git diff` non-zero exit now emits `non-md=true`. Every error path leans toward running CI, per the stated priority of never missing a code change.

The PR description's promise ("missing SHAs … fall through to non-md=true") now actually holds in the code.

PR #74, opened 2026-04-23
Typed editors catalog + fix programmatic-write change-event echo

codemonkeychris:
## Summary

Two related landings in one PR. The second was discovered while building the first.

### 1. Typed editor catalog + specialized editors demo (commit `c4c6a43`)
- `Editors` catalog of editor factories (Text, CheckBox/Toggle, Number, Date/Time, Uri, Combo/EnumCombo, Color/ColorCompact) and matching `CellRenderers`.
- `TypedColumns` batteries-included column factories that pair editor + renderer (NumberColumn, ToggleSwitchColumn, DateColumn, HyperlinkColumn, ColorColumn, etc.).
- `SpecializedEditorsDemo` sample — three sections (auto-discovery DataGrid, explicit-typed-columns DataGrid, PropertyGrid) over one shared INPC-aware Gizmo source; edits in any surface propagate via INPC.
- DataGrid cell `OnTapped` gets a commit-before-begin guard so an in-flight edit is not clobbered when the user clicks a new cell.
- Selftest: `SpecializedEditorsTests` mount coverage for every editor factory against real WinUI controls.
- Unit test: `TypedEditorsTests`.

### 2. Fix: suppress programmatic-write change-event echo (commit `b6d155b`)

**Bug.** When a Reactor `Update<Control>` handler wrote a value-bearing DP (`cp.Color = n.Color`, `nb.Value = n.Value`, `ts.IsOn = n.IsOn`, …), the control raised its change event. If the user had wired an `onChange` callback, that callback got invoked with the value Reactor just wrote — indistinguishable from a real user edit. For components bound to an external selection (e.g. a `PropertyGrid` watching the currently-selected DataGrid row), the echo silently wrote the *new* selection's value back into the *previous* selection's state: a cross-row byte-for-byte value swap on selection change. Every scalar property of the previous selection was overwritten, not just the one that was visible. See [`docs/research/investigations/change-event-echo.md`](./docs/research/investigations/change-event-echo.md) for the full incident report.

**Fix.** A `ChangeEchoSuppressor` (per-control `ConditionalWeakTable`-backed counter):
- `BeginSuppress(control)` increments before a programmatic write that will raise a change event.
- Every change-event handler calls `ShouldSuppress(control)` as its first line and returns early if it consumed a token.

Every `Update<Control>` now: (a) calls `SetElementTag` first (defense-in-depth), (b) equality-guards the write, (c) `BeginSuppress`es before the assign. Every `Mount<Control>` gates its user-callback dispatch behind `ShouldSuppress`. Pooled controls (`TextField`, `ToggleSwitch`) also re-tag before re-rent so the retained handler sees the new element.

Covers `ColorPicker`, `NumberBox`, `ToggleSwitch`, `CheckBox`, `RadioButton`, `RadioButtons`, `ComboBox`, `Slider`, `RatingControl`, `DatePicker`, `CalendarDatePicker`, `TimePicker`, `ToggleSplitButton`, `PasswordBox`, `TextField`, `AutoSuggestBox`.

**Regression coverage.** `EchoSuppressionFixtures` — 12 fixtures × 48 TAP assertions. Each editor asserts the three invariants: mount does not fire `onChange`, programmatic re-render does not echo, simulated user edit still fires `onChange`.

## Follow-ups (noted in the investigation doc, out of scope here)
- `.OnPointerPressed` / `.OnTapped` stack handlers across re-renders — a single click fires the enqueued handler 4–10 times. Idempotent today so not visibly broken, but wasted work.
- `ObservableListDataSource.OnItemPropertyChanged` emits a `DataChanged` per INPC event with no coalescing. One real edit currently triggers 5+ grid re-renders.
- Consider extracting a `SetSuppressed(control, current, target, assign)` helper to name the pattern, plus a Roslyn-based unit lint that fails any `Update<X>` assigning a control property before `SetElementTag`.

## Test plan
- [x] `dotnet test tests/Reactor.Tests` — 6562 pass, 0 fail
- [x] `dotnet run --project tests/Reactor.AppTests.Host -- --self-test` — 1996 pass, 0 fail (including 48 new echo-suppression asserts and the existing `ControlUpdate_InputControls`, `ControlUpdate_DateTimePicker`, DataGrid edit fixtures that exercise continuous-value update paths)
- [x] Manual: reproduce the original Row 0 → Row 2 AccentColor swap in the Specialized Editors tab — confirmed fixed.

### DIFF
--- a/docs/research/investigations/change-event-echo.md
+++ b/docs/research/investigations/change-event-echo.md
@@ -1,0 +1,109 @@
+# Change-event echo — incident report and framework-level mitigations
+
+**Date:** 2026-04-23
+**Area:** `Reconciler.Update*` / `Reconciler.Mount*` for value-bearing WinUI controls.
+**Severity:** Silent cross-component state corruption.
+
+## What happened
+
+The DataGrid's `Specialized Editors` demo exhibited an "exact value swap" bug: clicking Row A's Accent cell and then Row B's Accent cell caused Row A's `AccentColor` to be overwritten with Row B's color byte-for-byte. The bug reproduced every time.
+
+Root cause was *not* in the DataGrid edit pipeline. It was in the **`PropertyGrid` bound to the selected row**, which renders an `Editors.Color()` (a full WinUI `ColorPicker`). When the selection shifted from Row A to Row B:
+
+1. `UpdatePropertyGrid` → `UpdateColorPicker` ran with `n.Color = rowB.AccentColor`.
+2. `UpdateColorPicker` wrote `cp.Color = n.Color` **before** calling `SetElementTag(cp, n)`. The ColorPicker's tag at that moment was still `rowAColorPickerElement` (the previous render).
+3. `ColorPicker.ColorChanged` fired synchronously (in WinAppSDK 2.0-preview for this property). The handler read the stale tag, invoked *Row A's* `OnColorChanged` with `rowB.AccentColor`, and the PropertyGrid wrote `rowB.AccentColor` into `rowA.AccentColor`.
+
+Because the echo fires on every programmatic write where the tag hasn't been swapped yet, the same issue silently corrupted `Quantity`, `UnitPrice`, `OrderDate`, and `Duration` on row A too — we only noticed the color because it was visually obvious.
+
+## The class of bug
+
+Every Reactor updater for a WinUI control whose change event fires from a programmatic property write is vulnerable:
+
+| Control | Change event fired by programmatic set |
+|---|---|
+| `ColorPicker.Color`     | `ColorChanged` — confirmed sync (or near-sync) |
+| `NumberBox.Value`       | `ValueChanged` — confirmed |
+| `ToggleSwitch.IsOn`     | `Toggled` — confirmed |
+| `Slider.Value`          | `ValueChanged` — confirmed |
+| `DatePicker.Date`       | `DateChanged` — confirmed |
+| `CalendarDatePicker.Date` | `DateChanged` — confirmed |
+| `TimePicker.Time`       | `TimeChanged` — confirmed |
+| `CheckBox.IsChecked`    | `Checked` / `Unchecked` / `Indeterminate` — confirmed |
+| `RadioButton.IsChecked` | `Checked` / `Unchecked` — confirmed |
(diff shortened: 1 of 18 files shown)


codemonkeychris:
Rebased onto latest `main` (took #73 which added `SelectionEvt_*` / `ValueEvt_*` fixtures; merge was additive in the registry), and addressed the Copilot review feedback:

### Critical (correctness)
- **`UpdateSlider` / `UpdateNumberBox` Min/Max coercion echo** — real correctness gap in the initial fix. Min/Max writes can themselves coerce Value in-range and raise `ValueChanged` before the explicit Value assignment, so the previous single-token suppression leaked those echoes through. Now: `BeginSuppress` is issued before each of Min, Max, Value — one token per write that will actually fire. Added regression fixtures `EchoSuppress_SliderMinMaxCoercion` and `EchoSuppress_NumberBoxMinMaxCoercion` that shift the range so the existing Value is forced out-of-bounds; both pass.

### Quality / doc
- `Editors.cs`: dropped the unused `ToBrushHex` / `ContrastForeground` helpers.
- `ReflectionTypeMetadataProvider.ResolveEditorFromDataAnnotations` xml-doc: removed the `DataType.MultilineText` claim the code doesn't back; surfaced the Uri-typed Url branch that *is* implemented.
- `TypeRegistry.GetCellRenderer`: the fallback factory returned a fresh delegate per call; now the first resolution is inserted into `_cellRenderers` so per-cell render doesn't re-allocate.
- `SpecializedEditorsTests.ColorCompactEditorMounts`: matches the hex `TextBox` by its expected Crimson text instead of any `TextBox`, so it's robust against harness chrome.
- `TypedEditorsTests`: renamed / fixed two tests whose names overclaimed — `Number_Editor_For_Int_Hands_Int_To_Callback` didn't actually invoke the callback, and the DateColumn "DateOnly" test used a `DateTime` property.

### Test status
- Selftest: **2051 / 2051 pass** (68 echo-suppression asserts, including 6 new for Min/Max coercion).
- Unit tests: **6561 pass**, plus 1 flaky dispatcher test (`UseResourceTests.Null_Dispatcher_Runs_Continuation_Inline`) that passes in isolation and is unrelated to this PR.

PR #79, opened 2026-04-23
Reconciler: in-place Update for 14 controls that fell through to Mount

codemonkeychris:
## Summary

14 controls' Update handlers in `Reconciler.Update.cs` were `=> Mount(newEl, requestRerender)` — rebuilding the entire WinUI control on every reconcile. The visible symptom was ComboBox icon flicker in the TestApp TitleBar when moving the Counter demo's slider (and, more subtly, lost selection / scroll / pane / focus state on any re-render path that touched these controls).

Affected elements: `ComboBoxElement`, `RadioButtonsElement`, `SplitViewElement`, `TabViewElement`, `PivotElement`, `RelativePanelElement`, `SemanticZoomElement`, `ListBoxElement`, `SelectorBarElement`, `PopupElement`, `RefreshContainerElement`, `CommandBarFlyoutElement`, `SwipeControlElement`, `ParallaxViewElement`.

Each now has a proper `Update*` handler that diffs fields, reconciles children positionally (via `ChildReconciler` where applicable), and writes `SelectedIndex` / `SelectedItem` through `ChangeEchoSuppressor` so programmatic writes don't echo into user callbacks. `Mount` handlers for `TabView`, `Pivot`, `ListBox`, `SelectorBar` gain `ShouldSuppress` guards to pair with the new in-place selection writes.

## Regression coverage

`IdentityPreservationFixtures.cs` — 8 fixtures, 17 assertions. Each mounts the control under test next to a driver button, clicks the driver to trigger an unrelated re-render, and verifies the WinUI control is the *same instance* across the reconcile. One fixture mirrors the TestApp TitleBar pattern (ComboBox with element items) and also asserts the child-item controls survive.

## Scope

This does **not** touch the `SelfTriggered` propagation behavior (every descendant state change marks all ancestors as dirty, bypassing the memo check). That's a separate, larger design question — this PR fixes the immediate user-visible flicker and the state-loss class of bug, which were made far worse by the propagation but are independently worth fixing.

## Test plan

- [x] Build: solution builds clean.
- [x] Unit tests: 6562 pass, 0 fail.
- [x] Selftest: 584 fixtures pass, 0 fail (up from 576 with the new identity fixtures).
- [x] Manually verified TestApp Counter slider no longer flickers the TitleBar ComboBox icon.
- [ ] **E2E tests: 53/57 pass.** Three Appium tests timed out on WaitForText / NavigateToFixture: `Interactive_DataGrid_ClickEditTabCommit`, `Devtools_Menu_Toggle_Flows_Through_To_Subscribers`, `Devtools_Menu_Toggle_Is_Reversible`. These look like driver-level timeouts (session ref-count reuse, 5s implicit waits) rather than reconciler regressions — none of them touch the 14 changed controls — but flagging for a human spot-check before merge.

### DIFF
--- a/src/Reactor/Core/Reconciler.Mount.cs
+++ b/src/Reactor/Core/Reconciler.Mount.cs
@@ -1108,6 +1108,7 @@ public sealed partial class Reconciler
         tv.SelectionChanged += (s, _) =>
         {
             var t = (WinUI.TabView)s!;
+            if (ChangeEchoSuppressor.ShouldSuppress(t)) return;
             (GetElementTag(t) as TabViewElement)?.OnSelectionChanged?.Invoke(t.SelectedIndex);
         };
         tv.TabCloseRequested += (s, args) =>
@@ -1148,6 +1149,7 @@ public sealed partial class Reconciler
         pivot.SelectionChanged += (s, _) =>
         {
             var p = (WinUI.Pivot)s!;
+            if (ChangeEchoSuppressor.ShouldSuppress(p)) return;
             (GetElementTag(p) as PivotElement)?.OnSelectionChanged?.Invoke(p.SelectedIndex);
         };
         ApplySetters(pvt.Setters, pivot);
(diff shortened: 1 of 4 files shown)


codemonkeychris:
Closing in favor of a smaller, rebased PR — PR #76 (cd90291) landed independently with equivalent reconciler Update handlers for the same 14 controls. Reopening with just the IdentityPreservationFixtures regression coverage on top of main.

PR #89, opened 2026-04-24
reconciler: fix duplicate event subscriptions from aliased RCWs (#86)

### LINKED ISSUE #86: Reconciler: multiple C# RCWs for one native element accumulate duplicate event subscriptions
## Summary

The reconciler's per-FE event state is keyed in a `ConditionalWeakTable<FrameworkElement, EventHandlerState>` by C# reference. When WinUI hands the reconciler a **new C# RCW (Runtime Callable Wrapper) for the same underlying native element**, the table treats it as a brand-new FE, allocates a new `EventHandlerState`, and calls `fe.Tapped += state.TappedTrampoline;` (and analogous `+=` for the other routed events). The native element retains the *previous* RCW's subscription as well, so one native element ends up with N trampolines — one per RCW ever created for it. A single user click then invokes the handler N times.

`EnsureTappedSubscribed`'s `if (state.TappedTrampoline is null)` guard only protects against re-attach within the *same* state object; it can't detect the cross-RCW aliasing.

## Impact (observed)

Surface-level: DataGrid `Interactive_DataGrid_ClickEditTabCommit` — clicking Bob's cell after typing into Alice's editor fires the cell's `OnTapped` 2–3 times. First fire: commit Alice (correct) + begin Bob. Second fire: `IsEditing=true`, so it commits Bob's *unchanged* value, overwriting `EditStatus` with `"2:Bob,Jones"` instead of the expected `"1:Alicia,Smith"`.

Latent: any code that uses `OnTapped` / `OnPointerPressed` / `OnLostFocus` etc. on an element that may go through re-renders where the reconciler ends up with a fresh RCW for the same native element. Not currently observed in other tests, but the mechanism is general.

## Evidence

Instrumented `EnsureTappedSubscribed` to log `fe` identity via both `RuntimeHelpers.GetHashCode` and `Marshal.GetIUnknownForObject`, and tagged each trampoline with an incrementing `tramId`. Clicking Bob produced:

\`\`\`
[Tapped.trampoline] tramId=5 sender_rid=14732956 sender_iunk=0x1C184E8FD18 ... stateHash=17123155
[Tapped.trampoline] tramId=9 sender_rid=14732956 sender_iunk=0x1C184E8FD18 ... stateHash=40956631
\`\`\`

- Two distinct trampolines (different `tramId`, different `EventHandlerState`) fire on the same click.
- Same `sender_iunk=0x1C184E8FD18` — both trampolines are attached to the *same native element*.
- The two states were created for different C# wrappers at different `State.NEW` log points — one for a wrapper with `rid=815388`, another for `rid=14732956`. Both subscribed via `fe.Tapped += trampoline` on what turned out to be the same native control.

`ElementPool.TryRent` / `Return` are **not** invoked during this scenario (verified with log instrumentation) — so pooling is not the mechanism. The reconciler's Mount path calls `new TextBlock()` (Reconciler.Mount.cs:243), which per WinRT semantics should allocate a fresh native element. The aliasing appears to arise elsewhere — likely from how WinUI re-exposes the same native element through a new RCW when the C# reference is reacquired (e.g., via tree traversal or a WinRT getter). I did not fully pin down the specific code path that yields the aliased wrapper.

## Why the prior pool-detach fix (from earlier investigation) didn't work

A previous attempt added `Reconciler.ClearPooledEventHandlers` and called it from `ElementPool.CleanElement` on Return. Targeted the wrong lifecycle — the pool isn't involved here, so that code never ran. Reverted.

## Workaround in place

\`src/Reactor/Controls/DataGrid/DataGridComponent.cs\` — cell `OnTapped` is now idempotent: if `state.IsEditing && EditingRowKey == capturedRowKey && EditingColumnName == capturedColName`, the handler no-ops. Repeated fires become harmless; the first fire's commit+begin sticks. See PR #[TBD] / commit ce37958's successor.

This is a symptom-level patch — the reconciler still accumulates duplicate subscriptions. Any other component using routed-event modifiers on an element that ends up with multiple RCWs remains at risk.

## Proposed fix (deeper)

Options considered:

1. **Attached property on the native side** (preferred). Store a sentinel (e.g., a `bool` via `DependencyProperty.RegisterAttached`) on the native element itself to mark \"Reactor has already attached trampolines here\". Because the value lives on the native object, it's shared across all RCWs. On `ApplyEventHandlers`, check the sentinel first; if already set, retrieve the existing `EventHandlerState` (stored alongside, e.g., via `fe.GetValue(stateProp)`) and update its `Current*` fields rather than allocating a new state + attaching new trampolines.

2. **Use `fe.Tag` as the state carrier.** Simpler but `Tag` is often claimed by apps — would need a wrapper type that contains both app `Tag` and reactor state.

3. **Detach-on-unmount.** When the reconciler knows an FE is being unmounted (not pool-returned — actually unmounted from the live tree), explicitly `fe.Tapped -= state.TappedTrampoline`. Doesn't fully solve the problem if re-mount produces a fresh RCW that the old RCW's state still references via closure, but would reduce accumulation.

Option 1 is the cleanest and most bulletproof. Would affect `GetOrCreateEventState` + all `Ensure*Subscribed` helpers in \`Reconciler.cs\`.

## Reproducer

- Branch: main, commit ce37958 or later
- Test: \`dotnet test tests/Reactor.AppTests -p:Platform=x64 --filter \"FullyQualifiedName~Interactive_DataGrid_ClickEditTabCommit\"\`
- Failed before idempotency workaround; passes after. The workaround doesn't eliminate the duplicate fires — it just makes the DataGrid tolerant of them.

## Related files

- \`src/Reactor/Core/Reconciler.cs\` — `EnsureTappedSubscribed` (~line 2539), `GetOrCreateEventState`, `_eventStates` ConditionalWeakTable
- \`src/Reactor/Core/Reconciler.Mount.cs:241\` — `MountText` (and analogous for other types)
- \`src/Reactor/Core/ElementPool.cs\` — not the cause, but historically suspected; leave as-is

## Investigation artifact

Full diagnosis writeup at \`C:\temp\e2e-test-investigation.md\` on the investigation machine (local file, not committed). Summary: 2–3 trampolines fire per Bob click; same IUnknown across trampolines; distinct `EventHandlerState` objects attached from distinct C# wrappers. Pool not involved.

codemonkeychris:
## Summary

- Fixes #86. Root cause was that per-FE `EventHandlerState` was keyed by C# reference in a `ConditionalWeakTable`. When WinUI handed the reconciler a fresh RCW for the same native element (re-read via `Panel.Children[i]`, `VisualTreeHelper`, etc.), the table treated it as a new key, allocated a new state, and attached another trampoline via `fe.Tapped += ...`. The native element accumulated N trampolines — one per RCW — and a single click fired the user handler N times.
- Moves per-element reactor state onto the native `DependencyObject` via an attached DP (`ReactorAttached.StateProperty`). DP values live on the native DO, so every RCW observes the same `ReactorState` — one `EventHandlerState`, one trampoline set, no accumulation.
- Same DP carries both the reactor `Element` pointer (replacing the `fe.Tag` squat) and the `EventHandlerState`. `Tag` is returned to application code. `ConditionalWeakTable` removed.
- Reverts the symptom-level idempotency guard from #87 on `DataGridComponent.OnTapped`; the DataGrid no longer needs it.

## Why two patterns were consolidated

Both the `Element`-pointer tracking (previously on `Tag`) and the event-state tracking (previously in a weak table) had the same underlying hazard — identity keyed on C# reference rather than on the native DO. Bundling them in a single `ReactorState` on one DP fixes both aliasing paths at once and drops a data structure. `Tag` becomes available to app code.

## Pool lifecycle

`ClearElementTag` nulls out `state.Element` but preserves `state.Events` so already-attached trampolines survive pool return/rent. The trampoline dispatches through `state.Current*`, which `Ensure*Subscribed` refreshes on the next mount/update. Re-attaching subscriptions on every rent would re-introduce the accumulation bug.

## Perf (10 arm64 Release runs, `--percent 50 --duration 7`)

| Metric | Baseline (main) | This PR | Verdict |
|---|---|---|---|
| Avg FPS | 8.71 ± 0.11 | 8.63 ± 0.12 | within CI |
| Avg Working Set (MB) | 483.4 ± 1.5 | 478.7 ± 4.0 | marginal improvement (removing CWT) |
| Peak Working Set (MB) | 505.0 ± 1.9 | 503.1 ± 4.4 | within CI |

No regression. The small memory delta is consistent with removing one `ConditionalWeakTable` entry per callback-bearing FE.

## Validated

- [x] 6562/6562 unit tests pass (`Reactor.Tests`)
- [x] 590/590 self-tests pass (`Reactor.SelfTests`, in-process WinUI harness)
- [x] `Interactive_DataGrid_ClickEditTabCommit` passes **with the #87 idempotency workaround reverted** — root cause is fixed, symptom guard no longer needed
- [x] Build clean across the solution (x64 Debug, arm64 Release)

## Out of scope

The remaining `Tag` squats in `Reconciler.Mount.cs` (~1836–1911) and `Reconciler.Update.cs` (~2723–2836) where `AppBarButton.Tag = cmd` / `MenuFlyoutItem.Tag = data` stash reactor data structs consumed by native event closures. Different pattern from `Element` tracking; can move in a follow-up if we want `Tag` fully returned to the app.

## Test plan

- [ ] Run `dotnet test tests/Reactor.Tests` on CI
- [ ] Run `dotnet test tests/Reactor.SelfTests` on CI
- [ ] Run `dotnet test tests/Reactor.AppTests --filter "FullyQualifiedName~Interactive_DataGrid_ClickEditTabCommit"` on CI
- [ ] Spot-check other event-heavy surfaces: Counter samples, DataGrid row-edit mode, MenuFlyout click paths

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -215,3 +215,6 @@ samples/apps/reactorfiles/Native/reactorfs/target/
 # Cobertura reports are regenerated on demand and should never be committed.
 coverage/
 *.cobertura.xml
+
+# Stress-perf benchmark run artifacts (local measurement data).
+tests/stress_perf/bench-results/
--- a/src/Reactor/Animation/AnimationHelper.cs
+++ b/src/Reactor/Animation/AnimationHelper.cs
@@ -22,7 +22,7 @@ internal static class AnimationHelper
         // Resolve animation curve: prefer ambient WithAnimation scope,
         // fall back to element's .Animate() config curve.
         var curve = AnimationScope.HasScope ? AnimationScope.Current : null;
-        curve ??= (element is FrameworkElement fe && fe.Tag is Element tagEl)
+        curve ??= (element is FrameworkElement fe && Reconciler.GetElementTag(fe) is Element tagEl)
             ? tagEl.AnimationConfig?.Curve : null;
 
         if (curve is null)
@@ -240,7 +240,7 @@ internal static class AnimationHelper
     internal static void SetOrAnimateVector3(UIElement element, string property, Vector3 value)
     {
         var curve = AnimationScope.HasScope ? AnimationScope.Current : null;
-        curve ??= (element is FrameworkElement fe && fe.Tag is Element tagEl)
+        curve ??= (element is FrameworkElement fe && Reconciler.GetElementTag(fe) is Element tagEl)
             ? tagEl.AnimationConfig?.Curve : null;
 
         if (curve is null)
--- a/src/Reactor/Controls/DataGrid/DataGridComponent.cs
+++ b/src/Reactor/Controls/DataGrid/DataGridComponent.cs
@@ -614,25 +614,10 @@ public class DataGridComponent<T> : Component<DataGridElement<T>>
             {
                 var capturedRow = index;
                 var capturedCol = c;
-                var capturedRowKey = rowKey;
-                var capturedColName = col.Name;
                 cell = cell.OnTapped((sender, e) =>
                 {
                     Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread()?.TryEnqueue(() =>
                     {
-                        // Idempotency guard: if we're already editing exactly this cell,
-                        // skip commit+begin. The underlying native element may have
-                        // accumulated multiple Tapped subscriptions across re-renders
-                        // (multiple C# wrappers of the same native control), causing one
-                        // click to invoke this handler more than once; without this guard
-                        // the second invocation would commit the freshly-begun edit with
-                        // the unchanged value, clobbering the previous row's commit.
-                        if (state.IsEditing
-                            && state.EditingRowKey?.Equals(capturedRowKey) == true
-                            && state.EditingColumnName == capturedColName)
-                        {
-                            return;
-                        }
                         // Commit any in-flight edit BEFORE starting a new one.
                         // BeginEdit unconditionally overwrites _editingValue with the
                         // new cell's current value, which would destroy the in-flight
--- a/src/Reactor/Controls/DataGrid/ResizeGrip.cs
+++ b/src/Reactor/Controls/DataGrid/ResizeGrip.cs
@@ -55,7 +55,7 @@ internal static class ResizeGripRegistration
                     var child = r.Mount(el.Child, rerender);
                     if (child is not null) panel.Children.Add(child);
                 }
-                panel.Tag = el;
+                Reconciler.SetElementTag(panel, el);
                 return panel;
             },
             update: (r, oldEl, newEl, panel, rerender) =>
@@ -83,7 +83,7 @@ internal static class ResizeGripRegistration
                     panel.Children.Clear();
                 }
 
-                panel.Tag = newEl;
+                Reconciler.SetElementTag(panel, newEl);
                 return null;
             });
     }
--- a/src/Reactor/Core/ChildReconciler.cs
+++ b/src/Reactor/Core/ChildReconciler.cs
@@ -77,7 +77,7 @@ internal static class ChildReconciler
                 // the initial count). For callback-free elements we still avoid
                 // the children.Get COM call.
                 if (newEl.HasCallbacks && children.Get(i) is FrameworkElement fe)
-                    fe.Tag = newEl;
+                    Reconciler.SetElementTag(fe, newEl);
                 continue;
             }
 
@@ -280,7 +280,7 @@ internal static class ChildReconciler
         for (int i = prefixLen; i < searchEnd && i < children.Count; i++)
         {
             var child = children.Get(i);
-            if (child is FrameworkElement fe && fe.Tag is Element tagElement)
+            if (child is FrameworkElement fe && Reconciler.GetElementTag(fe) is Element tagElement)
             {
                 var key = GetKey(tagElement, i);
                 keyToIndex.TryAdd(key, i);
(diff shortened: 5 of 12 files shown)


codemonkeychris:
Addressed Copilot's review in 6ac5357.

**Comments 1 & 2 (ClearElementTag leaking for permanent unmount):**
Split into two helpers to make intent explicit at call sites:

- `ClearElementTag(fe)` — unchanged (nulls Element, preserves Events). Used by `ElementPool.CleanElement` and `ListView` item recycle where the FE stays reactor-managed.
- `DetachReactorState(fe)` — nulls Element, clears all `EventHandlerState.Current*` delegates via a new `ClearCurrentHandlers()` helper, drops `state.Events`. Any orphan trampoline on the native event source becomes a no-op. Used by `XamlHost` / `XamlPage` unmount — the paths Copilot specifically flagged.

Trampoline delegate fields themselves (`TappedTrampoline` etc.) are left in place; detaching them would require going back through each typed WinUI event (`fe.Tapped -= ...`) and that's not safe from this helper without knowing what was actually subscribed. Clearing `Current*` is sufficient because the trampoline pattern is `(s,e) => state.CurrentX?.Invoke(s,e)` — a null Current* means no user-visible side effect.

**Comment 3 (bench script):** added the `Test-Path` guard with an actionable build-first error message.

Validated: unit tests 6562/6562, self-tests 590/590, `Interactive_DataGrid_ClickEditTabCommit` E2E passes with the #87 idempotency workaround still reverted.

PR #92, opened 2026-04-25
feat(overlay): layout-cost overlay (spec 032)

codemonkeychris:
## Summary

- Implements spec 032 — a per-Component layout-cost overlay backed by `Microsoft-Windows-XAML` ETW events.
- Renders a green outline around each Reactor `Component` subtree plus a 64×20 sparkline badge tracking measure+arrange ms over the last ~6 s.
- Unified `OverlayHostWiring` so the reconcile-highlight overlay (#88) and the layout-cost overlay paint into a single shared wrapper Canvas. Adding a third overlay is a sub-renderer + a flag check; no per-feature wrapper dance.
- Attribution coordinate composition through the `EventPairing` stack — lifted WinUI's ETW Arrange events emit parent-relative `VisualOffsetX/Y`, not root-relative as the spec assumed. Without the composition, deep events all attributed to whichever Component sat at the screen top-left.
- Flag-gated zero-cost-when-off: ETW session stops on flag-off, idle ticker pauses, sub-overlays dispose. Highlight list `.Add` and LC mount-event work also flag-gated (pre-existing leak when the highlight flag was toggled off).
- Spec updated with an "Implementation findings" section (what the original spec got wrong about lifted WinUI, design departures, file layout) and a "Future ETW improvements" section ranking paths to better attribution.

Behind `ReactorFeatureFlags.ShowLayoutCost`, default `false`. Phases 4 (Heatmap / Inspect modes) and 5 (detailed-view pin, telemetry, perf gate) deferred per spec.

## Test plan

- [x] `tests/Reactor.Tests` unit tests — pure logic for color ramps, meter math, anchor placement, surface-through, attribution, rollup EMA, spatial index, pointer map, event pairing (incl. new `NestedArrange_PairedRect_IsRootRelative` for root-origin composition), ring buffer, consumer.
- [ ] `tests/Reactor.AppTests.Host` selftest fixtures — new `LayoutCostOverlayTests` (back-fill, mount-while-on, unmount-while-on, toggle off→on, bounds refresh).
- [ ] Manual smoke in `samples/Reactor.TestApp` — toggle overlay on/off, switch tabs while on, scrub Rows slider on the Layout Cost demo, observe outlines + sparklines update; confirm no exceptions on flag-off → wrapper-teardown.
- [ ] `dotnet format` clean.

## Notes for reviewers

The `Implementation findings` section in `docs/specs/032-layout-cost-overlay-design.md` (added in this PR) is the best entry point. It captures everything the original spec got wrong about lifted WinUI's actual ETW emission, why we ship without the native-pointer interop, the sparkline-vs-bar-meter trade, and the unified-wiring rework. If this feature got ripped out and someone needed to bring it back, that section is the recovery path.

Native-pointer interop (the path to deterministic attribution without coord composition) is documented as Future ETW Improvement Option A — out of scope for v1, would need a separate spec.


codemonkeychris:
Addressed all eight comments from the Copilot review pass in commit d06a7ae. Per comment:

**[3142305402] CloseOrphanSessions cross-process kill** — Now parses the pid suffix and only stops sessions whose owning process is dead (`Process.GetProcessById` + `HasExited`). Live sibling Reactor processes keep their sessions. Access-denied on `HasExited` is treated as "alive" to fail safe.

**[3142305441] s_liveConsumers unbounded growth** — Per-instance `_registeredInLiveConsumers` flag; each consumer adds exactly one weak ref regardless of Start/Stop cycles. We also prune dead weak refs while holding the lock.

**[3142305407] TimestampTicks comment inaccuracy** — Updated the `RawLayoutEvent.TimestampTicks` comment to clarify it's `TraceEvent.TimeStamp.Ticks` (DateTime ticks, derived from QPC but not raw QPC). Arithmetic is interval-only so DateTime's 100 ns resolution is fine.

**[3142305419] MeterVisual uses non-monotonic clock** — Switched sparkline bucket timing from `DateTime.UtcNow.Ticks` to `Stopwatch.GetTimestamp()` with `Stopwatch.Frequency`-based bucket-tick conversion. DST / NTP / manual clock changes can no longer make buckets advance backwards.

**[3142305422] ReconcileHighlightOverlay disposal** — Class now implements `IDisposable`. `Dispose` walks remaining sprite children, disposes each, then disposes the container, brushes, and the cached fade animations.

**[3142305426] ShowLayoutCost_DefaultIsFalse flakiness** — Test renamed to `ShowLayoutCost_GetterReadsAssignedValue` and rewritten to round-trip the getter/setter contract with save/restore. No longer claims to validate the default (which is unobservable in xUnit due to process-wide static state).

**[3142305432] ShowLayoutCost XML docs out of date** — Docs updated to describe the shipped mid-session behavior: enabling lazily builds the pipeline + wrapper and starts ETW; disabling stops ETW, disposes the sub-overlay, pauses the idle ticker, and (if no other overlay flag is on) tears the wrapper down.

**[3142305436] IsOverlayChrome attached property is dead code** — `LayoutCostOverlayAttached.cs` removed. Its only setter call (in `OverlayHostWiring`) had no readers — the descendant walk that would have consumed it was dropped along with the authored-count walk. Overlay-chrome filtering remains as a manual API (`LayoutCostAttribution.MarkOverlayChrome(elementId)`) pending the native-pointer interop. The canvas is Composition-pure anyway so there's no leak today; comment in `OverlayHostWiring.SetContentViaWrapper` now explains this.

All tests still green: 6,631 unit + 595 selftest, 0 failed.

codemonkeychris:
**Selftest flake tracking** — ran the suite 15× to characterize a flake noticed in earlier runs:

- 14/15 passes (all 595 fixtures green)
- 1/15 fails: `DataGrid_ScrollPerfRelative` — `dg=9.60 vl=6.04 ratio=1.6x` (threshold 1.4×)
- 0/75 LayoutCost-fixture invocations failed across the whole batch.

Failure is in `DataGridScrollFixtures.cs:317`, a perf-ratio comparison between DataGrid and VirtualList scroll latency. The fixture's own comment (lines 308–316) acknowledges the threshold sits at the boundary of normal ARM64 dev-box variance:

> "ARM64 dev boxes land at ~1.3–1.4x… The headroom to 1.4x absorbs that variance"

Not a regression from this PR — my changes don't touch DataGrid or VirtualList, and the only always-paid additions in the reconciler are ~5 bool-reads per UIElement mount (negligible vs the 1–2 ms measurement jitter at single-digit-ms scroll latencies). Pre-existing threshold-boundary flakiness, separately addressable by either bumping the threshold to 1.6× or switching from median to min of the N samples (min is less sensitive to GC/preemption spikes than median).

PR #93, opened 2026-04-25
async: replace UI-affined locks with dispatcher affinity

codemonkeychris:
## Summary

- Drops `PendingScope._lock` and `FocusRevalidationService._lock`. Both are now UI-thread-affined: lazy owner-thread capture + DEBUG-only `AssertOwnerThread` on every public method. Production callers (hook lifecycle, dispatcher-marshalled continuations, WinUI activation) already run on the UI thread; the dispatcher itself is the synchronization mechanism.
- Drops `threadSafe:true` from `Pending`'s rerender reducer — its only trigger is `PendingScope.Changed`, which is now UI-affined upstream.
- Adds `IHookDispatcher.InvokeAsync<T>(Func<T>)` as a typed escape hatch for background code that needs to *read* UI-affined state and get a value back (rare; default is still fire-and-forget `Post`).
- Three `PendingTests` dropped `await Task.Delay(...)` polls that resumed on thread-pool threads — with default-mode `TaskCompletionSource` the continuations run inline on `SetResult`, so the synchronous assertion is correct and avoids tripping the new affinity check.
- Updates `docs/reference/async-system.md` §1, §7.1, §8 and adds a new §8.1 documenting the per-component split (which locks stay, which became affinity assertions, and what's deferred).

## Deferred (documented)

- `InfiniteResource._lock` stays — three threading tests in `UseInfiniteResourceThreadingTests` deliberately drive `ItemAt`/`EnsureRange` from 8 background threads to verify the documented thread-safety contract. Removing requires either narrowing the contract (and deleting those tests) or migrating them to drive concurrent access through a dispatcher.
- `threadSafe:true` reducers in `UseResource`/`UseInfiniteResource`/`UseMutation` stay — the test-suite `InlineDispatcher` runs `Apply` on whichever thread completed the underlying `Task`; dropping requires a marshalling test dispatcher with explicit drain.
- `QueryCache` slot lock and `MutationHookState._lock` stay by design — shared cross-cutting state with off-UI-thread eviction; `RunAsync` is intentionally any-thread.

## Test plan

- [x] `dotnet test tests/Reactor.Tests` — 6562 passed, 0 failed
- [x] `Reactor.AppTests.Host --self-test` × 20 runs (42,000 checks total) — only two unrelated timing flakes observed (`Mutation_Completed`, `ScrollPerf_Ratio`), both in code paths not touched by this change
- [x] Stress-ran `FocusRevalidationTests` × 5 to verify lazy owner-thread capture survives xUnit's thread-switching between sequential tests

### DIFF
--- a/docs/reference/async-system.md
+++ b/docs/reference/async-system.md
@@ -937,9 +937,9 @@ synchronization mechanism." Today's split:
 | `MutationHookState._lock` | `Monitor` lock | `RunAsync` is intentionally callable from any thread; the lock protects `_pendingCount`/`_lastResult`/`_error` against the cross-thread caller. |
 | `InfiniteResource._lock` | `Monitor` lock | Documented thread-safe contract; threading tests drive `ItemAt`/`EnsureRange` from background threads. Production callers (virtualized list controls during layout) are UI-thread-affined, but the contract is the broader one. |
 | `UseResource` / `UseInfiniteResource` / `UseMutation` rerender reducer | `threadSafe: true` | The hook continuation `Apply` runs on the dispatcher thread in production, but the test-suite `InlineDispatcher` runs `Apply` on whatever thread completed the underlying `Task`. The `threadSafe` reducer is what makes those test paths safe. |
-| `PendingScope` | UI-thread affinity (`AssertOwnerThread`, lazy-captured, DEBUG-only) | All hook lifecycle paths run during render / cleanup / dispatcher-`Apply`, all UI thread. |
-| `FocusRevalidationService` | UI-thread affinity (`AssertOwnerThread`, lazy-captured, DEBUG-only) | Same. |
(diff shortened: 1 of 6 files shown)


codemonkeychris:
Addressed Copilot review in 45b66f2:

1. **PendingScope XML doc** — rewrote the threading paragraph to describe the actual lazy-on-first-call capture, and listed `AnyLoading` / `Count` alongside the mutators.
2. **Pending's reducer (correctness)** — reverted to `threadSafe: true`. The reviewer is right: `state.Dispatcher == null` paths run `Apply` inline on the Task completion thread, where `SetLoading → Changed → tick()` reaches the reducer off-UI. PendingScope's affinity assertion catches it in DEBUG; the reducer's lock keeps the rerender path safe in Release. Added a code comment explaining the edge case.
3. **FocusRevalidationService XML doc** — same lazy-capture wording fix.
4. **§8.1 doc claim** — qualified the `PendingScope` and `Pending`-reducer rows to call out the no-dispatcher edge case explicitly. `Pending`'s reducer row now reads `threadSafe: true` and explains why.

All 117 async unit tests still pass.

PR #99, opened 2026-05-01
Add Reconciler coverage fixtures + fix two flaky tests

codemonkeychris:
## Summary

- Add 38 new selftest fixtures (`tests/Reactor.AppTests.Host/SelfTest/Fixtures/ReconcilerBigCoverageFixtures.cs`) that exercise in-process Reconciler code paths. Boosts merged unit+selftest coverage of the Reconciler class from **65.7% → 74.5%**. Remaining gap is dominated by gesture/drag-drop/pointer-driven branches that require an OS input driver (already covered by the WinAppDriver E2E tier).
- Fix `Interactive_OnPan_Drag_ReportsTranslationAndPhase` failing reliably in the full E2E suite (passed in isolation): `NavigateToFixtureFresh` was unconditionally calling `ResetFixture` even when switching to a different fixture, racing the navigator's re-render and causing the next nav click to land on stale `ScrollView` geometry. Make the Reset conditional and bump the navigation timeout from 5s to 10s.
- Fix `UseResourceTests.Null_Dispatcher_Runs_Continuation_Inline` flaking under parallel test load: the threadpool continuation can take longer than a 10ms tick, so poll up to 1s instead.

## Test plan

- [x] `dotnet test tests/Reactor.Tests` — 6631 passed, 0 failed (11.2 s)
- [x] `dotnet run --project tests/Reactor.AppTests.Host -- --self-test` — 0 failures (108.7 s)
- [x] `dotnet test tests/Reactor.AppTests` (E2E w/ WinAppDriver) — 57 passed, 0 failed (140.8 s)


codemonkeychris:
Addressed all 5 Copilot review comments in `a073553`:

1. **`UseResourceTests.cs`** — switched the rerender flag to `int` with `Volatile.Write`/`Volatile.Read` (the continuation runs on the threadpool while the polling loop runs on the test thread, so a memory barrier is needed) and replaced `DateTime.UtcNow` with `Stopwatch` for a monotonic timeout.
2. **`AppTestBase.cs` docs** — moot for the final code: commit `58884e9` replaced the conditional Reset with an unconditional Reset plus a click retry, which is exactly what the existing `<remarks>` already describe.
3. **Unused `WinUI` alias** — removed.
4. **`Indeterchildlikeate` typo** — fixed to `Indeterminate`.
5. **`FrameAndMapMount` mismatch** — renamed class to `FrameMount`, updated section header to note MapControl is intentionally skipped (requires a map service token), and renamed the registry key `RBC_FrameAndMapMount` → `RBC_FrameMount`.

Re-ran all three test tiers after the changes:
- Unit: 6631 passed, 0 failed (10.4 s)
- Selftest: 0 failures (110.6 s)
- E2E: 57 passed, 0 failed (155.2 s)

PR #128, opened 2026-05-03
fix(devtools): accept HTTP.SYS-owned ports in PortOwnership check

codemonkeychris:
## Summary

`mur devtools` was unable to connect to any Reactor app launched with `--devtools run`. Every named verb returned exit 4 ("no live session"), and `session list` silently GC'd the lockfile on the way out. Reproduced cleanly against `samples/apps/chat` on Windows 11.

**Root cause**: `LockfileReader.IsLive` calls `PortOwnership.IsPortOwnedBy(port, lockfile.pid)` to defeat lockfile spoofing. The check enumerates the kernel TCP table and matches `row.OwningPid == lockfile.pid`. But Reactor's MCP server uses `HttpListener`, which routes through the kernel-mode HTTP.SYS driver — Windows attributes the listening socket to **PID 4 (System)**, not the user-mode app:

```
> Get-NetTCPConnection -State Listen -LocalPort 51965
LocalAddress LocalPort OwningProcess AddressFamily
127.0.0.1        51965             4
```

So no row ever matched, `IsLive` returned false, and `EndpointDiscovery.FindLiveHttpSessions` deleted the only valid lockfile.

## Fix

Extracted the policy into a pure `MatchListener(rows, port, pid, enumerationComplete)` predicate. Accepts the lockfile pid when **either**:

1. Any LISTEN row on the port directly attributes to it — preserves strict attribution for raw `TcpListener` users.
2. **Enumeration of both address families succeeded** *and* every LISTEN row on the port attributes to `PortOwnership.HttpSysPid` (PID 4) — HTTP.SYS holds the socket cleanly with no competing user-mode listener.

Mixed ownership (HTTP.SYS + a different user-mode pid on the same port) is rejected as ambiguous. Same-user spoof resistance for HTTP.SYS ports falls back to the per-launch random bearer token already enforced by the HTTP probe (spec 025 §5 / TASK-001) — documented in the class XML doc.

Other improvements rolled in:
- Enumerates **AF_INET6** in addition to AF_INET so `[::1]:<port>` bindings are reachable.
- Iterates **all** matching rows instead of bailing on the first port hit (the original code returned false on the first matching-port row regardless of owner).

## Review feedback addressed

[@copilot-pull-request-reviewer](https://github.com/apps/copilot-pull-request-reviewer) flagged a real spoof gap on partial TCP-table enumeration: if `TryGetTcpTable(AF_INET, …)` failed but `TryGetTcpTable(AF_INET6, …)` succeeded with only HTTP.SYS rows, the original fix would have accepted the lockfile — but a competing user-mode listener could exist on the unseen family.

Fixed in `c88ab25`. `IsPortOwnedBy` now tracks per-family enumeration success and threads it through to `MatchListener` as `enumerationComplete`. The HTTP.SYS-only fallback is gated on full enumeration; **direct pid attribution still works under partial enumeration** so legitimate sessions on rare degraded-enum systems don't break. XML doc updated on both methods to call out the policy.

## Test plan

- [x] **12** unit tests in `tests/Reactor.Tests/Devtools/PortOwnershipTests.cs` covering direct match, no listener, different user-mode pid, HTTP.SYS-only, HTTP.SYS dual-stack (v4+v6), HTTP.SYS + competing user-mode rejected, direct match wins over HTTP.SYS, irrelevant ports ignored, arg validation, **partial-enumeration HTTP.SYS-only rejected, partial-enumeration direct accepted, partial-enumeration direct-alongside-HTTP.SYS accepted**.
- [x] All **326** Devtools tests in `Reactor.Tests` pass (post-fix; was 323 in the original PR before the 3 new partial-enum tests).
- [x] End-to-end against `samples/apps/chat`: `mur devtools session list / version / windows / components / tree / state / screenshot` all return exit 0 with live data after the fix.
- [x] **End-to-end against `samples/Reactor.TestApp`** (validating the post-review fix specifically): launched via `mur devtools samples/Reactor.TestApp/Reactor.TestApp.csproj`, then ran `session list --pretty` → discovered `pid 43348 port 65323 http://127.0.0.1:65323/mcp` (the exact case the fix enables); `version` → reported the c88ab25 build tag back; `windows` → discovered main window; `tree --view summary` → 36 KB visual-tree JSON; `screenshot --out testapp.png --wait-idle` → 16 KB PNG of the Counter demo. Confirms the lockfile-discovery + HTTP-probe round-trip works under both the original fix and the partial-enumeration tightening.
- [ ] Reviewer: spot-check on x64 (only ARM64 hardware was available locally).

### DIFF
--- a/src/Reactor.Cli/Devtools/PortOwnership.cs
+++ b/src/Reactor.Cli/Devtools/PortOwnership.cs
@@ -39,6 +39,16 @@ internal static class PortOwnership
     /// <see cref="HttpSysPid"/> (HTTP.SYS) on at least one address family while
     /// no other user-mode process holds a competing listener on the same port.
     /// Returns false when no LISTEN row exists for the port at all.
+    /// <para>
+    /// <b>Failure mode for partial TCP-table enumeration:</b> if either
+    /// <c>AF_INET</c> or <c>AF_INET6</c> enumeration fails (rare; e.g. locked-down
+    /// loopback policy, kernel resource pressure), the HTTP.SYS-only fallback
+    /// is disabled — we'd otherwise accept a session whose only HTTP.SYS row is
+    /// on the visible family while a competing user-mode listener could exist
+    /// on the unseen family. Direct pid match is still honored on whatever
+    /// rows we did get back. This keeps the spoof defense intact under
+    /// degraded enumeration.
+    /// </para>
     /// </summary>
     public static bool IsPortOwnedBy(int port, int pid)
     {
@@ -46,11 +56,14 @@ internal static class PortOwnership
         if (pid <= 0) return false;
 
         var rows = new List<TcpListenerRow>();
-        if (TryGetTcpTable(AF_INET, out var v4)) rows.AddRange(v4);
-        if (TryGetTcpTable(AF_INET6, out var v6)) rows.AddRange(v6);
+        bool v4Ok = TryGetTcpTable(AF_INET, out var v4);
+        if (v4Ok) rows.AddRange(v4);
+        bool v6Ok = TryGetTcpTable(AF_INET6, out var v6);
+        if (v6Ok) rows.AddRange(v6);
         if (rows.Count == 0) return false;
 
-        return MatchListener(rows, port, pid);
+        bool enumerationComplete = v4Ok && v6Ok;
+        return MatchListener(rows, port, pid, enumerationComplete);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


codemonkeychris:
Addressed the partial-enumeration concern in `c88ab25`.

**Change**: `IsPortOwnedBy` now tracks whether *both* `AF_INET` and `AF_INET6` enumerations succeeded and threads that through to `MatchListener` as a new `enumerationComplete` parameter. The HTTP.SYS-only fallback is gated on `enumerationComplete == true`; if either family query fails, only direct pid attribution is honored. The spoof gap you described — accepting an HTTP.SYS-only signature on the visible family while a competing user-mode listener could exist on the unseen family — is closed.

This is reviewer option 2 ("only allow strict direct-pid match when enumeration is incomplete") rather than option 1 ("fail closed entirely on either-family failure"). Option 2 keeps legitimate sessions working on systems where one family's TCP-table query fails (rare but real — e.g. very locked-down loopback policy, kernel resource pressure under heavy load) without giving up spoof resistance.

**New tests**:
- `Match_PartialEnumeration_HttpSysOnly_RejectedAsUnsafe` — the exact scenario you flagged.
- `Match_PartialEnumeration_DirectMatch_StillAccepted`.
- `Match_PartialEnumeration_DirectMatchAlongsideHttpSys_StillAccepted`.

The 9 existing `MatchListener` tests pick up the new parameter as `enumerationComplete: true` (the live-system path's natural state). 12/12 `PortOwnership` tests pass; 326/326 Devtools tests pass.

XML doc updated on both `IsPortOwnedBy` and `MatchListener` to call out the partial-enumeration policy.

PR #164, opened 2026-05-06
fix(apptests): detect locked desktop / disconnected session

codemonkeychris:
## Summary
- E2E runs against WinAppDriver fail every `Click()` when the workstation locks (idle timeout, manual lock, RDP disconnect), surfacing as a flood of `WebDriverException "An unknown error occurred in the remote end"` that's indistinguishable from real test flake. Observed in a recent 50-run loop where 5 consecutive iterations failed 48/57 tests with identical stack traces, then recovered the moment the desktop was unlocked.
- Add `SessionInteractivityGuard` (P/Invoke to `OpenInputDesktop` + `WTSConnectState`). Locked screens switch the input desktop from `Default` to `Winlogon`; RDP disconnect changes `WTSConnectState` to `WTSDisconnected`. Both are deterministic and cheap.
- On detection: emit `Assert.Inconclusive` (not Fail) so the .trx outcome separates environmental from real failures, and write a marker file at `$E2E_LOCK_MARKER_PATH` so external multi-run loops can stop scheduling further iterations instead of generating more false negatives.

Hook points:
- `[TestInitialize]` in `AppTestBase` — preflights every test method.
- `try/catch (WebDriverException)` in `NavigateToFixture` — rechecks; lock → Inconclusive, otherwise rethrow as a real failure.
- Same recheck around `TestSession` session bootstrap so a class-init lock bails out cleanly without spending 30s booting WinAppDriver into a doomed run.

## Test plan
- [x] Build: `dotnet build tests/Reactor.AppTests`
- [x] Smoke run on unlocked desktop: `dotnet test tests/Reactor.AppTests` → 57/57 pass, no marker written.
- [ ] Lock-while-running: run `dotnet test`, lock the workstation mid-run, verify subsequent tests are reported `Inconclusive` (not `Failed`) and the marker file is written.
- [ ] Pre-locked: lock the workstation, run `dotnet test`, verify `TestSession.AssemblyInit` short-circuits before launching WinAppDriver/Host.

### DIFF
--- a/tests/Reactor.AppTests/Infrastructure/AppTestBase.cs
+++ b/tests/Reactor.AppTests/Infrastructure/AppTestBase.cs
@@ -18,8 +18,14 @@ public class AppTestBase
     /// </summary>
     protected static WindowsDriver<WindowsElement> Session => TestSession.Session;
 
-    // No TestInitialize/TestCleanup — NavigateToFixture handles fixture switching.
-    // Resetting between every test wastes ~5s on the implicit wait timeout.
+    // Per-test interactivity preflight — bails out as Inconclusive (not Failed)
+    // when the workstation is locked or the session is disconnected, so flake
+    // reports don't drown in environmental noise.
+    [TestInitialize]
+    public void GuardSessionInteractive()
+    {
+        SessionInteractivityGuard.EnsureInteractive("TestInitialize");
+    }
 
     private static string? _currentFixture;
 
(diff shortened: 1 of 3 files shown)


codemonkeychris:
Addressed all four review comments in `bf0c012`:

1. **AppTestBase L67-68 (dead code / misleading comment):** the `when (attempt == 0)` filter meant iteration 1's `WebDriverTimeoutException` always propagated out, making the post-loop `WaitForText` line genuinely unreachable. Removed the dead lines and the comment rather than papering over it with a longer timeout — the existing in-loop retry has been working.

2. **SessionInteractivityGuard L33 (NULL handle ≠ Locked):** now reads `Marshal.GetLastWin32Error()` immediately and only maps `ERROR_ACCESS_DENIED` (the documented signal that the calling thread can't access the active input desktop while Winlogon's secure desktop is up) to `Locked`. Other errors return `Unknown`, and `EnsureInteractive`/`RecheckAfterWebDriverFailure` now both pass `Unknown` through without throwing Inconclusive — avoids masking real failures.

3. **SessionInteractivityGuard L116 (race + stale carryover):** swapped to `FileMode.CreateNew` for an atomic first-writer-wins. The `IOException` branch is the "marker already exists, preserve the originating writer" path. The runner is responsible for clearing the path between iterations (it points at a fresh per-run directory each loop, so stale carryover isn't a concern in the supported flow).

4. **TestSession L48 (mid-init lock surfaces as TimeoutException, not WebDriverException):** extended the bootstrap `try` to also cover `WaitForHostWindow()`, and the catch now matches both `WebDriverException` and `TimeoutException` via a `when` pattern. A lock between the AssemblyInit preflight and the host window appearing is now reclassified to Inconclusive.

Build clean.

PR #176, opened 2026-05-06
fix(FlexPanel): honor VerticalAlignment.Stretch as definite block axis

codemonkeychris:
## Summary

- `FlexPanel`'s block axis was keyed off the panel's *declared* `Height` only — the parent's offer was ignored even when `VerticalAlignment.Stretch` + a finite parent slot meant \"fill\". The canonical web flex pattern of `header(auto) / body(flex:1) / footer(auto)` filling a viewport required a hardcoded `.Height(N)`, and a `flex:1` body resolved to an empty pool.
- Three coordinated changes in `MeasureOverride`, `ArrangeOverride`, and the Yoga `MeasureFunction` wrapper fix this without a parent-type sniff and without regressing the wobble fix from #172.
- New self-test `Flex_HeaderBodyFooterFillsParentSlot` covers the canonical pattern under a Reactor host with no explicit `.Height(N)` on the column.

## Why three changes

1. **`MeasureOverride`** — when no explicit `Height`, `VerticalAlignment.Stretch`, and parent offer is finite, treat block axis as definite. Symmetric counterpart to the existing `HorizontalAlignment.Stretch` inline-axis fill — same WinUI Stretch contract on both axes.
2. **`ArrangeOverride`** — same opt-in reruns Yoga at `finalSize.Height` to cover the case where the parent arranges at a different size than the measure offer. `_arranging` continues to suppress child re-measure, preserving the resize-wobble fix from #172.
3. **`MeasureFunction` wrapper (load-bearing)** — differentiate Yoga's `Exactly` mode (real stretch-fit, pass finite) from `AtMost` / `Undefined` (basis / content phase, pass infinity). Without this, a nested FlexPanel with default `VerticalAlignment.Stretch` would treat `AtMost`'s soft cap as a fill target and report the cap as its DesiredSize, defeating the outer's flex-grow distribution across siblings. With it, nested FlexPanels behave correctly under any intermediate (Border, Grid, ContentControl, etc.).

## Test plan

- [x] `dotnet test tests/Reactor.Tests` — 6819 passed, 0 failed
- [x] `dotnet test tests/Reactor.SelfTests` — 668 passed, 0 failed (includes the new `Flex_HeaderBodyFooterFillsParentSlot` fixture)
- [x] `dotnet run --project tests/Reactor.AppTests.Host -- --self-test --filter \"Flex\"` — all 31 flex fixtures pass, including new HBF
- [x] Manual: a demo app with `FlexColumn(titleBar, header, ..., listView.Flex(grow: 1, basis: 0)).Padding(24)` (no explicit `.Height`) now fills the window and the ScrollView produces a scrollbar instead of overflowing
- [ ] Recommended for reviewers: spot-check the wobble-prone scenarios (pure-width window resize over a header-rich column) — `_arranging` should still suppress the rerun's child re-measure, and Measure-time DesiredSize is still content-sized when parent offer is infinite


codemonkeychris:
Update pushed (`3aec44e`) addressing both Copilot review comments and the broader agent review.

## What I changed

**1. AtMost → infinity mapping reverted (Copilot point 1, agent point 1).** Both reviewers correctly flagged that mapping `YogaMeasureMode.AtMost` to `+∞` would under-report wrapped text height: `TextBlock.Measure(width=∞)` yields a 1-line result, `TextBlock.Measure(width=200)` yields the correct wrapped multi-line height. The wrapper now passes finite `w + margin` for both `AtMost` and `Exactly`, restoring the standard WinUI Measure contract.

The nested-FlexPanel discrimination (the original reason I changed the wrapper) is now done via a `[ThreadStatic] _outerYogaHeightMode` the wrapper publishes around `child.Measure`. `FlexPanel.MeasureOverride` reads it to tell:
- `null` → not nested under a flex measurement → use the WinUI Stretch contract.
- `Exactly` → stretch-fit allocation, fill.
- `Undefined` / `AtMost` → outer flex wants content size, do not fill (otherwise the inner reports the cap as `DesiredSize` and defeats the outer's `flex-grow` distribution).

This handles nesting through any intermediate (Border, Grid, ContentControl) without a brittle parent-type sniff, which was the failure mode of an earlier draft.

**2. Class name (Copilot point 2).** Renamed `FlexColumnFillsParentSlot` → `FlexHeaderBodyFooterFillsParentSlot` to match the registry key.

**3. New regression test for AtMost (agent point 5).** `Flex_AtMostPreservesTextWrapping` measures long text in a 200px column and asserts wrapped height ≥ 40px. Under a broken AtMost-as-infinity wrapper the height collapses to ~18px (1 line) — guards against the regression both reviewers flagged.

## What I'm pushing back on

**Agent point 2** — \"keep auto block-size content-sized during measure unless Height is explicit; the arrange-time rerun handles fill.\"

I tried this. It does not fix the user's primary scenario. Diagnostic from running the HBF fixture with measure-time `fillBlockAxis` removed:
```
# DIAG col=784x750 sv.Render=640 sv.Viewport=640 sv.Extent=640 sv.Scrollable=0 sv.Desired=640
```
The column resolved to its content height (750) instead of the host's slot. The host's `ContentPresenter` arranges at `min(slot, child.DesiredSize) = DesiredSize` when the child is shorter than the slot, so `ArrangeOverride.sizeChanged` is false and the rerun never fires. Without measure-time fill, `flex-grow` children have no pool to expand into in the *underflow* case, which is the normal pre-overflow state of any web flex layout.

The arrange-time rerun is necessary but not sufficient — it covers the slot-overflow case (parent passes slot, child wants more) but misses the slot-underflow case (parent passes slot, child wants less). The user's `dryrun.cs` is the underflow case (8 items × 80px + chrome = 750 < 768 window), which is exactly when measure-time fill matters.

**The trade-off the agent is right about** is real, though: a `Border` with auto-height wrapping a default-Stretch `FlexColumn` will inflate to the parent's offer rather than shrink-wrap to content, because we are mapping \"Stretch + finite parent\" to CSS web-flex `height: 100%` at measure time. New self-test `Flex_BorderAutoHeightInflatesWithStretch` makes this testable and documents the opt-out — `.VAlign(VerticalAlignment.Top)` makes the inner shrink-wrap as expected. Asserts both halves so a future refactor that lands web-flex semantics without auto-height inflation has a known signpost to flip.

**Agent points 3 (AlignContent default) and 4 (arrange-time stale measurements)** — both fair observations but out of scope for this PR. AlignContent default is a separate semantic change (would affect any wrapped multi-line flex layout); arrange-time stale measurements is the existing wobble-fix tradeoff from #172, which this PR doesn't change.

## Test results

- `dotnet test tests/Reactor.Tests` — 6819 passed, 0 failed
- `dotnet test tests/Reactor.SelfTests` — 670 passed, 0 failed (3 new fixtures: HBF, Border auto-height, AtMost text wrapping)

PR #229, opened 2026-05-10
Replace em-dashes and check for .NET 10 instead

dotMorten:
## Summary

Script failed to execute due to use of em-dash'es instead of normal dashes, but weren't escaped.
Also updated version check to .NET 10 to match MUR target.

## Test plan

- [x] Manually ran `install-skill-kit.ps1`


codemonkeychris (approved):
thanks!

PR #239, opened 2026-05-10
Avoid adding the temporary tools path to global path environment

dotMorten:
## Summary
The integration-test fixture no longer lets dotnet mutate the user-scoped PATH.

I updated tests\Reactor.IntegrationTests\Packaging\CreateTemplateTests.cs to set `DOTNET_ADD_GLOBAL_TOOLS_TO_PATH=false` in the test command environment, which keeps the CLI’s temp `dotnet-home\.dotnet\tools` path process-local instead of appending it to your global PATH. 
## Test plan
- [x] Manually ran `dotnet run --project tests\Reactor.IntegrationTests\Reactor.IntegrationTests.csproj` and verified global environment paths no longer gets modifed.


codemonkeychris (approved):
Magic

PR #250, opened 2026-05-12
Spec 038 Phase 3 — Class-A rules, correctness fixes, EC3 results, template typo fix

codemonkeychris:
## Summary

- **Phase 3 (Tier-3 rules):** three new Class-A induced rules driven by the cross-agent reproducibility audit (gpt-5.5 + claude-sonnet-4.6 × 525-run corpora each) — `GridSizeFactoryParensRule` (CS1955, 146 events combined, top freq in both corpora, first cross-tier rule), `GridSizePxRenameRule` (CS0117, 9 events), `TextBlockStyleHintRule` (CS1061/CS0117, 5 events across two syntactic shapes). `ThemeBackgroundSuffixRule` reclassified Class-B → Class-A in the file-header comment.
- **Two critical correctness fixes** uncovered by end-to-end smoke testing — both were silently no-op'ing every rule firing in production despite all unit tests passing:
  1. `CompilationLoader` now resolves `ProjectReference` outputs from `project.assets.json`'s `libraries.<id>` entries with `type=project`. Without this, Reactor itself (a project reference for every sample app) was invisible to `RuleSymbolResolver` and every rule's `DeclaredTargets` failed — the whole registry self-disabled on every real invocation.
  2. Suggest-gate carve-out for Tier-3 rules. `SuggesterOrchestrator` takes `tier2Enabled: bool`; Tier-3 rules always run when their diagnostic code surfaces, Tier-2 stays gated. EC2 watch-item ("Phase-3 rules are the right lever — not Phase-2.x gate tuning") finally addressed in code.
- **Template-identity typo fix** (`Micrsoft.UI.Reactor.CSharp` → `Microsoft.UI.Reactor.CSharp` in `template.json`) — was breaking `dotnet new reactorapp` resolution against accumulating template caches; hit 20/20 runs across both arms of the EC3-original batch.
- **Template-shape cleanup for agent reasoning:** removed `<ImplicitUsings>enable</ImplicitUsings>` from the scaffolded csproj; explicit using directives (`System + Microsoft.UI.Reactor + .Core + .Layout + Xaml + Xaml.Controls + static Factories`) baked into `App.cs`. Every symbol now traces to a visible `using` at the top of the file. The starter App.cs only uses three of the seven imports — the rest are there for the namespaces the agent reaches for within the first few turns of any real app.
[...]

PR #258, opened 2026-05-13
Polish default template: TitleBar + content padding

nmetulev:
Use the modern Windows TitleBar (drag region, system menu, themed caption) as the top-of-window element and wrap content in a Border with 24px padding. Apply the same polish to the `mur --create` scaffolder so both entry points produce a presentable starter app. Align the scaffolder's emitted usings with the dotnet new template (PR #250) so generated apps have the common WinUI/Reactor namespaces ready to go.


codemonkeychris:
i think i made the same type of changes :(

please sync and resolve conflicts, i bet we are close on the same implementation

PR #271, opened 2026-05-13
Add standalone IconElement for the element tree

### LINKED ISSUE #257: [Feature] There's no mapping for Microsoft.UI.Xaml.Controls.IconSourceElement
### Problem

When trying to render a symbol icon, I was looking for something to expose [IconSourceElement](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.iconsourceelement?view=windows-app-sdk-1.8) as an Element to return, but found no references to it in the code base.

### Proposed solution

```
public record IconElement(IconData data) : Element
{
    public double? Width { get; init; }
    public double? Height { get; init; }
    internal Action<WinUI.IconElement>[] Setters { get; init; } = [];
}
```

And potentially AnimatedIconElement etc extends a common iconelement type?

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

- [x] I have searched [existing issues and specs](https://github.com/microsoft/microsoft-ui-reactor/issues?q=is%3Aissue) for prior discussion of this idea.

sundaramramaswamy:
## Summary

Adds a new `IconElement(IconData Data)` record that wraps the existing `IconData` hierarchy so icons can be placed as standalone children in any layout container.

**Before:** `IconData` records (`SymbolIconData`, `FontIconData`, etc.) could only be used as data slots on composite controls (NavigationViewItemData, MenuFlyoutItemData, etc.). There was no way to render a standalone icon in the element tree.

**After:**
```csharp
Icon("Home")
Icon(FontIcon("\uE700", "Segoe Fluent Icons"))
Icon(BitmapIcon(new Uri("ms-appx:///icon.png")))
```

## Changes

| File | Change |
|------|--------|
| `Element.cs` | New `IconElement(IconData Data) : Element` record |
| `Dsl.cs` | `Icon(IconData)` and `Icon(string)` factory methods |
| `Reconciler.Mount.cs` | `MountIcon` — delegates to existing `ResolveIcon` |
| `Reconciler.Update.cs` | `UpdateIcon` — patches same-type, replaces on cross-type change |
| `ElementExtensions.cs` | `.Set(Action<IconElement>)` fluent modifier |
| `ElementRecordCoverageTests.cs` | 4 new unit tests |

## Testing

- All 7237 unit tests pass (0 failures)
- Full solution builds cleanly

Closes #257

### DIFF
--- a/skills/reactor.api.txt
+++ b/skills/reactor.api.txt
@@ -81,6 +81,9 @@ HyperlinkButton(Command command) → HyperlinkButtonElement
 HyperlinkButton(string content, Uri navigateUri = null, Action onClick = null) → HyperlinkButtonElement
 If(bool condition, Func<Element> then, Func<Element> otherwise = null) → Element
 Image(string source) → ImageElement
+Icon(IconData data) → IconElement
+Icon(Symbol symbol) → IconElement
+Icon(string symbol) → IconElement
 ImageIcon(Uri source) → ImageIconData
 InfoBadge() → InfoBadgeElement
 InfoBadge(int value) → InfoBadgeElement
@@ -308,6 +311,7 @@ HyperlinkButtonElement.TextLink() → HyperlinkButtonElement
 ImageElement.ImageFailed(Action<string> handler) → ImageElement
 ImageElement.ImageOpened(Action handler) → ImageElement
 ImageElement.NineGrid(Thickness nineGrid) → ImageElement
+IconElement.Set(Action<IconElement> configure) → IconElement
 ImageElement.Set(Action<Image> configure) → ImageElement
 InfoBadgeElement.Set(Action<InfoBadge> configure) → InfoBadgeElement
 InfoBarElement.ActionButtonClick(Action handler) → InfoBarElement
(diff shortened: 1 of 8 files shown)


skills/reactor.api.txt
```diff
@@ -73,6 +73,8 @@ Heading(string content) → TextBlockElement
 Hyperlink(string text, Uri navigateUri) → RichTextHyperlink
 HyperlinkButton(Command command) → HyperlinkButtonElement
 HyperlinkButton(string content, Uri navigateUri = null, Action onClick = null) → HyperlinkButtonElement
+Icon(IconData data) → IconElement
+Icon(string symbol) → IconElement
```
dotMorten, line 86: Do we need one that takes the Symbol enum? It'll be a lot easier to code again instead of a stringified version of it? https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.symbolicon.symbol?view=windows-app-sdk-1.8#microsoft-ui-xaml-controls-symbolicon-symbol

It would then generate the [SymbolIcon](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.symbolicon?view=windows-app-sdk-1.8) IconElement

codemonkeychris: yeah, i think the enum version is better.

codemonkeychris (changes requested):

skills/reactor.api.txt
sundaramramaswamy: Good call - added an Icon(Symbol) overload in f18aed9 so you get full IntelliSense on the enum. The string overload is kept for raw glyph strings not in the enum.

sundaramramaswamy:
@codemonkeychris All feedback has been addressed — added the Icon(Symbol) enum overload in f18aed9. Could you re-review when you get a chance?

codemonkeychris (approved):

PR #296, opened 2026-05-14
refactor: normalize ShowAddButton to IsAddButtonVisible

### LINKED ISSUE #268: API consistency: boolean-style modifiers mix adjective, Is*, and imperative naming
External API review flagged that boolean-style modifiers across the surface mix several naming styles:

- Adjective: \`Disabled\`, \`Editable\`, \`Selectable\`, \`Closable\`, \`Active\` (\`omd.html:135-146, 182, 197, 239, 388, 486\`)
- \`Is*\`: \`IsTabStop\`
- Imperative: \`ShowAddButton\`

DSLs usually choose one style for modifiers. Mixing all three undermines IntelliSense discoverability — users can't predict whether the toggle is \`IsX\`, \`X\`, or \`ShowX\`.

**Proposed normalization:** pick one rule (e.g. bare adjective for fluent modifiers — \`Disabled()\`, \`Editable()\`, \`AddButtonVisible()\`) and migrate the outliers, with \`[Obsolete]\` shims for the old names.

Source: external API consistency review, 2026-05.

sundaramramaswamy:
Rename `ShowAddButton()` → `IsAddButtonVisible()` to align with WinUI's `Is*` naming convention for boolean properties (`Control.IsTabStop`, `TabView.IsAddTabButtonVisible`, etc.).

- **`IsTabStop`** — kept as-is (already matches WinUI naming per review feedback)
- **`ShowAddButton`** → **`IsAddButtonVisible`** (imperative verb → `Is*` boolean)
- Old `ShowAddButton` kept as `[Obsolete]` shim for backward compatibility
- `reactor.api.txt` updated in both locations

Fixes #268

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -476,7 +476,7 @@ class App : Component
     }
 
     static Element NavBtn(string label, string current, Action<string> set) =>
-        Button(label, () => set(label)).Disabled(label == current);
+        Button(label, () => set(label)).IsEnabled(!(label == current));
 }
 ```
 
--- a/docs/_pipeline/ai-author-skill.md
+++ b/docs/_pipeline/ai-author-skill.md
@@ -702,7 +702,7 @@ case (no deps); `Memo(ctx => …, deps)` to also re-render when any dep changes;
 `.Background(color|ThemeRef)`, `.Foreground(color|ThemeRef)`,
 `.CornerRadius(n)`, `.WithBorder(color|ThemeRef, thickness?)`,
 `.HAlign(alignment)`, `.VAlign(alignment)`,
-`.Disabled(bool)`, `.Visible(bool)`, `.WithKey(string)`,
+`.IsEnabled(!bool)`, `.IsVisible(bool)`, `.WithKey(string)`,
 `.Flex(grow?, shrink?, basis?)`, `.ToolTip(string)`,
 `.FocusTrap(FocusTrapHandle)`,
 `.Set(control => { /* raw WinUI access */ })`
--- a/docs/_pipeline/apps/accessibility/App.cs
+++ b/docs/_pipeline/apps/accessibility/App.cs
@@ -83,7 +83,7 @@ class AccessibleFormDemo : Component
             CheckBox(agree, setAgree, label: "I accept the terms")
                 .TabIndex(3),
             Button("Register", () => { })
-                .Disabled(!valid).TabIndex(4).AccessKey("R")
+                .IsEnabled(valid).TabIndex(4).AccessKey("R")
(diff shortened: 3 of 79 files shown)


codemonkeychris (changes requested):
I think we should align with the WinUI names, IsXXX

sundaramramaswamy:
@codemonkeychris Addressed — updated to align with WinUI's `Is*` naming convention:

- **`IsTabStop`** stays as-is (already matches `Control.IsTabStop`)
- **`ShowAddButton`** → `IsAddButtonVisible` (matches `TabView.IsAddTabButtonVisible` pattern)

Force-pushed with the narrower scope.

sundaramramaswamy:
Thanks for the direction, Chris. Agreed on `ShowAddButton` — that's clearly an imperative outlier and is now `IsAddButtonVisible`.

For the remaining adjective modifiers (`Disabled`, `Editable`, `Selectable`, `Closable`, `Active`), wanted to surface a trade-off before expanding scope:

The adjective forms are sugar over the `Is*` record properties — e.g. `Closable(false)` sets `IsClosable = false`, `Editable()` sets `IsEditable = true`. Renaming these to `IsClosable()`, `IsEditable()`, `IsActive()` is straightforward.

The one case that doesn't map cleanly is `Disabled()`. The WinUI property is `IsEnabled` — opposite polarity. Today you write `Button("Save").Disabled()`, which reads naturally. The `Is*` equivalent would be `IsEnabled(false)`, which loses the ergonomic default.

Should we do (1) or (2)?

1. rename all five to `Is*` (with `IsEnabled(false)` replacing `Disabled()`)
2. rename the four that map directly (`Editable` → `IsEditable`, `Closable` → `IsClosable`, `Active` → `IsActive`, `Selectable` → `IsTextSelectionEnabled`) and keep `Disabled` as-is since it inverts `IsEnabled`?

codemonkeychris:
> The one case that doesn't map cleanly is `Disabled()`. The WinUI property is `IsEnabled` — opposite polarity. Today you write `Button("Save").Disabled()`, which reads naturally. The `Is*` equivalent would be `IsEnabled(false)`, which loses the ergonomic default.

Please look at the equivalent in WinUI and follow the pattern to align.

sundaramramaswamy:
@codemonkeychris Addressed your feedback — all boolean modifiers now use exact WinUI property names (IsEnabled, IsReadOnly, IsEditable, IsClosable, IsActive, IsVisible, IsTextSelectionEnabled, IsDisabledFocusable, IsLightDismissEnabled, IsAddTabButtonVisible). Old names kept as [Obsolete] shims. Ready for re-review when you get a chance.

codemonkeychris (approved):

PR #324, opened 2026-05-18
fix(virtualization): proper ItemsRepeater recycling — stop leaking every realized container

codemonkeychris:
## Summary

- `ElementFactory<T>` now reuses `args.Element` from `RecycleElement` via a recycle stack + `Reconciler.Reconcile` instead of always Mounting fresh, matching the WinUI `IElementFactory` contract.
- `FlexPanel` no longer subscribes to `XamlRoot.Changed` per instance — DPI scale is read lazily inside `MeasureOverride`, which removes the leak path that `Loaded`/`Unloaded` couldn't close because those events don't fire for `ItemsRepeater`-recycled containers.
- Three new SelfTest fixtures pin the IElementFactory recycling contract so this regression can't sneak back in.

## Root cause

Reading `microsoft-ui-xaml-lift`:

- `ViewManager.cpp:865-869` — at realize, the framework does `children.Append(element)` only when `GetParent(element) != *repeater`. Recycled elements **stay parented** to the ItemsRepeater.
- `ViewManager.cpp:153-194` — `ClearElementToElementFactory` calls our `RecycleElement` but does NOT remove from `repeater.Children`. The remove branch runs only when there is no `ItemTemplate`.
- `ItemTemplateWrapper.cpp:120-136` — the built-in `RecyclingElementFactory` cycles the same UIElements via `RecyclePool`; the framework's same-children-set design depends on the factory re-yielding its elements.
- `ItemsRepeater.cpp:653-655` comment: *\"ownership of the items is with Repeater today because the wrapper is doing the recycling as opposed to the DataTemplate.\"*

Our factory's \"build fresh on every realize, drop reference on recycle\" pattern produced **one orphan in `Repeater.Children` per realize**. At 2,750 scroll cycles the fixed-height demo had ~2,750 UIElements parented forever; managed heap climbed 20→70 MB and working set climbed 247→448 MB.

The variable-height demo additionally crashed with stowed exception `0xC000027B` because the orphan pile-up interacted with `RefreshRealizedItems` — `_listState.ByKey[key].Index` resolved a stale Element to a now-foreign realized container, and a mismatched `Reconcile` walked a structurally-divergent tree (rows where `i % 7 == 0` had an extra TextBlock vs. neighbors) on the XAML thread.

## Fix details

**`ElementFactory<T>`** (commit `f719976`)

- New `Stack<UIElement> _recyclePool` and `Dictionary<UIElement, Element> _lastElementByControl`.
- `RecycleElement(args)`: drop `_mountedElements` / `_keyByControl` tracking for the key, **don't** call `UnmountChild`, push to `_recyclePool`.
- `GetElement(args)`: if pool non-empty, pop a control, look up its prior Element, call `Reconciler.Reconcile(oldElement, newElement, control, ...)`, and reuse the control. Only Mount fresh when the pool is empty.
- Removed `PoolInteractiveLeaves` / `IsPoolableInteractive` — the per-leaf pooling path conflicted with whole-row reuse.

**`Reconciler` reverse-lookup map** (`_keyByControl`)

Supports O(1) cleanup in `RecycleElement` and prevents `RefreshRealizedItems` from running `Reconcile` against a stale `_mountedElements` entry whose key→index resolved to a different realized row. This was the proximate cause of the variable-height crash.

**`FlexPanel`**

Replaced `OnLoaded` + `XamlRoot.Changed += OnXamlRootChanged` with `SyncPointScaleLazy()` called at the top of `MeasureOverride`. Same DPI-tracking behavior, zero subscriptions. WinUI does not reliably fire `Loaded`/`Unloaded` for ItemsRepeater-recycled containers (observed: 12 unloads for 2,774 constructions), so the prior subscribe path leaked every FlexPanel through XamlRoot's multicast delegate list.

## Tests (commit `5c33ddc`)

Three SelfTest fixtures in `ElementFactoryRecyclingFixtures.cs` exercise the `IElementFactory` contract on a standalone `ElementFactory<T>` (no live LazyVStack, so framework-driven realizes don't collide with the test's cycles):

| Fixture | What it pins | Pre-fix value |
|---|---|---|
| `EFR_Factory_BoundedDistinctControls_AcrossManyRealizeCycles` | 100 realize+recycle cycles return ≤ 5 distinct UIElement instances | 100 |
| `EFR_Factory_RecycledControlIsReusedOnNextRealize` | `RecycleElement(X)` followed by `GetElement(...)` returns the same X by reference | not equal |
| `EFR_Factory_BookkeepingBoundedAcrossCycles` | `_mountedElements` / `_keyByControl` / `_recyclePool` / `_lastElementByControl` counts stay bounded at ≤1 after 50 cycles | 50 / 50 / 0 / 0 |

The fixtures use internal `Debug*Count` accessors on `ElementFactory<T>` (gated by the existing `InternalsVisibleTo` on `Reactor.AppTests.Host`).

## Holistic testing-gap analysis

This bug class — \"WinRT/UIElement retained by some framework or static path that our code can't easily inspect\" — is hard to catch with traditional assertions because:

1. **GC ordering is non-deterministic** for CCWs. Naïve `WeakReference` checks misfire.
2. **`Loaded`/`Unloaded` are not a reliable lifecycle signal** inside `ItemsRepeater` (we now know).
3. **Process-level memory metrics are noisy** without forced Gen2 + finalizers, and even then native CCW retention skews working-set numbers more than managed heap.

Other Reactor code paths with similar exposure (followups, not in this PR):

- `Reconciler._interactionTrackers` — `static Dictionary<UIElement, ...>`, leaks if `Unmount` doesn't fire (same trap we just hit).
- `Reconciler._keyframeTriggerValues` — same shape.
- Any `_pending*` collections that capture a `UIElement` in a deferred lambda (e.g., `_pendingConnectedAnimationStarts`, `ApplyMoveAnimations` dispatcher-enqueued closure).
- Any `Window`-rooted event subscription (`CompositionTarget.Rendering`, `XamlRoot.Changed`).

A reusable test scaffold for future leak coverage would look like the EFR fixtures here: build a small isolated factory or component tree, drive the suspect path N times directly, then assert the internal collection counts are bounded. Two reasons that pattern works where naïve managed-heap profiling fails: (a) you assert against deterministic internal state, not GC outcomes; (b) you isolate from the framework's own activity so deltas mean what you think they mean. I'd recommend adding similar coverage when we touch the other suspect paths above.

## Empirical verification (from the original repro)

Scrolling DataSystemDemo → VirtualList (Fixed Height, 100k items, 14,506 realize/recycle cycles), with instrumented Gen2+finalizer GC at each readout:

| Metric | Before | After |
|---|---|---|
| FlexPanel constructions | 2,774 (1:1 w/ realizes) | **50 (flat)** |
| Managed heap | 20 → 70 MB | **17 MB (flat)** |
| Working set | 247 → 470 MB | **238 MB (flat)** |

Variable-height crash (stowed exception `0xC000027B`) also no longer reproduces.

## Test plan

- [x] `dotnet test tests/Reactor.Tests` — 7,648 passed / 0 failed
- [x] `dotnet test tests/Reactor.SelfTests` — 717 passed / 0 failed (+3 new EFR fixtures)
- [x] Manual: DataSystemDemo → VirtualList (Fixed Height) → scroll 14k+ cycles, memory flat
- [x] Manual: DataSystemDemo → VirtualList (Variable Height) → scroll heavily, no crash

### DIFF
--- a/src/Reactor/Core/ElementFactory.cs
+++ b/src/Reactor/Core/ElementFactory.cs
@@ -33,6 +33,31 @@ public sealed partial class ElementFactory<T> : IElementFactory
     private readonly Dictionary<string, Element> _mountedElements =
         new(global::System.StringComparer.Ordinal);
 
+    // Reverse lookup: realized WinUI control → key. Lets RecycleElement drop
+    // the matching _mountedElements entry in O(1) when ItemsRepeater hands a
+    // container back. Without this, entries accumulate one per unique key as
+    // the user scrolls (every realize adds; recycle never removes), and on
+    // any subsequent re-render RefreshRealizedItems walks stale entries
+    // whose row.Index now points at a different logical row's container —
+    // running Reconcile against a mismatched UIElement tree.
+    private readonly Dictionary<UIElement, string> _keyByControl = new();
+
+    // Recycle pool for proper WinUI ItemsRepeater integration. The framework
+    // keeps every realized UIElement parented to the repeater forever and
+    // expects the factory to cycle them — see ViewManager.cpp:865-869 in the
+    // microsoft-ui-xaml-lift source: on realize, it skips Append if the
+    // returned control is already parented to the repeater. So a recycled
+    // container must come back out via GetElement to keep the working set
+    // bounded; allocating fresh on every realize creates one orphan in
+    // Children per call.
+    private readonly Stack<UIElement> _recyclePool = new();
+
+    // Last Element bound to a given realized control. On reuse from the
+    // recycle pool, this is the oldElement passed to Reconciler.Reconcile so
+    // the existing WinUI tree gets diffed-in-place against the new content
+    // rather than thrown away and re-mounted.
+    private readonly Dictionary<UIElement, Element> _lastElementByControl = new();
+
     public ElementFactory(
         IReadOnlyList<T> items,
         Func<T, int, Element> viewBuilder,
@@ -160,8 +185,38 @@ public sealed partial class ElementFactory<T> : IElementFactory
 
         var item = _items[index];
         var element = _viewBuilder(item, index);
+
+        UIElement? control;
+        if (_recyclePool.Count > 0)
+        {
+            // Reuse a previously-recycled container. The framework still has
+            // it parented to the ItemsRepeater, so the ViewManager.cpp:866
+            // Append-skip kicks in and the visual tree stays stable.
+            var reused = _recyclePool.Pop();
+            if (_lastElementByControl.TryGetValue(reused, out var oldElement))
+            {
+                var replacement = _reconciler.Reconcile(oldElement, element, reused, _requestRerender);
+                control = replacement ?? reused;
+            }
+            else
+            {
+                // Defensive: pool entry without a tracked oldElement should
+                // not happen — fall back to re-mounting on top of it.
+                control = _reconciler.Mount(element, _requestRerender);
+            }
+        }
+        else
+        {
+            control = _reconciler.Mount(element, _requestRerender);
+        }
+
         _mountedElements[key] = element;
-        var control = _reconciler.Mount(element, _requestRerender);
+        if (control is not null)
+        {
+            _keyByControl[control] = key;
+            _lastElementByControl[control] = element;
+        }
+
         return control ?? new TextBlock { Text = "" };
     }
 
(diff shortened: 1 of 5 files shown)


codemonkeychris:
Addressed Copilot's review (commit \`cd8d244\`). Three of the four findings were real defects; the fourth is a design-scope question I've tracked for a follow-up.

| # | Finding | Status |
|---|---|---|
| 1 | `Reconcile` returns a non-null replacement → popped \`reused\` becomes an orphan in \`Repeater.Children\` (heterogeneous rows) | Fixed: \`DetachFromParent\` + drop \`_lastElementByControl\` for the orphan |
| 2 | \`RefreshRealizedItems\` updates \`_mountedElements\` but leaves \`_lastElementByControl\` stale → next reuse diffs the wrong tree | Fixed: refresh now updates both dicts |
| 3 | Recycle preserves Component \`UseState\`/\`UseEffect\` across logical items because keySelector isn't propagated to row Element.Key | **Not in this PR.** Matches WinUI \`RecyclingElementFactory\`'s documented behavior. Reactor should give users a clean opt-in to per-item state reset — tracking as separate work since it's a design question, not a regression fix |
| 4 | \`RecycleElement\` skips \`UnmountChild\` and \`Reconciler.Unmount\` doesn't descend into ItemsRepeater children → row Components' cleanups never run on LazyStack unmount, leaking timers/subscriptions | Fixed: added ItemsRepeater branch to \`UnmountRecursive\` (walks via VisualTreeHelper since the C# projection of ItemsRepeater is FrameworkElement, not Panel — confirmed in \`microsoft-ui-xaml-lift/.../ItemsRepeater.idl:154\`) |

Three new SelfTest fixtures (one per fix) pin each as a regression gate:

- \`EFR_Factory_ReplacementOnRootTypeChange_DropsOldControlTracking\`
- \`EFR_Factory_RefreshRealizedItems_SyncsLastElementByControl\`
- \`EFR_LazyStack_Unmount_CleansUpAllRecycledRowComponents\` — drives a counter via \`UseEffect(() => () => count++)\` on Components inside rows, then unmounts the LazyStack and asserts cleanups fired

Tests: 7,648 unit / 720 self (was 717, +3 new) all pass.

codemonkeychris:
Filed #326 to track the per-row Component state design question (finding #3 from the Copilot review).

PR #354, opened 2026-05-20
loc: replace reflection-based args with tuple-params + IDictionary overloads

codemonkeychris:
## Why

PR #353 skipped `Localization_LocaleSwitching` under AOT because
`t.Message(key, new { name = "World" })` reflects over the anonymous type
to build the ICU args dictionary, and the trimmer drops the property
getters under NativeAOT (ctor + backing fields are statically reachable
from the `new { ... }` site, but property getters aren't, since the only
reader was a reflective `GetProperties()` call). `GetProperties()`
returned empty, `MessageFormat.FormatMessage` threw on `{name}`, the
`catch` swallowed, and the raw ICU pattern leaked through to the UI.

We don't want a reflection-based dispatcher in this code path anyway —
it forces every `[UnconditionalSuppressMessage]` we''ve written on the
type, and the security model (refuse arbitrary DTOs via
`IsAcceptableArgsContainer`) is doing nothing the type system couldn''t
do statically.

## What

Replace the `object?`-typed `args` parameter with two AOT-safe overloads:

```csharp
// 1. Explicit dict — for shared/programmatic args
public string Message(MessageKey key, IDictionary<string, object>? args = null);

// 2. Tuple-params — for the compact inline case
public string Message(MessageKey key,
    (string Name, object Value) arg1,
    params (string Name, object Value)[] more);
```

Callers go from this:

```csharp
t.Message(Greeting, new { name = "World" })
t.Message(SearchResults, new { count = 0, query = "test" })
```

to this:

```csharp
t.Message(Greeting, ("name", "World"))
t.Message(SearchResults, ("count", 0), ("query", "test"))
```

Same shape for `RichMessage`. The tagged variant uses the dict overload
because `params` must be last:

```csharp
t.RichMessage(Loc.Profile.Welcome,
    args: new Dictionary<string, object> { ["name"] = "Alice" },
    tags: new() { ["bold"] = text => Text(text).Bold() })
```

`ValueTuple` is a nominal type with concrete fields — building a
`Dictionary<string,object>` from tuple values needs zero reflection,
so the whole IL2026/IL2070 surface for this method disappears.

## Deleted

- `ToArgsDictionary`, `IsAcceptableArgsContainer`, `_propertyCache`
- The `[UnconditionalSuppressMessage("Trimming", "IL2026"/"IL2070")]` attrs
- `using System.Reflection`, `System.Collections.Concurrent`,
  `System.Diagnostics.CodeAnalysis` (now unused)
- The `[LocArgs]` soft opt-in marker (string-name match in
  `IsAcceptableArgsContainer`)

## Migrated callers

- Selftest + AppTests localization fixtures (two files)
- `IntlAccessorTests`, `RichMessageTests`
- `Reactor.Cli/Loc/SourceRewriter`: emits tuple syntax now. CLI codegen
  for interpolated strings now writes `t.Message(key, ("name", expr))`
  instead of `t.Message(key, new { name = expr })`. Updated
  `SourceRewriterTests` assertions.
- Docs: `_pipeline/templates/localization.md.dt` narrative and
  `_pipeline/apps/localization/App.cs` snippets (the compiled
  `docs/guide/localization.md` regenerates from these). Plus the
  hand-edited `docs/reference/localization-howto.md` example.

## AOT skip list

- `Localization_LocaleSwitching` removed from `DefaultAotSkipPatterns`
  in `SelfTestRunner.cs`.
- "Localization with anonymous-type args" row removed from
  `docs/aot-support.md` — there''s no anonymous-type path left to break.

## Verified

- `dotnet build src/Reactor` — clean (AOT warnings remain as errors).
- `dotnet test tests/Reactor.Tests --filter "FullyQualifiedName~Localization"` — 282/282 pass.
- `dotnet run --project tests/Reactor.AppTests.Host -- --self-test`
  (JIT, full suite) — 0 failures.
- AOT republish + selftest pending verification on this PR; will report
  back with numbers.

## Breaking change

`t.Message(key, new { name = "..." })` no longer compiles. The compiler
error is reasonably clear ("cannot convert anonymous type to
`IDictionary<string, object>`"); no analyzer needed. Migration is a
mechanical `new { name = X }` → `("name", X)`.

## Base

This PR is based on `codemonkeychris/aot-selftest-skips-round3`
(PR #353). When that merges, GitHub will retarget this PR to `main`.


codemonkeychris:
**AOT verification complete.**

Full AOT selftest run against the published binary on this branch:

| Metric | Before (PR #353) | After this PR |
|---|---|---|
| `ok` sub-assertions | 2184 | **2224** |
| `not ok` | 0 | **0** |
| `# SKIP` | 193 | **192** |
| `Localization_LocaleSwitching` | skipped | **runs, 40/40 pass** |

The 40 newly-passing sub-assertions cover Greeting / Plural_Zero / Plural_Five / Search_Zero / Search_One / Search_Many / Direction across `en-US`, `ko-KR`, `ar-SA`, `ja-JP`, `de-DE`, plus en-US round-trip. Tuple-params overload behaves identically under AOT and JIT.

Pushed `CONTRIBUTING.md` count refresh to match (192 skipped, ~543 passed, 0 failed).

codemonkeychris:
Addressed all 5 reviewer findings in 615b10f7:

1. **BuildArgs null semantics** — switched the tuple value type to `object?`, drop null-valued tuples before they hit the dict (matches the removed `ToArgsDictionary`''s skip-nulls behavior), added an `IntlAccessorTests` case pinning it.
2. **cref XML warnings on RichMessage doc** — the cref referenced a signature with C# keyword generics + nullability, which doesn''t resolve. Replaced with a prose reference + `<c>RichMessage</c>`.
3. **SourceRewriter ICU param-name escaping** (two comments, same finding) — now uses `SymbolDisplay.FormatLiteral(param, quote: true)` to produce a valid C# string literal so a placeholder with `\"` or `\\` can''t corrupt the emitted source.
4. **Template phrasing** — tightened `(name, value)` to `(\"name\", value)` with an explicit sentence that the first item is the placeholder name as a string literal. Regenerated `docs/guide/localization.md` from the template.

JIT localization tests: 283/283 (the +1 is the new null-skip case).

PR #401, opened 2026-05-24
feat(046): docking content types & reserved document area

codemonkeychris:
## Summary

Spec 046 is an additive amendment to spec 045 that adds an IDE-class content-type vocabulary on top of the existing `DockNode` algebra. Defaults preserve spec-045 behavior for layouts that don't opt in.

- **New surface:** `DockGroupRole` ({ `General`, `DocumentArea`, `ToolWindowStrip` }) on `DockTabGroup`; `[Flags] DockSides` mask on `ToolWindow.AllowedSides`; public `DockLayoutOps` façade (`InsertPaneAtTarget` / `RemovePane` / `MovePaneToTarget` / `FindContainer`); `DockHostModel.Dock(content, DockTabGroup, target)` overload; JSON round-trip for `role` / `allowedSides`.
- **Role-aware routing:** `Dock(content, Center)` prefers a category-matched group (Document → DocumentArea, ToolWindow → ToolWindowStrip), falls back to any accepting group with a `DockOperationLog` diagnostic when nothing matches.
- **Reserved-empty DocumentArea:** an empty `DocumentArea` survives as a visible reserved well when it's the only one in the tree; empty arms next to a non-empty sibling cull so split-drag residue collapses cleanly. `ShowWhenEmpty = true` opts out.
- **Drag-drop overlay filtering:** rejected targets dim (category mismatch or `AllowedSides` violation); `PinToSide` throws `InvalidOperationException` on a masked side.

## Bug fixes (Scene-J manual verification)

Three regressions the user surfaced while exercising the new VS-shaped layout in the showcase:

1. **Splitter cursor tracking in 3+ child splits.** Pair extent at drag capture now uses the measured `ActualSize` of leading + trailing children. The original `GetHostExtent()` (whole panel) misattributed grow space in 3+ child splits so the splitter drifted at ~67% of cursor speed.
2. **Open-new-doc-after-split lost the new doc.** Host's `layoutOverride` (shape-only tree carrying drag-modified splits) didn't invalidate when the app updated `manager.Layout`. Now tracks the leaf-key set across renders and discards the override when the set changes — preserves the override for stable Layouts and model-mutator paths (`DockHostModel.Dock` / `Hide` / `PinToSide`) where the app's Layout key set doesn't change.
3. **Close-non-last-in-split left an empty arm.** Refined prune rule: empty DocumentArea survives only when no non-empty DocumentArea exists in the tree; otherwise cull so the split collapses to the surviving sibling. The reserved-well case (all empty) still keeps one.

## Tests

- **Unit (`tests/Reactor.Tests/Docking/`):** 11 new files covering the §6.3 acceptance matrix, drop-filter, group-target resolution, JSON round-trip, role propagation through splits, floating-window role inheritance, layout-strategy interaction, and a 20-test stress suite (many docs, deep splits, mixed Doc/Tool, edge synthesis, AllowedSides + PinToSide). 613/613 docking tests pass.
- **Selftest (`tests/Reactor.AppTests.Host/SelfTest/Fixtures/NativeDockingRoleAwareFixture.cs`):** 7 host-level fixtures, 42 assertions — VS-shape open/close, reserved-well revival, 8-doc + repeated-split stress, edge-synthesis, AllowedSides mask. Specifically pins the three bug fixes above. Full `NativeDocking*` selftest run: 0 failures.

## Sample

`samples/apps/dock-showcase/App.cs` adds Scene J — Roles & AllowedSides — with the VS-shaped layout, Open / Close All / PinToSide trial buttons, and a side panel surfacing `DockOperationLog` events including splitter PRESS/MOVE/RELEASE/SOLVE traces so future cursor-tracking regressions are diagnosable without binary debugging.

## Bonus

Drive-by fix to `docs/guide/images/advanced/error-boundary-retry.png` — previous capture froze on the "Loaded." success state instead of the error/Retry UI the page documents.

## Test plan

- [ ] `dotnet build Reactor.slnx -c Debug -p:Platform=x64` — clean
- [ ] `dotnet test tests/Reactor.Tests --filter Docking` — 613/613 pass
- [ ] `dotnet run --project tests/Reactor.AppTests.Host -- --self-test --filter NativeDocking` — 0 failures
- [ ] Manual smoke on `samples/apps/dock-showcase` Scene J: splitter tracks 1:1, Open New Document after split lands in a DocumentArea, closing a doc in a split arm collapses cleanly


codemonkeychris:
## CR feedback addressed in 29040547

**Applied:**

| # | Comment | Fix |
|---|---|---|
| Copilot | `DockDropTargetOverlayControl.cs:133` unused `DisabledMaskShift` | Removed the constant. The shift loop uses `1 << (int)disabled[i]` directly so the named constant was dead. |
| Copilot | `DockLayoutSerializer.cs:449` null/empty comment wrong | Rewrote: "null (field omitted) → All; empty list → None. Symmetric with the writer above..." |
| CodeQL | `DocumentAreaStressTests.cs` unused `docs` list | Removed declaration and `docs.Add(d)`. |
| CodeQL | `RoleAwareStrategyTests.cs` dead `model` assignment | Removed the first `model =` block; kept the `PassThroughStrategy` setup. |
| CodeQL | `samples/apps/dock-showcase/App.cs:1349` foreach+if → `OfType<>` | Applied. Added `using System.Linq`. |
| CodeQL | `NativeDockingRoleAwareFixture.cs:65` foreach+if | Collapsed inner loop to `OfType<TabViewItem>().Any(...)`. |
| CodeQL | `NativeDockingRoleAwareFixture.cs:434` empty catch | Captures `ex.Message`; added a new `RoleAware_Mask_PinToSideAllowsBottom_NoError` check that asserts no exception surfaced on the allowed side. |

**Left as-is, with reasoning:**

- **`DockHostNativeComponent.cs:1311` `LeafKeysMatch` and `:1331` `SetEquals` "use Where".** Both helpers are called from the render diff path (override-staleness check, once per render). The current foreach-with-early-return is the same shape as `.All(b.ContainsKey)` but allocation-free, and this file deliberately doesn't import `System.Linq`. The bot's suggested `Where(...).Any()` is strictly more expensive than the existing loop for no readability win in a four-line method.
- **`Spec046IntegrationTests.cs:66` foreach → `.Select(...)`.** The loop body mutates `layout` with `RemovePane` and reassigns it; converting to `.Select(key => FindGroupOfRole(layout, ...).First(...))` would rely on closure-captured `layout` being re-read between deferred enumerations. It happens to work but it's a fragile pattern to leave in a regression test. Kept the explicit loop.

**Verification**

- `dotnet build Reactor.slnx -c Debug -p:Platform=x64` — clean.
- `dotnet test tests/Reactor.Tests --filter Docking` — 613/613 pass.
- `dotnet run --project tests/Reactor.AppTests.Host -- --self-test --filter NativeDocking_RoleAware` — 0 failures across all 7 fixtures, including the new `RoleAware_Mask_PinToSideAllowsBottom_NoError` assertion.

PR #415, opened 2026-05-26
feat(hosting): layered-window attrs — Opacity, NoActivate, IgnorePointerInput

codemonkeychris:
## Summary

Adds three Win32 layered-window extended-style primitives to `WindowSpec` + `ReactorWindow`, with matching runtime mutators and selftest coverage. These are foundation primitives for the upcoming spec 045 §2.6 VS-style tab tear-off, but they're independently useful (HUD overlays, drag previews, splash/loading windows that shouldn't steal focus, etc.) and decoupled here to keep the eventual docking PR smaller.

- **`Opacity`** (commit `dfb42fa9`) — window-wide alpha in [0..1] via `WS_EX_LAYERED` + `SetLayeredWindowAttributes`. `1.0` strips the layered style so opaque windows pay zero compositor overhead. Already useful standalone.
- **`NoActivate`** — applies `WS_EX_NOACTIVATE` so the window appears without stealing foreground activation. Matches VS tool-window / drag-preview behavior. Applied during chrome setup **before** `Activate` so the first show observes the flag.
- **`IgnorePointerInput`** — applies `WS_EX_TRANSPARENT` so mouse events pass **through** the window to whatever's underneath. Only effective on layered windows (i.e. paired with `Opacity < 1.0`). Lets a translucent preview sit on top of clickable UI without blocking input.

All three follow the same shape: declarative field on `WindowSpec` (applied during `ApplyChrome`) plus a public mutator on `ReactorWindow` that mirrors the new value into `_spec` so a later `Update()` diff sees the live state.

## Test plan

Selftest fixtures in `WindowModelFixtures.cs`, registered in `SelfTestFixtureRegistry.cs`:

- [x] `WindowModel_OpacityRoundTrip` — spec + mutator round-trip, asserts `WS_EX_LAYERED` bit, `1.0` strips it, out-of-range throws (6 asserts)
- [x] `WindowModel_NoActivateRoundTrip` — spec + mutator + `WS_EX_NOACTIVATE` bit (5 asserts)
- [x] `WindowModel_IgnorePointerInputRoundTrip` — spec + mutator + `WS_EX_TRANSPARENT` bit (5 asserts)

All three pass on ARM64 (16/16 asserts).

## Context

This PR is split out of a larger spec 045 §2.6 implementation (VS-style tab tear-off floating windows). A spike validated that these three primitives are necessary for the tear-off visual to coexist with WinUI's drop-target overlay, so they're being landed first to keep the eventual docking PR focused on the dock-specific changes.

### DIFF
--- a/src/Reactor/Hosting/ReactorWindow.cs
+++ b/src/Reactor/Hosting/ReactorWindow.cs
@@ -1002,12 +1002,18 @@ public sealed class ReactorWindow : IDisposable
         if (opacity >= 1.0)
         {
             // Strip WS_EX_LAYERED so the compositor fast-path is restored.
-            // No-op when it wasn't set (the common case).
-            if (isLayered)
-            {
-                nint cleared = (nint)((long)current & ~NativeOpacity.WS_EX_LAYERED);
-                _ = NativeOpacity.SetWindowLongPtr(_hwnd, NativeOpacity.GWL_EXSTYLE, cleared);
-            }
+            // Also strip WS_EX_TRANSPARENT — that style is only meaningful
+            // on layered windows, so leaving it set after un-layering would
+            // wedge the window in an inconsistent extended-style state.
+            // Mirror IgnorePointerInput=false into _spec so Update() diffs
+            // see the live state.
+            long currentBits = (long)current;
+            long strippedBits = currentBits & ~(NativeOpacity.WS_EX_LAYERED | NativeOpacity.WS_EX_TRANSPARENT);
+            if (strippedBits != currentBits)
+                _ = NativeOpacity.SetWindowLongPtr(_hwnd, NativeOpacity.GWL_EXSTYLE, (nint)strippedBits);
+            var prevSpec = Volatile.Read(ref _spec);
+            if (prevSpec.IgnorePointerInput)
+                Volatile.Write(ref _spec, prevSpec with { IgnorePointerInput = false });
             return;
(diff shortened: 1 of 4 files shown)


codemonkeychris:
Pushed `2fd80912` addressing Copilot's three findings:

**Fixed**
- `WindowSpec.Validate()` now rejects `IgnorePointerInput=true` with `Opacity>=1.0` — the silent-no-op contract is gone (Copilot finding #1).
- `ApplyOpacity` strips both `WS_EX_LAYERED` *and* `WS_EX_TRANSPARENT` when going to opaque, and mirrors `IgnorePointerInput=false` into `_spec`. No more "stuck transparent on an un-layered window" state (finding #2).
- `SetIgnorePointerInput(true)` throws `InvalidOperationException` when the window isn't layered, instead of silently setting a meaningless flag (finding #3).

Added `WindowModel_OpacityIgnorePointerInvariants` fixture (7 new asserts) covering all three paths. 23/23 selftest asserts pass on ARM64.

**Not addressed**
The `github-code-quality` bot's `LibraryImport` suggestions — kept `DllImport` to stay consistent with the existing P/Invoke style in this file (`NativeIcon`, `NativeOwnership`, `NativeDpi` all use the legacy attribute). A `DllImport → LibraryImport` migration is best done across the whole file (or the codebase) as its own change, not piecemeal in this PR.

PR #425, opened 2026-05-27
feat: one-command bootstrap (bootstrap.ps1 + mur doctor/upgrade + dotnet-tool mur)

codemonkeychris:
## Summary

Collapses the six-step Setup in `docs/guide/getting-started.md` into a single command: `./bootstrap.ps1`. Adds two new `mur` verbs (`doctor`, `upgrade`) and packs `mur` as a `dotnet tool` global install so it's on PATH cross-shell with no per-arch `$env:Path` editing.

The signed-NuGet path (spec 022) is unchanged — this is strictly the source-checkout developer flow.

- **`bootstrap.ps1`** — pre-flight → `dotnet pack` mur → `dotnet tool install/update -g` → `mur pack-local` → `dotnet new uninstall && install` template → symlink/copy `plugins/reactor`. Idempotent; flags: `-SkipPlugin`, `-SkipMurInstall`, `-Configuration`.
- **`mur doctor`** — five checks (SDK ≥ 10 / `mur` resolvability / `local-nupkgs/` currency / `reactorapp` template registration / Claude plugin). Each FAIL prints a copy-pasteable remediation.
- **`mur upgrade`** — lightweight refresh path (re-pack framework + templates, reinstall template with uninstall-first, refresh plugin). Does **not** update `mur` itself (a process can't replace its own binary mid-run) — prints the bump-mur command on completion.
- **`Reactor.Cli.csproj`** — `PackAsTool=true`, `ToolCommandName=mur`, `PackageId=Microsoft.UI.Reactor.Cli`. Legacy `bin/<arch>/` mirror layout retained for existing PATH-based installs.
- **`PackLocalCommand` / `--regen-api`** — repo-root discovery now prefers CWD before `AppContext.BaseDirectory`, so a globally-installed `mur` (living under `~/.dotnet/tools`) still resolves the checkout the user is sitting in.
- **`docs/guide/getting-started.md`** — Setup rewritten around `bootstrap.ps1`, new "After `git pull`" / "Verify the install" subsections, and a "Manual setup" section that walks through exactly what bootstrap does step-by-step (eight numbered `dotnet pack` / `dotnet tool install` / `mur pack-local` / `dotnet new install` / symlink calls — no Reactor magic).
- **`README.md`** — "Quick start" replaced its one-line guide-link with the actual `git clone` + `./bootstrap.ps1` + `dotnet new reactorapp` block plus a paragraph of what bootstrap does. Links to the manual walkthrough.

## Test plan

- [x] `dotnet build src/Reactor.Cli/Reactor.Cli.csproj -c Release -p:Platform=ARM64` succeeds with 0 warnings
- [x] `./bootstrap.ps1` ran end-to-end from a checkout with no global `mur` installed — produced both `local-nupkgs/Microsoft.UI.Reactor.0.0.0-local.nupkg` and the templates nupkg, registered `dotnet new reactorapp`, symlinked the Claude plugin
- [x] `mur --version` resolves cross-shell from `~/.dotnet/tools/mur.exe` after bootstrap
- [x] `mur doctor` reports **PASS** on all 6 checks post-bootstrap
- [ ] `dotnet new reactorapp -n MyApp && cd MyApp && dotnet run` (not yet exercised on this PR — covered by existing template tests)
- [ ] x64 host validation (verified on ARM64 only — bootstrap detects `$env:PROCESSOR_ARCHITECTURE` so x64 should follow the same path)
- [ ] `mur upgrade` smoke-tested (the moving parts are exercised separately by `pack-local` + `dotnet new install`, but the verb itself wasn't run end-to-end)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,21 @@ Many of the experiments in this repo — the charting stack, accessibility valid
 
 ## Quick start
 
-See **[docs/guide/getting-started.md](docs/guide/getting-started.md)** for the full walkthrough — prerequisites, building the framework and `mur` CLI from source, scaffolding your first app, and running through hello-world, a todo list, and a calculator with screenshots at each step.
+Reactor doesn't ship a signed NuGet yet, so you build the framework, the `mur` CLI, and the project template from source. `bootstrap.ps1` collapses that into a single command — ~3 minutes per machine.
 
+```powershell
+git clone https://github.com/microsoft/microsoft-ui-reactor.git
+cd microsoft-ui-reactor
+./bootstrap.ps1
 
+dotnet new reactorapp -n MyApp
+cd MyApp
+dotnet run
+```
+
+`bootstrap.ps1` packs `mur` as a `dotnet tool` global install (cross-shell PATH, no per-arch `$env:Path` edits), packs the framework + project templates into `local-nupkgs/`, registers the `dotnet new reactorapp` template, and installs the Reactor agent plugin under `~/.claude/plugins/reactor`. Re-run it (or `mur upgrade` for a lighter refresh) after `git pull`. Verify a working install with `mur doctor`.
+
+Prefer to wire it up by hand? **[docs/guide/getting-started.md](docs/guide/getting-started.md#manual-setup)** has a no-magic walkthrough of the exact `dotnet pack` / `dotnet tool install` / `dotnet new install` calls `bootstrap.ps1` makes, plus the full hello-world → todo → calculator tour.
 
 ---
 
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -73,10 +73,14 @@ foreach ($line in $sdkOutput) {
     }
 }
 if (-not $has10OrLater) {
-    Write-Host '    [warn] .NET 10 SDK not detected — Reactor requires 10+. https://dotnet.microsoft.com/download' -ForegroundColor Yellow
-} else {
-    Write-Ok ".NET SDK present"
+    Fail @"
+.NET 10+ SDK not detected — Reactor requires 10 or later.
+Installed SDKs:
+$($sdkOutput -join "`n")
+Install the latest .NET SDK from https://dotnet.microsoft.com/download and re-run ./bootstrap.ps1.
+"@
 }
+Write-Ok ".NET SDK present"
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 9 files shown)


codemonkeychris:
Pushed `95cfc144` to address CI + Copilot review.

**CI**

Both `Docs compile` and `Docs tier-drift` failed with the same root cause: `REACTOR_DOC_TIER_005` — "no reference table found in the first half of the page". Adding the Manual setup section pushed the existing element-reference table past the rendered guide's midpoint. Fixed by adding an 8-row summary table (`step → what it produces`) at the top of the Manual setup section. Locally, `mur docs check-tier --topic getting-started` is now clean apart from the pre-existing W001 winui-ref warning shared by 24 other templates.

The XLINK warnings on `Next Steps` (UseEffect/UseMemo unlinked) at line 752 were existing main-branch state and are non-fatal (CompileCommand only fails on tier-lint errors in `--ci`, not on cross-link warnings).

**Copilot review (6 points)**

1. **DoctorCommand exit-code contract mismatch** — fixed. Header now says "exits non-zero only when there are FAILs — WARNs still exit 0".
2. **"Checks (in order)" list doesn't match emitted checks** — fixed. Expanded from 5 to 7 entries (framework nupkg + template nupkg are separate checks; repo-checkout discovery has its own warn behavior; `dotnet new` template check runs unconditionally).
3. **Misleading "skipping ... template checks" warn message** — fixed. Now reads "skipping local-feed checks (the `dotnet new` template check below still runs)".
4. **`RunDotnet` Process.Start failure handling** — fixed. Wrapped `Process.Start` in try/catch + handle `null` return; prints a clear "verify .NET 10+ on PATH" remediation and returns 1 (or 0 for the best-effort `dotnet new uninstall` path that uses `ignoreExitCode: true`).
5. **bootstrap.ps1 missing .NET 10 → only warns** — fixed. Now `Fail`s with the actual `dotnet --list-sdks` output, so users get a single clear prerequisite error instead of a downstream `dotnet pack` failure.
6. **No unit tests for doctor/upgrade verbs** — acknowledged, deferred. The doctor/upgrade surfaces shell out to `dotnet`/`Process.Start` and walk `~/.dotnet/tools` / `~/.claude/plugins/` — testing them in the same shape as `CleanLocalCommandTests` would require either heavy fakes for the process boundary or a temp-dir test fixture. Worth doing, but bigger than this PR's scope. Tracking it for a follow-up.

PR #430, opened 2026-05-27
fix: clean-machine bootstrap (analyzer pack path) + winget auto-install + CI workflow

codemonkeychris:
## Summary

User-reported failure on a freshly-cloned machine: `mur pack-local` blew up with `NuGet.Build.Tasks.Pack.targets(222,5): error Could not find a part of the path '...\src\Reactor.Analyzers\bin\Debug\netstandard2.0\'`. This PR fixes the underlying packaging bug, hardens `bootstrap.ps1` so missing prerequisites get winget-installed instead of just being reported, and adds a CI workflow that would have caught the regression.

### Three commits, three concerns

1. **`fix(pack): analyzer DLL path on clean machines`** — `src/Reactor/Reactor.csproj` packs the analyzer + localization-generator DLLs via `<None Include="...\bin\$(Configuration)\netstandard2.0\..." />` (no `$(Platform)` segment). Both projects are `<ProjectReference>`s, so when `dotnet pack -p:Platform=x64` runs, the inherited Platform cascades to the transitive analyzer build and the output lands in `bin\x64\Debug\netstandard2.0\` instead. Pack reads the platform-less path, doesn't find it, dies. Masked on dev boxes by any prior AnyCPU build leaving a stale DLL behind. **Fix:** set `<AppendPlatformToOutputPath>false</AppendPlatformToOutputPath>` on both analyzer csprojs — they're netstandard2.0 IL with no platform-specific code, so output should always land at `bin\$(Configuration)\netstandard2.0\` regardless of inherited Platform. Reproduced + verified locally on a wiped bin/obj tree.

2. **`feat(bootstrap): winget auto-install for .NET 10 + optional WinAppSDK prompt`**
   - `Install-WithWinget` helper: runs winget with `--silent --disable-interactivity --accept-source-agreements --accept-package-agreements`, treats already-installed exit `-1978335189` as success, and rebuilds `$env:Path` from Machine + User registry strings so freshly-installed binaries resolve in the same shell.
   - **.NET 10 SDK** missing → dump installed-SDK list, `winget install Microsoft.DotNet.SDK.10`, re-probe. Hard-fail if winget itself isn't on PATH (App Installer is an OS-level prereq this script doesn't try to repair).
   - **Windows App SDK** is tri-state via `-InstallWinAppSdk`. Unspecified → prompt the dev (default no) since the framework defaults to `WindowsAppSDKSelfContained=true` and the machine runtime is only needed for framework-dependent deployment. `-InstallWinAppSdk` → force-install non-interactively (for CI / fresh-dev-box automation). `-InstallWinAppSdk:$false` → skip silently. Preserves the existing self-contained happy path (no surprise 50MB install on every clean run) while making framework-dependent deployment one keystroke away.

3. **`ci: bootstrap.yml — exercise bootstrap.ps1 end-to-end on a clean runner`** — new Windows-only workflow scoped to PRs touching `bootstrap.ps1`, the CLI source, the templates project, `Directory.Build.props`, or the workflow itself. Pipeline: pre-flight diagnostic dump → bootstrap.ps1 non-interactively → `mur --version` from a fresh shell → `mur doctor` → verify `local-nupkgs/` content → `dotnet new list reactorapp` → scaffold + restore + build a TestApp → `mur upgrade --skip-plugin` → re-run bootstrap.ps1 for idempotence. Deliberately omits `actions/setup-dotnet` so bootstrap.ps1 itself is responsible for the SDK install path.

## Test plan

- [x] Reproduced the clean-machine failure locally by wiping `src/Reactor.Analyzers/{bin,obj}` and `src/Reactor.Localization.Generator/{bin,obj}` and `src/Reactor/{bin,obj}` — `mur pack-local` failed the same way as in the user's screenshot before the csproj fix, succeeded after.
- [x] Verified post-fix DLL layout: both analyzer DLLs land at `bin/Debug/netstandard2.0/Reactor.Analyzers.dll` (no `x64`/`ARM64` segment).
- [ ] Watch the new `bootstrap.yml` workflow fire on this PR and confirm green end-to-end (will show in CI; please block merge until that goes green).
- [ ] x64 / framework-dependent deployment paths: not exercised in this PR (the workflow does `-c Release` build only). Worth adding `dotnet publish -p:WindowsAppSDKSelfContained=false` if we want the runtime-install path also gated.

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -164,8 +164,14 @@ jobs:
       - name: Build TestApp (default WindowsAppSDKSelfContained=true)
         shell: pwsh
         run: |
+          # WindowsAppSDKSelfContained=true (inherited from
+          # Directory.Build.props) requires a concrete arch to embed the
+          # runtime under — AnyCPU is rejected by the SelfContained
+          # target. Pass the host arch explicitly.
+          $arch = if ($env:PROCESSOR_ARCHITECTURE -eq 'ARM64') { 'ARM64' } else { 'x64' }
           dotnet build TestProjects/TestApp/TestApp.csproj `
               -c Release `
+              "-p:Platform=$arch" `
               --nologo -v:m
           if ($LASTEXITCODE -ne 0) { throw "TestApp build exited $LASTEXITCODE" }
 
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -47,6 +47,10 @@ jobs:
       - name: Pre-flight — show toolchain baseline
         shell: pwsh
         run: |
+          # Diagnostic-only step. Reset $LASTEXITCODE on exit so a non-zero
+          # winget status (which is common — "not installed" returns
+          # -1978335166) doesn't fail the workflow before bootstrap.ps1 even
+          # starts.
           Write-Host "=== dotnet --list-sdks ==="
           if (Get-Command dotnet -ErrorAction SilentlyContinue) {
               dotnet --list-sdks
@@ -58,8 +62,12 @@ jobs:
           winget --version
           Write-Host ""
           Write-Host "=== Windows App Runtime 2.0 (winget list) ==="
-          winget list --id Microsoft.WindowsAppRuntime.2.0 --exact 2>$null
+          # --accept-source-agreements: a fresh runner hasn't accepted the
+          # msstore source terms. Without it, winget prompts and fails on a
+          # non-interactive shell.
+          winget list --id Microsoft.WindowsAppRuntime.2.0 --exact --accept-source-agreements 2>&1
           Write-Host "exit=$LASTEXITCODE"
+          $global:LASTEXITCODE = 0
 
       - name: Run bootstrap.ps1 (non-interactive)
         shell: pwsh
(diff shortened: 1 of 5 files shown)


codemonkeychris:
Rebased onto current `main` (`dac72fe3`) to pick up @clintrutkas's #429 (Improving onboarding) which landed earlier today and touched the same two files.

**Conflict resolution:**
- **`bootstrap.ps1`** — #429's inline `winget install Microsoft.DotNet.SDK.10` + 2-line PATH refresh is a strict subset of this branch's `Install-WithWinget` helper + `Test-DotnetSdk10` probe (which additionally handles the "dotnet present but pre-10" case and centralizes the Machine+User PATH rebuild for all winget calls). Resolved by keeping this branch's version.
- **`README.md`** — #429's `& (Get-Process -Id $PID).Path -ExecutionPolicy Bypass -File .\bootstrap.ps1` invocation was preserved (this branch doesn't touch README, so it fast-forwarded cleanly).

`bootstrap.ps1` parse-checked OK after resolution. Force-pushed.

The new `bootstrap.ps1 (windows-latest)` workflow ([run 26524266103](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/26524266103)) is firing on this PR — which validates both the original pack fix (the analyzer DLL would land in the wrong directory on the clean runner without the csproj change) and the bootstrap auto-install path (windows-latest doesn't have WindowsAppRuntime 2.0 preinstalled). Will report back when it completes.

PR #431, opened 2026-05-27
fix(stress): per-fixture hang watchdog + WaitForIdle cap raise

codemonkeychris:
## Summary

Two race-driven flakes surfaced in the recent 500x selftest stress run on `main` (CI Stress run 26513480330): 4 of 10000 iterations failed (~99.96% pass), with 2 of 4 in `NativeDocking_*` fixtures. Investigation showed both root causes were in **test infrastructure**, not the new docking window manager work (#418/#420/#421/#415).

### Root causes

1. **`ReactorHost.WaitForIdleAsync(maxYields: 10)`** silently returned after 10 low-priority dispatcher yields whether or not the render loop had actually settled. Under CI VM contention, callers like `Harness.Render()` moved on against a half-rendered tree — producing flakes like `NativeDocking_RoleAware_SplitDocArea_CloseNonLast` failing to `FindText("body:d1")` on iteration 3/500.

2. **`SelfTestRunner.HangWatchdogLoop`** used a global 60 s constant equal to `EventSubscriptionLeakBaseline`'s own `FixtureTimeout`. On worst-case CI ticks (~60 s for 200 renders + 200 reconciles, as the fixture's own comment acknowledged), the two raced — the watchdog won and FailFasted the shard instead of letting the fixture's graceful timeout fire.

### Changes

| File | Change |
|---|---|
| `src/Reactor/Hosting/ReactorHost.cs` | `WaitForIdleAsync` default cap 10 → 50; `Debug.WriteLine` when the cap actually hits so the next flake leaves a greppable trail |
| `tests/.../SelfTest/SelfTestRunner.cs` | `FixtureProgress` carries a per-fixture `HangThreshold = max(60 s, FixtureTimeout + 30 s)`; watchdog uses that, not the global constant |
| `tests/.../SelfTest/SelfTestFixtureBase.cs` | doc comment updated to reflect new watchdog rule |
| `tests/.../Fixtures/NativeDockingReliabilityFixture.cs` | `EventSubscriptionLeakBaseline` `FixtureTimeout` 60 → 120 s; heartbeat `H.Check` after warmup and every 25 cycles |

The heartbeat `H.Check` calls don't reset the watchdog (it's still elapsed-since-fixture-start), but they make `ok` lines visible in the log so a future hang reveals which 25-cycle window it lived in, not just "no progress".

## Test plan

- [x] `src/Reactor` and `tests/Reactor.AppTests.Host` both compile clean (0 errors, only pre-existing XML-comment / nullability warnings)
- [ ] Kick off a 1000x CI Stress run targeting selftests on this branch
- [ ] Confirm no flakes in the `NativeDocking_*` fixtures or `DataGrid_KeyboardAndPrivateRenderPaths`
- [ ] Confirm `EventSubscriptionLeakBaseline` per-iteration time stays well under the 120 s budget (target: still ~15 s local, ~30-60 s under CI contention)
- [ ] Spot-check stress logs for `[Reactor.WaitForIdle] yield cap hit` lines — if any appear, the 50-yield cap is being approached and we'll want to investigate the underlying render loop

### DIFF
--- a/src/Reactor/Hosting/ReactorHost.cs
+++ b/src/Reactor/Hosting/ReactorHost.cs
@@ -881,7 +881,7 @@ public sealed class ReactorHost : IDisposable
     /// RenderLoop callbacks and Low-priority re-renders all complete before returning.
     /// Used by test harnesses to replace blind Task.Delay waits.
     /// </summary>
-    public Task WaitForIdleAsync(int maxYields = 10)
+    public Task WaitForIdleAsync(int maxYields = 50)
     {
         if (_disposed) return Task.CompletedTask;
         if (_renderPending == 0 && !_isRendering && !_needsRerender)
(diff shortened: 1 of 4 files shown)


codemonkeychris:
Stress run summary for https://github.com/microsoft/microsoft-ui-reactor/actions/runs/26526776797:

- Overall selftest reliability: **997/1000 iterations passed = 99.7%**.
- Scope: **1000 selftest iterations across 40 shards**; unit/integration shards were skipped.
- Docking-window-specific signal: **1 docking-related failure in 1000 iterations**, or roughly **99.9% docking iteration reliability** for this run.

Failed tests observed during the run:

| Shard / Iteration | Test | Failure |
|---|---|---|
| Shard 5 / iteration 1 | `DataGrid_ScrollPopulatesData` | `ScrollPop_InitialRender - assertion failed` |
| Shard 12 / iteration 17 | `AsyncResource.Framerate.DataGridScroll` | `exceeded 30s` |
| Shard 40 / iteration 25 | `NativeDocking_A11y_HostLandmarkAndPaneAutomationIds` | `A11y_PaneAutomationId_ActiveTabFound - assertion failed` |

PR #454, opened 2026-05-29
fix(navigationview): reconcile menu items in place to preserve expansion

codemonkeychris:
## Problem

`NavigationView` with hierarchical menu items mis-behaved on every interaction:

- clicking a category expanded it then immediately collapsed it (a flash);
- clicking a child item collapsed its parent;
- selection state churned on each render.

## Root cause

`UpdateNavigationView` (and the V1 `NavigationViewDescriptor`) did
`MenuItems.Clear()` + full rebuild whenever the `MenuItems` array reference
changed. Consumers typically build a fresh `MenuItems` array each `Render()`
(e.g. via LINQ), so the `ReferenceEquals` guard never held and the rebuild
fired on essentially every re-render.

Per-item expansion lives only on the live `NavigationViewItem` — there is no
`IsExpanded` on `NavigationViewItemData` — so recreating the containers reset
expansion to `false` every time. Because selecting an item triggers a
re-render, the just-expanded category (or a child's parent) snapped shut on the
same click.

## Fix

Reconcile menu items **in place** instead of clear-and-rebuild:

- match existing `NavigationViewItem`s by `Tag`;
- update only changed `Content` / `Icon`;
- recurse into children;
- reuse the container so its runtime `IsExpanded` (and selection) survive the
  re-render.

A fast path skips all detaching when the item structure is unchanged (the
common case). Mirrored into the V1 `NavigationViewDescriptor` so V1 ON behaves
identically to V1 OFF. Mount-time selection now recurses so a child tag selects
correctly.

## Tests

- New E2E guard: `NavigationView_Hierarchical` host fixture +
  `NavigationViewInteractionTests`. Verified to **fail on the pre-fix build**
  (expanded category collapses after the selection re-render) and **pass after**.
- Existing NavigationView self-tests, including the V1
  `Desc_NavigationView_MountUpdate` coverage, pass (0 failures).
- Also adds a TreeView E2E guard exercising the Gallery's basic text-tree path.

## Out of scope

Compact (icon-only) pane interaction is deferred to a follow-up.

### DIFF
--- a/src/Reactor/Core/Reconciler.Mount.cs
+++ b/src/Reactor/Core/Reconciler.Mount.cs
@@ -1450,10 +1450,7 @@ public sealed partial class Reconciler
         }
         if (nav.Content is not null) nv.Content = Mount(nav.Content, requestRerender);
         if (nav.SelectedTag is not null)
-        {
-            foreach (var mi in nv.MenuItems.OfType<WinUI.NavigationViewItem>())
-                if (mi.Tag as string == nav.SelectedTag) { nv.SelectedItem = mi; break; }
-        }
+            nv.SelectedItem = FindNavItemByTag(nv.MenuItems, nav.SelectedTag);
         SetElementTag(nv, nav);
         if (nav.OnSelectedTagChanged is not null)
             nv.SelectionChanged += (s, args) =>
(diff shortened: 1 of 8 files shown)


codemonkeychris:
Addressed the review feedback in dd7ce549:

- **Duplicate-key reuse (correctness):** the slow-path rebuild now consumes the reuse-map entry via `Dictionary.Remove(key, out …)` when a container is reused. Duplicate sibling keys — duplicate `Tag`s, or duplicate `Content` when no `Tag` is set — now fall through to a freshly created container instead of adding the same WinUI item to the collection twice. Applied to both the legacy arm and the V1 descriptor.
- **Explicit filtering:** the snapshot loops now filter explicitly with `.OfType<…>()` / `.Where(…)`.

Verified: NavigationView self-tests (incl. `Desc_NavigationView_MountUpdate`) pass with 0 failures, and the `NavigationViewInteractionTests` E2E guard still passes (and still fails on the pre-fix build).

PR #468, opened 2026-05-30
perf(reconciler): gate V1 ReactorState tagging on HasCallbacks (spec 047 §4.4 follow-up)

codemonkeychris:
## What

Gate `V1HandlerAdapter`'s `ReactorState` attached-DP write on `Element.HasCallbacks`, restoring the legacy reconciler's allocation discipline for callback-free leaves (TextBlock / Border / StackPanel / Image — the bulk of any real tree).

## Why

Spec-047 §4.5 collapsed all built-in controls onto the V1 dispatch adapter. The adapter tagged **every** mounted control with a `ReactorState` (one allocation + one attached-DP write per mount), where the legacy hand-coded reconciler had only tagged callback-bearing elements (see `ChildReconciler.cs:91` and `Reconciler.Update.cs:83`). That regressed the bytes-per-render of the most common element in any real tree — a callback-free display leaf — by **+21.6%**.

Routed-input modifiers (`.OnPointerPressed` etc.) dispatch through `ModifierEventHandlerState.Current*` slots refreshed each render by `ApplyEventHandlers` and do **not** resolve via the tag, so they are intentionally outside the gate. Implicit composition animations registered by `ApplyPropertyAnimation` continue to fire on direct visual writes, so `.Animate()` on callback-free leaves still works without a tag.

## Change

New `Reconciler.SetElementTagIfNeeded(FrameworkElement, Element)` helper:

- **Cheap-refreshes** the live `Element` on an existing `ReactorState` (callback-bearing controls, pooled reuse, echo/setter scope) — no alloc.
- **Skips** the allocation + DP write when the element has no callbacks and the control has no state yet.

`V1HandlerAdapter.Mount` / `Update` now call `SetElementTagIfNeeded` instead of `SetElementTag`. The public `Reconciler.SetElementTag` API is unchanged.

## Perf results

5000 iterations on the same dev box that captured the pre-V1 baseline (bytes/render = `allocBytes ÷ 5000`):

| Bench | Pre-V1 (legacy) | Post-migration V1 | **Gated V1 (this PR)** |
|---|---:|---:|---:|
| M1 Mount_Leaf_NoCallback | 1,071 | 1,302 **(+21.6%)** ⬅ the regression | **980 (−8.5% vs pre-V1)** |
| M3 Mount_Leaf_ThreeCallbacks | 8,953 | 8,461 (−5.5%) | 8,428 (−5.9%) |
| M8 Update_OneLeafChanged | 425 | 362 (−14.8%) | **310 (−27% vs pre-V1)** |

**Read-out**

- **M1** is the headline. Callback-free leaf went 1,071 → 1,302 B/render post-migration; this PR brings it to 980 B/render — below pre-V1, because §4.4's `ElementExtras` bucketing genuinely shrank the element once the tagging tax is removed.
- **M3** is unchanged: Button has callbacks → tagged identically before and after. This is the correctness signal — the gate changes nothing for callback-bearing controls.
- **M8** improved further — fewer ReactorState refreshes on the no-op skip path.

> **Note — separate, larger regression on M12 not addressed here.**
> M12 `Pool_Rent_HotPath` sits at ~1,238 B/render against a pre-V1 baseline of 218 (~5× blow-up). This PR's gate partially helps (1,238 → ~980) but doesn't close the gap. That's a TextBlock pooling regression — unrelated mechanism, scoped to its own follow-up.

## Validation

**Selftest suite (982 fixtures):** 1 pre-existing flake (`GridViewLazy_RealizedCount=252_LessThanHalf`) on both baseline and this branch — no new failures introduced.

**Targeted fixtures pass:**

- `Trampoline_*` — event dispatch / handler rebind after rerender
- `EventStateSplit_*` — `ModifierEventHandlerState` allocation discipline (verifies state IS allocated for callback-bearing controls)
- `EchoSuppress_*` — value-controlled echo round-trips
- `Modifier_*` — routed-input handlers
- `Selection_*` — selection events
- `Pool_*` (incl. `ExtProof_Pool_TagSet`) — pool rent/return tag contract (uses public `SetElementTag` API, unchanged)
- `ValueChange_*`, `TextBox_*` — text/value callbacks

## Risk

Low. The gate exactly mirrors the legacy reconciler's `HasCallbacks` discipline (`ChildReconciler.cs:91`, `Reconciler.Update.cs:83`), which has shipped for the entire history of the codebase. The set of code paths that read `ReactorState.Element` for callback-free controls is empty: routed-input dispatches via `Current*` slots, implicit animations fire on visual writes, and `GetListState` / `GetItemViewSource` / `ChangeEchoSuppressor` are only installed on controls whose elements have callbacks.

---

### DIFF
--- a/src/Reactor/Core/Reconciler.cs
+++ b/src/Reactor/Core/Reconciler.cs
@@ -712,24 +712,47 @@ public sealed partial class Reconciler : IDisposable
     /// Spec 047 §4.4 follow-up — allocation-gated element tagging for the V1
     /// dispatch adapter. Refreshes an existing <see cref="ReactorState"/>'s
     /// element pointer cheaply, but only <em>allocates</em> a new
-    /// <see cref="ReactorState"/> (+ attached-DP write) for elements that
-    /// actually wire a control-intrinsic event trampoline
-    /// (<see cref="Element.HasCallbacks"/>).
+    /// <see cref="ReactorState"/> (+ attached-DP write) for elements whose
+    /// downstream code actually reads the element back via
+    /// <see cref="GetElementTag(FrameworkElement)"/>.
     ///
-    /// <para>Callback-free display leaves (TextBlock / Border / StackPanel /
-    /// Image — the bulk of a real tree) never dispatch into Reactor code, so
-    /// they don't need the tag. This mirrors the legacy reconciler, which
-    /// gated its tag refresh on <c>HasCallbacks</c> in the
-    /// <see cref="Update"/> skip short-circuit and left leaf controls
-    /// untagged. The pre-§4.5 V1 adapter tagged <em>every</em> control
-    /// unconditionally, allocating a ReactorState per leaf mount — the
-    /// per-render byte regression this restores the savings for.</para>
+    /// <para>Three things require a live element pointer on the control:</para>
+    /// <list type="bullet">
+    ///   <item>
+    ///     <b>Event trampolines</b> — <see cref="Element.HasCallbacks"/>.
+    ///     Routed-input modifiers (<c>.OnPointerPressed</c> etc.) dispatch
+    ///     through <see cref="ModifierEventHandlerState"/>'s <c>Current*</c>
+    ///     fields refreshed each render by <c>ApplyEventHandlers</c> and do
+    ///     <em>not</em> resolve via the tag, so they are intentionally not
+    ///     part of the gate.
+    ///   </item>
+    ///   <item>
+    ///     <b>Keyed child reconciliation</b> — <see cref="Element.Key"/>.
+    ///     <see cref="ChildReconciler"/> reads the child's tag to extract the
+    ///     stable key during keyed reorders (see <c>ChildReconciler.cs:303</c>
+    ///     and <c>:392</c>). A keyed callback-free element (e.g.
+    ///     <c>Border(...).WithKey("a")</c>) loses its identity across reorders
+    ///     without the tag.
+    ///   </item>
+    ///   <item>
+    ///     <b>Element-extras readback</b> — <see cref="Element.Extensions"/>.
+    ///     Several bucketed extras have engine paths that read the live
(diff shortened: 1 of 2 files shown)


codemonkeychris:
Pushed `23c32828` to address the CI failure + Copilot review.

## Root cause

The CI failure (`AAF_FlexColumn_MoveUnderAnimate_AttachesImplicitOffset` in both Selftests and AOT Selftests) and the Copilot review finding are the **same class of bug**: the `HasCallbacks`-only gate missed two readback paths the legacy reconciler implicitly preserved by tagging every control.

1. **Keyed child reconciliation** — `ChildReconciler.cs:303` and `:392` read the child's tag to extract the stable `Key` during keyed reorders. A callback-free keyed element (`Border(...).WithKey("a")` — the fixture's exact shape) lost identity, so the keyed Move never ran and the implicit Offset animation was never attached. Explains the CI failure cleanly.
2. **Element-extras readback** — as @copilot-pull-request-reviewer noted, several `Element.Extensions` features (`ElementTransition`, `ConnectedAnimationKey`, `InteractionStates`, `KeyframeAnimations`, …) have engine paths that read the live element via the tag.

## Fix

Extend the gate to `HasCallbacks || Key != null || Extensions != null`. The `Extensions != null` arm is a coarse but robust catch-all — the bucket being null is spec 047 §4.4's "truly inert leaf" signal, so any element with bucketed extras pays the tag rather than risk a future readback regression.

## Perf preserved

The M1 benchmark target (bare `TextBlock("hello")`) has **no callbacks, no key, no extras** → gate still applies → 980 B/render savings preserved. M3 (Button with callbacks) and M8 (bare leaf update) are likewise unaffected.

## Validation

Full selftest suite (991 fixtures) — 1 pre-existing flake (`GridViewLazy_RealizedCount=252_LessThanHalf`, also fails on baseline `main`, unrelated). The keyed-reorder fixture that regressed in CI now passes 3/3 in isolation and clean in the full run.

PR #471, opened 2026-05-30
fix(bootstrap): skip template uninstall on first run + add -Verbose diagnostics

codemonkeychris:
## Problem

Running `./bootstrap.ps1` on a fresh checkout fails at step 5 (template install):

```
==> Installing `dotnet new reactorapp` template
dotnet.exe : The template package 'Microsoft.UI.Reactor.ProjectTemplates' is not found.
At C:\code\microsoft-ui-reactor\bootstrap.ps1:267 char:1
+ & dotnet new uninstall Microsoft.UI.Reactor.ProjectTemplates 2>$null  ...
    + CategoryInfo          : NotSpecified: (The template pa...' is not found.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

## Root cause

The script preemptively uninstalls the template (so a same-version repack wins over the cached copy):

```powershell
& dotnet new uninstall Microsoft.UI.Reactor.ProjectTemplates 2>$null | Out-Null
```

The `2>$null` was meant to swallow the first-run "not installed" error, but it only redirects stderr. In **PowerShell 7.4+** the `$PSNativeCommandUseErrorActionPreference` preference defaults to `true`, so the non-zero exit code from `dotnet` becomes a terminating `NativeCommandError` — and the script's `$ErrorActionPreference = ''Stop''` then halts execution. The stderr redirect can''t intercept that path.

## Fix

Probe `dotnet new uninstall` (no args, which lists installed packages) first, and only run the uninstall when our package id is actually present:

```powershell
$installedTemplates = & dotnet new uninstall 2>&1 | Out-String
if ($installedTemplates -match ''Microsoft\.UI\.Reactor\.ProjectTemplates'') {
    & dotnet new uninstall Microsoft.UI.Reactor.ProjectTemplates | Out-Null
    if ($LASTEXITCODE -ne 0) { Fail ''`dotnet new uninstall` failed'' }
}
```

Silent on first run, idempotent on subsequent runs. Comment explains the PS 7.4 trap so it doesn''t get re-broken.

## Bonus: `-Verbose` diagnostics

The script already declares `[CmdletBinding()]`, so `-Verbose` is a free common parameter — no need to invent a custom flag. Added `Write-Verbose` checkpoints at every interesting decision point so future install issues are easier to debug without re-instrumenting:

- PS version + OS arch + repo path at startup
- Each `dotnet --list-sdks` line
- winget probe command + exit code (and full output under `-Verbose`)
- Host arch resolution + `dotnet pack` invocation
- Global tool `install` vs `update` branch
- `mur` resolution path (installed vs source fallback)
- Installed-template listing + uninstall decision (first-run skip vs cleanup)
- Plugin source/dest paths + symlink-vs-copy fallback reason
- PATH refresh after winget

Output that was previously sent to `Out-Null` (winget probe, `dotnet new uninstall`) is surfaced when `-Verbose` is on, so a developer can see *which* templates / packages the script actually found.

Doc block updated with a `.PARAMETER Verbose` section and a `./bootstrap.ps1 -Verbose` example.

## Verification

- Script parses clean (`[System.Management.Automation.Language.Parser]::ParseFile`)
- `Get-Command .\bootstrap.ps1` confirms `-Verbose` is bound via CmdletBinding
- End-to-end smoke of the template block in both quiet and `-Verbose` modes:
  - **Quiet / first-run (template not installed)**: skipped uninstall silently, `LASTEXITCODE = 0` — bug fixed
  - **Quiet / re-run (template installed)**: detected, uninstalled cleanly, `LASTEXITCODE = 0`
  - **`-Verbose`**: emits `VERBOSE:` lines at every checkpoint and dumps the raw template listing

Documentation/script-only change — no test suite needs to run.

### DIFF
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -33,6 +33,14 @@
     non-interactive scripts that explicitly don't want the runtime
     installed. Mutually exclusive with -InstallWinAppSdk.
 
+.PARAMETER Verbose
+    Common parameter (enabled by [CmdletBinding]). Surfaces extra
+    diagnostic output at every decision point: detected SDK list,
+    winget probe exit codes, host arch / pack args, global-tool
+    install-vs-update branch, template uninstall decision, plugin
+    symlink-vs-copy fallback, etc. Use this when bootstrap fails or
+    behaves unexpectedly and you want to see *why* a branch was taken.
+
 .EXAMPLE
     ./bootstrap.ps1
     Full bootstrap (prompts before installing WindowsAppRuntime).
@@ -45,6 +53,11 @@
     ./bootstrap.ps1 -InstallWinAppSdk -SkipPlugin
     Non-interactive: install everything (incl. WindowsAppRuntime) and
     skip the agent plugin. Suitable for CI / fresh-dev-box automation.
+
+.EXAMPLE
+    ./bootstrap.ps1 -Verbose
+    Print extra `VERBOSE:` diagnostics at every decision point. Useful
+    for debugging install failures or unexpected branch behavior.
 #>
 [CmdletBinding()]
 param(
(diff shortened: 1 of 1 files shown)


codemonkeychris:
Good catch — pushed eec69748 which honors the documented short-circuit:

- **Quiet path**: streams `dotnet new uninstall` through `Select-String -SimpleMatch -Quiet` to produce a `[bool]` without materializing the listing.
- **`-Verbose` path**: still captures via `Out-String` because we need both the line-by-line `templates> …` breadcrumbs and the substring test.

Verified all three combinations locally (quiet/not-installed, verbose/not-installed, quiet/installed) — all exit 0 with the expected branch.

PR #483, opened 2026-05-31
Improve hot reload robustness with state migration (spec 049)

codemonkeychris:
## Summary

Implements spec [049 — improvements to hot reload](docs/specs/tasks/049-improvements-to-hot-reload-implementation.md). Makes `dotnet watch` hot reload **preserve live component state** across edits instead of resetting on every recompile.

## What changed

| Path | Behavior |
|---|---|
| Edit a value-type/record stored in a hook | **Preserved** — migrated field-by-field; unmappable fields dropped (logged) |
| Rename a component type / change identity | **Preserved** — subtree migrated onto the new instance, hook state + WinUI controls kept |
| Add / remove / reorder hook calls | **Reset** — hook list cleared and re-mounted fresh (graceful, no crash) |

- **Phase 1 — hook-order recovery**: re-mounts hooks cleanly on shape change instead of corrupting hook order.
- **Phase 2 — state migration** (`ReactorHotReloadCopier`): by-name field copy with cycle guard; block-lists native/UI handles (IntPtr, Compositor, Visual, UIElement…).
- **Phase 3 — subtree migration** (`TryHotReloadMigrateComponent`): preserves hook state + native controls across component-identity changes.
- **Devtools**: per-hook `Migrated` flag; new `HotReloadStateMigrated` ETW event.
- **Escape hatch**: `HotReloadService.ResetAllContexts()` for a clean "lose everything, remount fresh" reload.
- **NativeAOT-safe**: every migration branch is gated on `IsHotReloadLive`/`WithinUpdatePass` and trims away when `MetadataUpdater.IsSupported == false`.
- **Watch fix**: `Minesweeper.csproj` defaults a host RID so `dotnet watch` ENC doesn't fail the Windows App SDK self-contained architecture check.
- **Docs**: rewrote the dev-tooling hot-reload section (supported-edit matrix, what-survives, effects stale-closure caveat, escape hatch, AOT note).

## Testing

- `dotnet build Reactor.slnx -p:Platform=x64` → **0 errors** (core lib trim/AOT-clean).
- HotReload xUnit suite → **17/17 pass**.
- Live WinUI selftests → `HotReload_ChildHookOrderRecovery` **6/6** + `HotReload_ComponentMigratesState` **7/7**.

## Deferred (annotated in the spec)

Element-record migration path, `IElementHandler.CanHotReloadMigrate` descriptor veto, and the NativeAOT publish smoke test are documented as out-of-scope follow-ups — the headless `Phase3_*` xUnit tests are superseded by the live-WinUI selftest (class components throw `COMException` headlessly).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,52 @@ Conventions for contributors:
 
 ### Added
 
+- **Hot reload: tree-wide hook-order recovery (spec 049 §5, Phase 1).**
+  Editing a **non-root** component under .NET Hot Reload to add, remove,
+  or reorder a hook now recovers in a single re-render instead of replacing
+  that component's subtree with the render-error fallback. The reconciler,
+  while inside a hot-reload pass, catches the resulting `HookOrderException`,
+  resets just that child's hook state, and re-renders it once — so the edit
+  applies and sibling/descendant state is preserved. Steady-state
+  (non-hot-reload) rendering is byte-for-byte unchanged: the new recovery arm
+  is gated on a `HotReloadService.WithinUpdatePass` filter that is only true
+  during a hot-reload-triggered render. Also fixes a latent effect-cleanup
+  leak where a staged `PendingCleanup` (from an effect whose deps changed in
+  the same render that later threw) was not drained on context teardown.
+
+- **Hot reload: live state migration across record/class shape changes
+  (spec 049 §6, Phase 2).** Editing a record or class that a hook stores
+  (`UseState` / `UseReducer` / `UseRef` / `UseMemo` / `UsePersisted`) now
+  migrates the live value onto the new shape instead of resetting it. At the
+  start of each hot-reload pass — before any `Render()` runs — every live
+  `RenderContext` value-swaps the hook cells whose stored type the runtime
+  reported as updated, copying fields by name onto a freshly-constructed
+  instance (`ReactorHotReloadCopier`); fields that can't be mapped are dropped
+  with a log line rather than throwing. Cycle-guarded and block-listed against
+  native handles (`IntPtr` / `Compositor` / `Visual` / `UIElement`). Devtools
+  hook snapshots carry a `Migrated` flag so the inspector can show which cells
+  were value-swapped.
+
+- **Hot reload: subtree migration on component-identity change (spec 049 §7,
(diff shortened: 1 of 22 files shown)


codemonkeychris:
Addressed the review feedback in 654c09d1:

- **AOT Selftests failure** `HotReload_ComponentMigratesState` is now in the runner's AOT skip list. Migration is reflection-based and JIT-only by design (production entry is gated on `HotReloadService.IsHotReloadLive` / `MetadataUpdater.IsSupported`, always false under NativeAOT). The fixture bypasses that gate to exercise the copier directly, so its migration-success assertions only hold under JIT. The sibling `HotReload_ChildHookOrderRecovery` (no reflection) still runs under AOT.
- **RenderContext.MigrateHooksForHotReload** the `WithinUpdatePass` guard now runs *before* the `Migrated` reset loop, so a stray call outside a pass is a true no-op and never wipes the prior pass's devtools annotations.
- **ReactorHotReloadCopier.TryMigrate** destination fields are now enumerated across the full type hierarchy (`GetInstanceFieldsRecursive`) so inherited private base-class fields migrate too. Also dropped the redundant `nint`/`nuint` block-list entries (aliases of `IntPtr`/`UIntPtr`).
- **ReactorHostControl** restored the orphaned `RecoverFromHookOrder` doc comment.
- The null-deref note at Reconciler.cs:1710 is a false positive: the `renderCtx` use site is guarded by `renderCtx is not null` in the catch `when` clause, and the `node.Component.Context` deref is inside `if (node.Component is not null)`.

Verified locally: build 0 errors, 9147 unit tests pass, full selftest suite 0 failures (JIT). AOT path will be confirmed by CI.

codemonkeychris:
Addressed CR finding #3 (generic catch in migration paths) in 4b41260b. Rather than fully removing the catches (the fallback-to-recreate behavior is intentional and desirable), I narrowed all three migration catch sites — ReactorHotReloadCopier.CreateInstance and both catches in Reconciler.TryHotReloadMigrateComponent — to filter out OutOfMemoryException/StackOverflowException. Fatal exceptions now propagate; ordinary reflection faults still drive a clean fallback (unmount/mount recreate or keep-old) instead of proceeding with a half-migrated component.

PR #506, opened 2026-06-01
User/sumand/blank test app for perf

subhajitm:


codemonkeychris:
I know this is still draft PR, but please put this next to the other *blank perf tests, so we have them all grouped in the same location

PR #512, opened 2026-06-01
fix(#511): restore TitleBar caption-button inset for RightHeader

### LINKED ISSUE #511: [Bug] TitleBar Right Header is misplaced
### What happened?

The Right Header content of the TitleBar has started overlapping with the window controls.

### Steps to reproduce

1. Run the ReactorGallery app
2. Observe the location of the dark-mode button overlapping window buttons:

Image

This regression was introduced in PR #455

### Reactor version / commit

56a79400f8927a44a2946b530d362a74762d765c

### Platform

x64

### .NET SDK version

10.0.300

### Windows App SDK version

2.0.1

### Confirmation

- [x] I have searched [existing issues](https://github.com/microsoft/microsoft-ui-reactor/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] This bug reproduces against the current `main` branch (or I've noted otherwise above).

codemonkeychris:
Fixes #511.

## Root cause (regression from PR #455)

WinUI 3 `TitleBar.OnApplyTemplate()` calls `UpdatePadding()`, which reads `AppWindow.TitleBar.RightInset` to size `RightPaddingColumn` — the column that keeps `RightHeader` content clear of the system min/max/close buttons. `RightInset` is non-zero **only** when `Window.ExtendsContentIntoTitleBar` is `true` at template-apply time, and `OnApplyTemplate` runs **before** `Loaded`.

The legacy `MountTitleBar` set `ExtendsContentIntoTitleBar = true` and called `SetTitleBar(...)` synchronously, pre-tree-attach. PR #455 ported `MountTitleBar` to a descriptor and moved both calls into the descriptor's own `Loaded` handler. By then `UpdatePadding()` had already pinned `RightPaddingColumn` to width `0`, so `RightHeader` content (e.g. the gallery's dark-mode button) rendered on top of the caption buttons.

Issue reporter's measurement matches exactly: last column was 138px before #455, 0px after.

## Fix

`src/Reactor/Core/V1Protocol/Descriptor/Descriptors/TitleBarDescriptor.cs` — move the `ExtendsContentIntoTitleBar = true` / `SetTitleBar(...)` write from the `Loaded` handler back into the `Imperative` mount lambda (synchronous, pre-tree-attach), restoring the legacy ordering.

## Regression test

`CovBoost2_TitleBarRightHeaderInset` in `tests/Reactor.AppTests.Host/SelfTest/Fixtures/CoverageBoostFixtures2.cs`:

- Hosts a `TitleBar` with a `RightHeader` in an **isolated `Window` + `ReactorHost`** (with `ActiveHostInternal` save/restore in `try`/`finally`) so the probe does not mutate the shared selftest harness window.
- Renders one frame, walks to the WinUI `TitleBar`'s `RightPaddingColumn` template part, and asserts `Width.Value > 0`.

Verified the test correctly **fails** (`width=0`) when the source fix is reverted and **passes** (`width=137`) with the fix in place.

## Validation

- `dotnet build src/Reactor/Reactor.csproj -p:Platform=x64` — 0 warnings / 0 errors
- `dotnet test tests/Reactor.Tests -p:Platform=x64 --no-build` — **9134 passed / 0 failed / 62 skipped**
- TitleBar selftest filter (4 fixtures, 22 checks) — all green; existing `FloatingTitleBar_*` / `FloatingMulti_*` fixtures unaffected

(Visual proof to be added in a comment.)


codemonkeychris:
Screenshot 2026-06-01 101619

visual of fix

codemonkeychris:
Addressed in 11b4969d.

The fixture now drives the isolated host directly:

- `Task.Delay(50) + Harness.Render()` removed — replaced with `await host.WaitForIdleAsync()` + `window.Content.UpdateLayout()` on the isolated window (you were right that `Harness.Render` was only draining `ReactorApp.PrimaryWindow`'s harness host, not the probe host).
- One `Low`-priority `DispatcherQueue.TryEnqueue` yield after the first `UpdateLayout` to drain WinUI's deferred template/realization messages (mirrors the wave pattern in `Harness.Render` itself).
- The `FindFirst<TitleBar>` probe is wrapped in a bounded re-query loop so a slow first wave doesn't false-negative.
- Re-verified that the fixture still fails (`width=0`) when the source fix is reverted and passes (`width=137`) with it in place.

I also captured the broader guidance in `TESTING.md` under a new **Selftest waiting patterns** subsection (`Render` only when you literally need to pump the event queue; `WaitFor` for anything that may be delayed; `host.WaitForIdleAsync` for isolated hosts; plus the `ActiveHostInternal` save/restore convention).

PR #515, opened 2026-06-02
Spec 051: Devtools trimmability and isolation

codemonkeychris:
## Summary

Implements **spec-051: Devtools Trimmability and Isolation**, plus two minimal AOT samples that measure the trim win.

### Phase 1 — Feature-switch gating
- Gate devtools activation behind the `Reactor.DevtoolsSupport` feature switch (`ReactorFeatures.IsDevtoolsSupported`, `[FeatureSwitchDefinition]`), so trimming/AOT drops the entire devtools tail when the switch is off.
- Remove the `bool devtools`/`bool preview` parameters from `ReactorApp.Run`; activation is now switch (capability) + `--devtools` arg (runtime opt-in).
- A switch-off binary that receives `--devtools run` prints an actionable error and exits non-zero.

### Phase 2 — Package isolation
- Split devtools into its own optional `Microsoft.UI.Reactor.Devtools` package so the IL is not even resolvable unless referenced.
- Apps reference it via `samples/Directory.Build.props` (Debug) / explicit project reference.

### Measurement samples + size minimization
Two minimal AOT apps measure the switch-off vs switch-on footprint. Both publish framework-dependent (not self-contained), strip managed + NativeAOT PDBs and XML docs, use `OptimizationPreference=Size`, and explicitly remove the Windows ML natives (onnxruntime/DirectML) that WindowsAppSDK force-pulls.

| | switch-off (`hello-world-aot`) | switch-on (`hello-world-aot-devtools-on`) | Δ devtools |
|---|---|---|---|
| app exe | 6.34 MB | 10.90 MB | **+4.56 MB** |
| total publish | 7.71 MB (7 files) | 12.27 MB (7 files) | +4.56 MB |

Everything except the native exe is byte-identical, so devtools costs ~4.56 MB of trimmed AOT image — exactly what the feature switch saves when off. (Baseline self-contained publish was 289 files / 156.7 MB before these strips.)

## Verification
- E2E gates pass: `Spec051_DevtoolsCliFallback_EmitsActionableError`, `Spec051_DevtoolsEndToEnd_SwitchOn_McpServerStarts`.
- Unit tests (`Reactor.Tests`) + selftests run on this branch.
- Implementation tracker committed.

🤖 Generated with GitHub Copilot CLI


codemonkeychris:
## Addressed Copilot review feedback

All three findings from the automated review are resolved in 986be2e2:

1. **`tools/list` empty `inputSchema`** (`McpDispatcher.cs`) — `BuildToolsListResult` now serializes each tool's registered `InputSchema` into the response instead of emitting `{}`, so MCP clients can discover argument shapes. Uses the same `[UnconditionalSuppressMessage]` pattern already established for devtools JSON serialization in `DevtoolsMcpServer` (the devtools package is the trimmable/opt-in path, not AOT-constrained). Added regression test `ToolsList_EmitsRegisteredInputSchema`.

2. **Packed targets filename** (`Reactor.csproj`) — the build props are now packed as `build/Microsoft.UI.Reactor.targets` (matching `PackageId`) so NuGet auto-imports them into consuming projects. Without this, package consumers never got the default `Reactor.DevtoolsSupport=false` switch and the trimmer couldn't fold away devtools. The source file keeps its short name for the repo's own `Directory.Build.props` import.

3. **Debug-only test devtools reference** (`tests/Directory.Build.props`) — dropped the `'$(Configuration)' == 'Debug'` clause. CI builds `Reactor.slnx` in **Release**, and the test hosts reference devtools types (e.g. `PreviewCaptureServer`) unconditionally, so the Debug-only reference broke the Release solution build. Verified `Reactor.Tests` now builds clean in Release.

### Verification
- `Reactor.Tests` Release build: **0 errors**.
- McpDispatcher unit tests: **10 passed** (incl. new schema regression test).
- Full unit suite: **9134 passed**; selftests: **1041 passed**.

### Final AOT size numbers
| | switch-off (`hello-world-aot`) | switch-on (`hello-world-aot-devtools-on`) | Δ devtools |
|---|---|---|---|
| app exe | 6.34 MB | 10.90 MB | **+4.56 MB** |
| total publish | 7.71 MB (7 files) | 12.27 MB (7 files) | +4.56 MB |

Both publishes are framework-dependent, PDB/doc-stripped, `OptimizationPreference=Size`, with Windows ML natives removed. Everything except the native exe is byte-identical — devtools costs ~4.56 MB of trimmed AOT image. (Baseline self-contained was 289 files / 156.7 MB.)

PR #540, opened 2026-06-06
Fix gallery source snippets + Copy button + Optional<T> factory audit (closes #538)

codemonkeychris:
Closes #538.

### 1. Gallery — fix the broken source snippets (the user-visible bug from #538)

The `Source code` expander on every gallery page hardcoded `\n` inside `@"…"` verbatim strings. Verbatim `\n` is the literal two-character escape, so snippets rendered as one long line with visible `\n` characters, overflowing the card.

- Rewrote every offending `@"...\n..."` source-code literal across **20 gallery pages** to a C# 11 raw string literal (`"""..."""`). The displayed text now matches exactly what the user would type — real line breaks, no `""`-doubled quotes, no escape soup.
- No post-processing of sample text — if a future sample genuinely needs `\n` displayed, it's honored.

### 2. Gallery — add a Copy button, selectable text, horizontal scroll

- New `CopyToClipboardButton : Component<string>` overlaid top-right of the source pane. Uses `Clipboard.SetContent` + `DataPackage`; flips its label to "Copied" for 1.5s. Wrapped in try/catch so a transient `RPC_E_CALL_REJECTED` from the clipboard layer can't crash the gallery. Per-click `UseRef<int>` generation token ensures rapid clicks reliably keep `Copied` showing 1.5s after the *last* click.
- `IsTextSelectionEnabled = true` and `TextWrapping.NoWrap` on the source `TextBlock` so manual Ctrl-C still works and long lines scroll horizontally instead of wrapping into noise.
- Inner `ScrollView` constructed with `ContentOrientation = ScrollingContentOrientation.Both` so the NoWrap text actually pans rather than getting clipped.

### 3. Framework — type controlled-prop factory parameters as `Optional<T>` with `Unset` default

While testing change #2, we discovered that **clicking Copy collapsed the expander**. Root cause was a framework bug exposed by the Copy button's setState: factory overloads like

`csharp
public static ExpanderElement Expander(string h, Element c,
    bool isExpanded = false, Action<bool>? onIsExpandedChanged = null) =>
    new(h, c, Optional<bool>.Of(isExpanded), onIsExpandedChanged);
`

silently wrapped the default-bool as `Optional<T>.Of(default)` → CONTROLLED-TO-FALSE. Every parent re-render (e.g. the Copy button's "Copied" → "Copy" flip) ran `ExpanderHandler.Update`, detected `IsExpanded` drift vs the live WinUI control, and wrote `false` back, collapsing the user-opened pane.

The fix: every factory whose underlying record property is `Optional<T>` now has a **single** overload typed `Optional<T>` with `= default` (==Unset):

- Omitting the parameter → uncontrolled (the WinUI control owns its state).
- Passing a value (`Slider(0.5)`, `Expander("h", c, true)`) → flows through the existing implicit `T → Optional<T>` conversion → controlled.
- Passing `Optional<T>.Unset` → explicit uncontrolled.

**Audited factories:** `TextBox`, `PasswordBox`, `NumberBox`, `AutoSuggestBox`, `RichEditBox`, `Slider`, `RatingControl`, `CheckBox`, `ThreeStateCheckBox`, `RadioButton`, `ToggleSwitch`, `ToggleSplitButton`, `RadioButtons`, `ComboBox(string[])`, `ListBox`, `SelectorBar`, `PipsPager`, `CalendarDatePicker`, `DatePicker`, `TimePicker`, `ColorPicker`, `Expander`.

**Notes:**
- `TabView` / `Pivot` / `ListView` / `GridView` / `FlipView` keep their `(int? selectedIndex, …)` overload alongside the new `(Optional<int> = default, …)` one as a convenience null→Unset shortcut.
- `ComboBox(Element[], …)` keeps its non-defaulted `selectedIndex` / `handler` parameters to disambiguate from `ComboBox(string[], …)` — there's an implicit `string → Element` conversion (`Element.cs:290`), so a collection-expression call like `ComboBox(["A","B"], selectedIndex: 0)` would otherwise match either overload.

**Tests:**
- New `tests/Reactor.Tests/DslFactoryUncontrolledDefaultsTests.cs` (23 tests) locks in "no-arg factory call → `Optional<T>.Unset`" for every Optional-typed controlled prop.
- `RichEditBoxElementTests.RichEditBox_Creates_With_Defaults` updated to assert `HasValue == false` (was previously asserting `Text == ""`, which was the silent-controlled-default bug).
- Existing `DslFactoryOptionalTests` cover the "passes a value" and "passes Optional<T>.Unset" axes — still pass.

## Verification

- `grep "@\".*\\\\n.*\"" samples\ReactorGallery` → zero matches.
- `dotnet build Reactor.slnx -p:Platform=x64` → 0 warnings, 0 errors.
- `dotnet test tests\Reactor.Tests` → **9224 passed, 0 failed, 62 skipped**.
- Gallery smoke-tested: source snippets render with real line breaks, Copy button works, **expander stays open after clicking Copy**.

### DIFF
--- a/samples/ReactorGallery/ControlPages/Collections/FlipViewPage.cs
+++ b/samples/ReactorGallery/ControlPages/Collections/FlipViewPage.cs
@@ -26,7 +26,12 @@ class FlipViewPage : Component
                                 .Background(c).Size(300, 200)
                         ).ToArray()
                     ),
-                    @"FlipView(\n    Border(TextBlock(""Item 1"")).Background(""#FF4444"").Size(300,200),\n    Border(TextBlock(""Item 2"")).Background(""#44AA44"").Size(300,200)\n)"),
+                    """
+                    FlipView(
+                        Border(TextBlock("Item 1")).Background("#FF4444").Size(300, 200),
+                        Border(TextBlock("Item 2")).Background("#44AA44").Size(300, 200)
+                    )
+                    """),
 
                 SampleCard("Data-Driven FlipView",
                     FlipView(
@@ -39,7 +44,13 @@ class FlipViewPage : Component
                             ).Center()
                         ).Background(c).Size(300, 200)
                     ),
-                    @"FlipView(\n    colors,\n    c => c,\n    (c, i) => Border(VStack(...)).Background(c).Size(300, 200)\n)")
+                    """
+                    FlipView(
+                        colors,
+                        c => c,
+                        (c, i) => Border(VStack(...)).Background(c).Size(300, 200)
+                    )
+                    """)
             ).Margin(36, 24, 36, 36)
         );
     }
--- a/samples/ReactorGallery/ControlPages/Collections/GridViewPage.cs
+++ b/samples/ReactorGallery/ControlPages/Collections/GridViewPage.cs
@@ -35,7 +35,13 @@ class GridViewPage : Component
                                 .Size(100, 100)
                                 .CornerRadius(ThemeResource.CornerRadius("ControlCornerRadius").TopLeft)                        ).ToArray()
                     ).SelectionMode(selectionMode),
-                    @"GridView(\n    items.Select(i =>\n        Border(TextBlock($""Item {i}"")).Size(100,100)\n    ).ToArray()\n).SelectionMode(ListViewSelectionMode.Multiple)",
+                    """
+                    GridView(
+                        items.Select(i =>
+                            Border(TextBlock($"Item {i}")).Size(100, 100)
+                        ).ToArray()
+                    ).SelectionMode(ListViewSelectionMode.Multiple)
+                    """,
                     OptionPanel(
                         TextBlock("Selection Mode"),
                         ComboBox(modes, mode, setMode)
@@ -50,7 +56,12 @@ class GridViewPage : Component
                             .Foreground("#FFFFFF")
                             .Size(80, 80).CornerRadius(ThemeResource.CornerRadius("OverlayCornerRadius").TopLeft)
                     ),
-                    @"GridView(\n    items, s => s,\n    (s, i) => Border(TextBlock(s)).Size(80,80)\n)")
+                    """
+                    GridView(
+                        items, s => s,
+                        (s, i) => Border(TextBlock(s)).Size(80, 80)
+                    )
+                    """)
             ).Margin(36, 24, 36, 36)
         );
     }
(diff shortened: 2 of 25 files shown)


codemonkeychris:
Addressed both review comments in 4046e560:

1. **CopyToClipboardButton timer race** — added a per-click generation token via `UseRef<int>`. Each click bumps the generation and only the matching continuation is allowed to reset the label. Rapid clicks now reliably keep `Copied` showing for 1.5s after the last click.

2. **Source `ScrollView` clipping with NoWrap** — set `ContentOrientation = ScrollingContentOrientation.Both` on the inner `ScrollView` so long no-wrap lines pan horizontally instead of getting clipped.

codemonkeychris:
**Update — combined the framework fix into this PR per discussion.**

The `Optional<bool>.Unset` workaround in `GalleryControls` was the wrong fix. The right fix was in the framework: the `Expander(string, Element, bool isExpanded = false, …)` overload was silently wrapping the default as `Optional<bool>.Of(false)` → controlled-to-false, so every parent re-render (e.g. the Copy button's `Copied` → `Copy` flip) was writing `IsExpanded=false` back to the live WinUI control and collapsing the user-opened pane.

New commit (b0926e96) audits every factory whose underlying record property is `Optional<T>` and collapses them to a single overload typed `Optional<T>` with `= default` (==Unset). 22 factories touched. The implicit `T → Optional<T>` conversion means existing call sites like `Slider(0.5)` keep compiling unchanged.

23 new `DslFactoryUncontrolledDefaultsTests` lock in the new behavior. Full solution build clean, 9224/9224 unit tests pass, gallery smoke-tested — expander state now persists across the Copy button re-renders.

codemonkeychris:
Addressed the `github-code-quality` generic-catch findings in 2253c68e:

1. **Outer catch on `Clipboard.SetContent`** — now typed as `System.Runtime.InteropServices.COMException`, which covers the documented transient (`RPC_E_CALL_REJECTED` / HRESULT `0x80010001` when another app holds the clipboard marshaller). Anything else propagates.

2. **Inner catch in the `Task.Delay` continuation** — removed entirely in favor of a proper React-style lifecycle guard: a `UseRef<bool>` `isMounted` flag flipped to `false` in a `UseEffect` cleanup. The background continuation now short-circuits on `!isMounted.Current` instead of calling `setCopied` on a torn-down RenderContext. No exception suppression at all on the late-callback path.

PR #551, opened 2026-06-09
Fix #548: rename AsyncValue.Match 'data' arm to 'loaded'; omitted reloading -> loading()

### LINKED ISSUE #548: Improve UseResource.Match arm naming and reloading default
## Problem

`UseResource(...).Match(...)` currently reads awkwardly in named-argument usage:

```csharp
resource.Match<Element>(
    loading: () => Heading("Loading ..."),
    data: value => RenderValue(value),
    error: ex => TextBlock($"Error\n{ex.Message}"),
    reloading: previous => Heading("Loading ..."));
```

Two API issues came up from real usage:

1. `data` is a weird name for the successful state arm. `loaded` reads better and matches how authors think about the state.
2. `reloading` is often duplicated from `loading`. It is optional today, but `AsyncValue<T>.Match` currently falls back to the `data` arm when `reloading` is omitted. For many call sites, the more natural default is to show the loading UI when a refresh is in progress.

Current implementation: `src/Reactor/Core/AsyncValue.cs` has `Match<TResult>(Func<TResult> loading, Func<T, TResult> data, Func<Exception, TResult> error, Func<T, TResult>? reloading = null)` and dispatches `Reloading` via `(reloading ?? data)(r.Previous)`.

## Expected fix

- Rename the `Match` success arm from `data` to `loaded`.
- Revisit the omitted-`reloading` behavior so authors do not need to duplicate the `loading` arm when they want refreshes to show loading UI.
- Update docs, samples, recipes, and tests that use named arguments.
- Consider source compatibility/migration for public named-argument call sites, since changing parameter names directly is source-breaking.

vriveras:
## Issue
- [#548](https://github.com/microsoft/microsoft-ui-reactor/issues/548) — Improve `UseResource.Match` arm naming and reloading default

## Fix
`AsyncValue<T>.Match<TResult>` gets two changes requested in the issue:

1. **Rename** the success-arm parameter `data` → `loaded` (reads better; matches how authors think about the *Loaded* state).
2. **Flip the omitted-`reloading` default.** When the value is `Reloading` and no `reloading:` handler is supplied, `Match` now falls through to the `loading()` arm (shows the loading UI) instead of reusing the success arm via `(reloading ?? data)(r.Previous)`. Authors who want stale-while-revalidate now opt in explicitly by passing `reloading:`.

```csharp
// after
Func<T, TResult> loaded,                                              // was: data
...
Data d => loaded(d.Value),
Reloading r => reloading is not null ? reloading(r.Previous) : loading(),  // was: (reloading ?? data)(r.Previous)
```

All in-repo call sites, the reloading-default unit test, the spec (020 §5.1, with an ADR supersession note), the user-guide template + generated guide, the historical task log, and `CHANGELOG.md` were updated.

> ⚠️ **Source-breaking:** external callers using the named argument `data:` must rename to `loaded:`. A back-compat overload is impossible (C# can't overload on parameter name alone). Flagged in the commit message, CHANGELOG, and spec ADR note.

## Validation
| Gate | Result |
|------|--------|
| `dotnet build src/Reactor/Reactor.csproj` (incl. AOT/trim-as-error) | ✅ 0 warnings, 0 errors |
| `dotnet test tests/Reactor.Tests` (full suite) | ✅ **9266 passed, 0 failed, 64 skipped** |
| `AsyncValue` focused suite | ✅ 19/19 |

Both behavioral tests pin the new contract:
- `Match_Reloading_Falls_Through_To_Loading_When_Null` — omitted `reloading:` → `loading()`.
- `Match_Reloading_Uses_Override_When_Provided` — explicit `reloading:` → handler used.

## Code review
Reviewed by the crew reviewer. Verdict was **REQUEST-CHANGES** on the first pass — code was correct/complete but docs (spec 020 §5.1, the `.md.dt` template prose, and a CHANGELOG entry) hadn't been updated. **All flagged items are now addressed in this PR.**

## Proof
**Change kind:** library / public-API change.

**What it does (user-visible):** the success arm is now `loaded:`; omitting `reloading:` shows the loading UI on refresh instead of silently re-rendering stale data; stale-while-revalidate is now an explicit opt-in.

**Evidence:** the two behavioral tests above exercise both the changed default and the explicit-opt-in path; the source diff shows the one-line dispatch flip. (No standalone driver `.exe` — the repo's `Reactor.SignaturesGen` codegen crashes on ARM64 dev boxes, pre-existing and unrelated; the xUnit tests run the same code paths.)

## How to verify locally
```bash
dotnet build src/Reactor/Reactor.csproj
dotnet test tests/Reactor.Tests --filter "FullyQualifiedName~AsyncValue"
# (on an ARM64 dev box add: -p:SkipSignaturesGen=true -r win-arm64)
```

## Conventions honored
- Immutable records; no new reflection (AOT/trim-as-error stays clean)
- Docs are generated — edited the `.md.dt` template (+ synced the generated guide)
- Spec-as-ADR: recorded the behavior flip as a supersession note in spec 020 §5.1
- CHANGELOG breaking-change entry added

Closes #548.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -195,6 +195,16 @@ Conventions for contributors:
 
 ### Changed (breaking)
 
+- **`AsyncValue<T>.Match` success arm renamed `data` → `loaded`; omitted
+  `reloading` now falls through to `loading()`.** The success delegate parameter
+  was renamed (source-breaking for named-argument callers; no back-compat
+  overload is possible because overloads can't differ by parameter name alone).
+  When the value is `Reloading` and no `reloading:` handler is supplied, `Match`
+  now renders the `loading()` arm instead of reusing the success arm
+  (`(reloading ?? data)(r.Previous)`). Pass `reloading:` explicitly to keep the
+  last-known value visible during a refresh (stale-while-revalidate).
+  (issue #548, spec 020 §5.1)
+
 - **`ReactorApp.Run` devtools parameters removed (spec 051 §13).** The
   `devtools:` and `preview:` overload parameters are gone. Enable devtools
   capability in the app project with `<RuntimeHostConfigurationOption
--- a/docs/_pipeline/apps/async-resources-cookbook/App.cs
+++ b/docs/_pipeline/apps/async-resources-cookbook/App.cs
@@ -104,7 +104,7 @@ class AfterUseResourceExample : Component
 
         return user.Match<Element>(
             loading: () => TextBlock("Loading…").Padding(24),
-            data: u => VStack(4,
+            loaded: u => VStack(4,
                 Heading("After: one hook").FontSize(14),
                 TextBlock(u.Name).FontSize(20).Bold(),
                 TextBlock(u.Role).Opacity(0.6)
@@ -200,7 +200,7 @@ class PendingFallbackExample : Component
                 deps: new object[] { "user-1" });
             return user.Match<Element>(
                 loading: () => TextBlock("• user loading…").Opacity(0.5),
-                data: u => TextBlock($"• {u.Name} ({u.Role})"),
+                loaded: u => TextBlock($"• {u.Name} ({u.Role})"),
                 error: ex => TextBlock($"• error: {ex.Message}"));
         }
     }
@@ -229,7 +229,7 @@ class PendingFallbackExample : Component
                 deps: new object[] { "stats" });
             return user.Match(
                 loading: () => TextBlock("• stats loading…").Opacity(0.5),
-                data: _ => TextBlock("• stats ready"),
+                loaded: _ => TextBlock("• stats ready"),
                 error: ex => TextBlock($"• error: {ex.Message}"));
         }
     }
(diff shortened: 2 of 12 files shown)


vriveras:
## ✅ Test & coverage gate — green

Verified on this PR's HEAD commit (`7ff16eed`).

### Full unit suite
`dotnet test tests/Reactor.Tests`

| Total | Passed | Failed | Skipped | Exit |
|------:|-------:|-------:|--------:|:----:|
| 9330  | **9266** | **0** | 64 | **0** |

No failures — the `Reloading`-default flip caused zero regressions across the suite.

### `AsyncValue<T>.Match` coverage (coverlet / cobertura)
Method `Match<TResult>`: **line 87.5% · branch 90%**. Every logical arm is exercised, and the decisive `reloading is not null ? reloading(r.Previous) : loading()` ternary — the exact line #548 changed — is covered on **both** sides:

| `Match` arm | Covering test |
|---|---|
| `Loading => loading()` | `Match_Dispatches_Loading` |
| `Data d => loaded(d.Value)` | `Match_Dispatches_Data` |
| `Error e => error(e.Exception)` | `Match_Dispatches_Error_With_Exception` |
| `Reloading r` **with** `reloading:` → `reloading(r.Previous)` | `Match_Reloading_Uses_Override_When_Provided` |
| `Reloading r` **without** handler → `loading()` | `Match_Reloading_Falls_Through_To_Loading_When_Null` |

The residual <100% is the optional-parameter default plumbing, not an untested dispatch branch. **No coverage gap.**

<sub>Build note: run locally on an ARM64 box with `-p:SkipSignaturesGen=true -r win-arm64` (the `Reactor.SignaturesGen` codegen crash is pre-existing/environmental and unrelated to this change; CI skips it too).</sub>

Issue #416, opened 2026-05-26
API discovery: `reactor.api.txt` only covers 5 categories, agents can't find the rest of the surface


codemonkeychris:
## Summary

`skills/reactor.api.txt` is the canonical index agents use to discover the Reactor API surface, but it only catalogs five sections:

- **Factories** (`Microsoft.UI.Reactor.Factories.*`)
- **Modifiers** (extension methods on `Element`)
- **Hooks** (`RenderContext` / `Component`)
- **Theme tokens**
- **Enums**

This leaves big swaths of public API undiscoverable from the index file, including (non-exhaustive):

- Properties / init-only fields on public records — e.g. `WindowSpec.Opacity`, `WindowSpec.NoActivate`, `WindowSpec.IgnorePointerInput`, plus all the existing fields like `WindowSpec.Title`, `MinWidth`, `Backdrop`, etc.
- Instance methods on public classes — e.g. `ReactorWindow.SetOpacity`, `SetPosition`, `SetSize`, `SetNoActivate`, `SetIgnorePointerInput`, `Update`, `Mount`, `Activate`, `Hide`, `Show`, `Close`, `CenterOnScreen`, …
- Public types themselves (records, sealed classes, structs) that aren't reachable via a Factories/Hooks/Modifiers method — e.g. `DockManager`, `DockTabGroup`, `DockSplit`, `WindowSpec`, `BackdropChoice`, `WindowKey`, `Command`, …
- Constructors and overloads of public types
- Events on public classes — e.g. `ReactorWindow.Closing`, `Activated`, `Closed`, `DpiChanged`

## Motivating example

While working on PR #415 (`feat/window-layered-attrs`), I added three new init-only properties to `WindowSpec` and three matching instance methods on `ReactorWindow`. After committing, ran `mur --regen-api` — both copies of `reactor.api.txt` came back **unchanged**, because none of the new surface fits the five sections the index catalogs.

That's not a bug in the regen tool — it's a gap in what the index covers. The end result is that an agent picking up this codebase cold has no way to discover (a) the new APIs we just added, or (b) most of the existing window/docking/command surface, from the index alone.

## Why it matters for agents

[...]

PR #554, opened 2026-06-09
Add `Public types` section to reactor.api.txt generator (#416)

vriveras:
# API discovery: `reactor.api.txt` only covers 5 categories, agents can't find the rest of the surface

## Issue
- [#416](https://github.com/microsoft/microsoft-ui-reactor/issues/416) — API discovery: `reactor.api.txt` only covers 5 categories, agents can't find the rest of the surface

## Lane
improvement — the author explicitly disclaimed a defect ("not a bug in the regen tool — it's a gap in what the index covers"); ask is to broaden coverage of an existing tooling artifact.

## Change
Closed the API-discovery gap by adding a 6th **"Public types"** section to the generated agent index (`skills/reactor.api.txt` + its mirror under `plugins/reactor/skills/reactor-dsl/references/`). The fix extracts the existing generator into a new ARM64-safe sibling class library (`tools/Reactor.ApiIndex`) — the same logic that previously lived only inside the apphost that crashes on ARM64 with `0xC0000005` — so an in-process xUnit test can drive it. The 5 original sections are emitted byte-for-byte unchanged (locked by `ExistingSections_Unchanged`), and a new `Index_IsUpToDate` test both regenerates the committed copies (`UPDATE_API_INDEX=1`) and fails at PR time if anyone adds public surface without regenerating — exactly the `WindowSpec.Opacity` scenario the issue called out.

## Validation
- `tests` — **PASS**. Full Reactor.Tests suite: `Failed: 0, Passed: 9286, Skipped: 64, Total: 9350, Duration: 11s` on ARM64 (`dotnet test tests/Reactor.Tests -p:SkipSignaturesGen=true -p:SkipReactorApiGen=true -r win-arm64`). No flakes.

## Iteration log
| # | lane | plan | diff | validate |
|---|------|------|------|----------|
| 1 | improvement | plan-1.md | diff-1.md | PASS |

### Change kind
improvement (issue #416 — broaden `reactor.api.txt` coverage)

### What this does (user-visible)
An agent reading the canonical Reactor API index can now discover the public **properties / init-only fields, instance methods, events, and constructors** of every public type in `Reactor.dll`. Previously the index only catalogued Factories, Modifiers, Hooks, Theme tokens, and Enums, so newly-added surface (e.g. `WindowSpec.Opacity`, `ReactorWindow.SetPosition`) was invisible from the index alone. A new `Index_IsUpToDate` xUnit test also FAILS at PR time if anyone adds public surface without regenerating the index.

### Evidence (full artifacts under `.plans/assigned-issue-autopilot/proofs/`)

**Sections of the canonical index — 5 → 6** (`sections-before-after.txt`)
```
BEFORE: Factories | Modifiers | Hooks | Theme tokens | Enums
AFTER : Factories | Modifiers | Hooks | Theme tokens | Enums | Public types (ctors / properties / methods / events)
```

**Metric-delta vs `origin/main`** (`metric-delta.txt`)
```
Metric                       Before   After
Sections in reactor.api.txt    5        6
Total lines                   1017     7201
WindowSpec hits                  1        8
Opacity hits                     5       17
ReactorWindow hits               2        8
SetPosition hits                 0        2
SetNoActivate hits               0        1
```

**Discovery matrix — every symbol the issue called out** (`discovery-matrix.txt`)
```
[YES] WindowSpec.Opacity              [YES] ReactorWindow.SetIgnorePointerInput
[YES] WindowSpec.NoActivate           [YES] ReactorWindow.Activate
[YES] WindowSpec.IgnorePointerInput   [YES] ReactorWindow.Close
[YES] ReactorWindow.SetOpacity        [YES] ReactorWindow.CenterOnScreen
[YES] ReactorWindow.SetPosition       [YES] ReactorWindow.Closing  (event)
[YES] ReactorWindow.SetSize           [YES] ReactorWindow.Activated (event)
[YES] ReactorWindow.SetNoActivate
```

**What an agent actually sees** (`windowspec-excerpt.txt`, `reactorwindow-excerpt.txt`)
```
### record WindowSpec
new()
AspectRatioBasis AspectRatioBasis { get; init; }
BackdropChoice Backdrop { get; init; }
…
bool IgnorePointerInput { get; init; }
…

### class ReactorWindow
…
Activate() → void
CenterOnScreen() → void
Close() → void
SetIgnorePointerInput(bool ignore) → void
SetNoActivate(bool noActivate) → void
SetOpacity(double opacity) → void
SetPosition(double x, double y) → void
```

**Behavior-preservation + PR-time regen guard** (`apiindex-tests.txt`)
6 focused xUnit tests, driven in-process so the ARM64 apphost crash is bypassed:
- `PublicTypes_Section_IsPresent`
- `PublicTypes_Surfaces_WindowSpec_Opacity`
- `PublicTypes_Surfaces_ReactorWindow_SetPosition`
- `PublicTypes_Surfaces_Constructors_And_Events`
- `ExistingSections_Unchanged` (5 original sections byte-identical)
- `Index_IsUpToDate` (committed index == `Generate(asm)`; `UPDATE_API_INDEX=1` regenerates both copies — the ARM64-safe regen path)

Result: `Failed: 0, Passed: 6, Skipped: 0, Total: 6, Duration: 756 ms`.

**Full suite still green** (`full-suite.txt`)
```
Passed!  - Failed: 0, Passed: 9286, Skipped: 64, Total: 9350, Duration: 11 s - Reactor.Tests.dll (net10.0)
```

### Acceptance criteria → coverage map
| Criterion from #416 | Met by |
|---|---|
| Discover public properties / init-only fields on every public type | `EmitPublicTypes` properties arm; discovery matrix `WindowSpec.*` rows |
| Discover public instance methods on every public type | methods arm; matrix `ReactorWindow.Set*` / `Activate` / `Close` / `CenterOnScreen` rows |
| Discover public events on every public type | events arm; matrix `Closing` / `Activated (event)` rows |
| Discover constructors / overloads | ctors arm; `new()` line in the WindowSpec block |
| Regen catches new surface added in a PR (the `WindowSpec.Opacity` example) | `Index_IsUpToDate` test — fails the build if the index is stale |
| Format remains diffable/reviewable (not a binary symbol dump) | one line per member, alphabetized, plain text |

## How to verify locally
```bash
# Focused (the new tests):
dotnet test tests/Reactor.Tests --filter "FullyQualifiedName~Tooling.ApiIndexGeneratorTests" -p:SkipSignaturesGen=true -p:SkipReactorApiGen=true -r win-arm64

# Regenerate the committed index copies (the ARM64-safe path; bypasses the crashing apphost):
$env:UPDATE_API_INDEX=1
dotnet test tests/Reactor.Tests --filter "FullyQualifiedName~Tooling.ApiIndexGeneratorTests.Index_IsUpToDate" -p:SkipSignaturesGen=true -p:SkipReactorApiGen=true -r win-arm64
Remove-Item Env:UPDATE_API_INDEX

# Full suite:
dotnet test tests/Reactor.Tests -p:SkipSignaturesGen=true -p:SkipReactorApiGen=true -r win-arm64
```

## Conventions honored
- Generator reflection lives in the **tooling** project (`tools/Reactor.ApiIndex`) — `Reactor.dll` shipping surface unchanged, no new reflection in core (AOT-clean preserved).
- xUnit + `[Collection("ConsoleTests")]` for the FS-mutating regen arm.
- Sibling-classlib pattern avoids re-firing `Reactor.SignaturesGen`'s AfterBuild apphost target on ARM64 (preferred safe default per the plan).
- ARM64 build flags `-p:SkipSignaturesGen=true -p:SkipReactorApiGen=true -r win-arm64` used throughout.


vriveras:
## 📎 Proof summary (also at the top of this PR's description)

Posting this in the conversation thread for visibility — the full proof package is in the PR body above and the source artifacts live in .plans/assigned-issue-autopilot/.

### Sections of the canonical index — 5 → 6
```
BEFORE: Factories | Modifiers | Hooks | Theme tokens | Enums
AFTER : Factories | Modifiers | Hooks | Theme tokens | Enums | Public types (ctors / properties / methods / events)
```

### Metric-delta vs `origin/main`
| Metric | Before | After |
|---|---|---|
| Sections in reactor.api.txt | 5 | **6** |
| Total lines | 1017 | 7201 |
| `WindowSpec` hits | 1 | 8 |
| `Opacity` hits | 5 | 17 |
| `ReactorWindow` hits | 2 | 8 |
| `SetPosition` hits | 0 | 2 |
| `SetNoActivate` hits | 0 | 1 |

### Discovery matrix — every acceptance-criterion symbol from #416
```
[YES] WindowSpec.Opacity              [YES] ReactorWindow.SetIgnorePointerInput
[YES] WindowSpec.NoActivate           [YES] ReactorWindow.Activate
[YES] WindowSpec.IgnorePointerInput   [YES] ReactorWindow.Close
[YES] ReactorWindow.SetOpacity        [YES] ReactorWindow.CenterOnScreen
[YES] ReactorWindow.SetPosition       [YES] ReactorWindow.Closing  (event)
[YES] ReactorWindow.SetSize           [YES] ReactorWindow.Activated (event)
[YES] ReactorWindow.SetNoActivate
```

### Tests
- **New focused tests** (Tooling.ApiIndexGeneratorTests): Failed: 0, Passed: 6, Skipped: 0, Total: 6, Duration: 756 ms
- **Full `Reactor.Tests` suite**: Failed: 0, Passed: 9286, Skipped: 64, Total: 9350, Duration: 11s on ARM64

### ARM64-safe regen path
```bash
$env:UPDATE_API_INDEX=1
dotnet test tests/Reactor.Tests --filter "FullyQualifiedName~Tooling.ApiIndexGeneratorTests.Index_IsUpToDate" -p:SkipSignaturesGen=true -p:SkipReactorApiGen=true -r win-arm64
Remove-Item Env:UPDATE_API_INDEX
```

A separate adversarial code review is being posted by the crew reviewer.

vriveras (commented):
## 💬 COMMENT — minor nits / suggestions

Adversarial review of PR #554 (commit `03e6332f`). Scope verified against `plan-1.md`: exactly the 8 files listed, no out-of-scope changes. The 5 original `Emit*` methods moved **byte-identical** from `Program.cs` into `ApiIndexGenerator.cs` — diffed body-by-body against `origin/main:tools/Reactor.SignaturesGen/Program.cs`, only the `System.Runtime.CompilerServices.ExtensionAttribute` references gained a `global::` prefix (semantically identical). No reflection leaked into `src/Reactor`. AOT/trim posture preserved.

Net assessment: solid, low-risk refactor that closes a real discovery gap. No blockers, no majors, a handful of small things worth a look before merge.

## Findings

### Nit · `tools/Reactor.ApiIndex/ApiIndexGenerator.cs:273-281` — `KindOf` mis-labels `record struct`
`KindOf` returns `"record"` only when a `<Clone>$` method is present, which is synthesized for **record classes** but not for **record structs** (the latter get value-equality members but no `Clone`). A `public record struct Foo(...)` therefore prints as `### struct Foo`. Not a blocker (the surface is still listed), but the label is wrong. Suggested fix: also check for the `EqualityContract` property (record class) or use `Type.GetCustomAttribute<System.Runtime.CompilerServices.IsReadOnlyAttribute>()` plus a presence check on the compiler-synthesized `PrintMembers` method, which fires for both record kinds.

### Nit · `ApiIndexGenerator.cs:225` and `:282-301` — `Short(t)` drops nesting / namespace, collision risk in `### …` headers
The Public-types section emits `### record Options` / `### class State` etc. with the unqualified name. `GetExportedTypes()` returns nested types as separate entries, so a nested `Foo+Options` and a top-level `Options` (or two unrelated nested `Options`) will both produce `### record Options` headers — agents can't tell them apart. Suggested fix: in the header only, use `t.FullName` (or `t.Namespace + "." + Short(t)`); keep the compact `Short()` for member parameter/return types.

### Nit · `ApiIndexGenerator.cs:31-39` — header preamble drift is NOT covered by `ExistingSections_Unchanged`
Three new comment lines were added to the file preamble (`# Sections: ...`, `# Public types now ARE covered ...`, `# public type (e.g. WindowSpec.Opacity, ...)`). This is fine and intentional, but `ExistingSections_Unchanged` only compares the span `[## Factories .. ## Public types)` so any future regression to the preamble would silently slip past that "behavior preservation" test. Consider either (a) noting in the test's XML/comment that the header preamble is intentionally *not* locked, or (b) adding a tiny assertion that the first non-blank line is still `# Reactor API — signatures index (generated)`.

### Nit · `tests/Reactor.Tests/Tooling/ApiIndexGeneratorTests.cs:97-113` — `UPDATE_API_INDEX=1` contract is only documented in the PR body
The env-var-driven regen path is a real ergonomic win on ARM64, but it's only documented in (a) a code comment, (b) the PR body, and (c) `proof-1.md`. A future contributor staring at a failing `Index_IsUpToDate` won't see how to fix it. Suggested fix: when the assert fails, include the remediation in the failure message, e.g. `Assert.Equal(committed, generated)` → wrap with `Assert.True(committed == generated, "skills/reactor.api.txt is stale — re-run with UPDATE_API_INDEX=1 to regenerate both committed copies.")`. (Also worth a one-liner under `TESTING.md` § "regen artifacts", but that's a separate PR.)

### Nit · `ApiIndexGenerator.cs:273-281` — `KindOf` doesn't surface `abstract` / `sealed` / `static`-on-instance modifiers
Discovery-grade detail: an agent can't tell from `### class ReactorWindow` whether it is sealed, abstract, or has static factory members. Not blocking (members are still listed), but `### sealed class ReactorWindow` would prevent agents from trying to subclass things they shouldn't. Mention only — happy to defer to a follow-up.

### Nit · `ApiIndexGenerator.cs:236-242` — properties are queried with `BindingFlags.Instance | BindingFlags.Static` but methods only with `Instance`
Asymmetric. Static methods on a non-static type (e.g., a factory-style `public static FooBuilder Build()`) won't appear in the type's block. Probably rare in Reactor's surface today, but if intentional, worth a comment; otherwise add `| BindingFlags.Static` to the method query too (consistent with the property arm).

### Info · `tools/Reactor.ApiIndex/Reactor.ApiIndex.csproj` — sibling-classlib correctly side-steps the AfterBuild apphost
Verified: `Reactor.ApiIndex.csproj` has zero `<Target>` entries; `Reactor.Tests.csproj` adds a `ProjectReference` only to `Reactor.ApiIndex`, not to `Reactor.SignaturesGen.csproj`. The `EmitReactorApiTxt` AfterBuild target in `Reactor.SignaturesGen.csproj` (lines 47–53) therefore does NOT fire from a `Reactor.Tests` build. ARM64-safe design holds. ✅

### Info · Index size 1017 → 7200 lines (~7×)
Per the SignaturesGen csproj comment on lines 24–25, `skills/reactor.api.txt` is "packed into the Reactor NuGet … and embedded into the mur CLI." A 7× growth (still ~120 KB of plaintext) is well within any reasonable embed budget, but call it out — if `mur` has a hard token-budget for the index when shipping it to an LLM context, the embed path may want a "compact mode" follow-up.

## What's good

- **Faithful refactor.** Five `Emit*` methods moved verbatim (verified by line-by-line diff vs `origin/main:tools/Reactor.SignaturesGen/Program.cs`). The `global::` prefix on `System.Runtime.CompilerServices.ExtensionAttribute` is the only delta and is semantically identical.
- **AOT/trim posture preserved.** All new reflection lives in `tools/Reactor.ApiIndex` (class library, `IsPackable=false`) and `tests/Reactor.Tests`. Nothing leaked into `src/Reactor`. No `MakeGenericType` / `MakeGenericMethod` / dynamic dispatch introduced.
- **Sibling-classlib design is the right call** — it solves the ARM64 apphost-crash problem without disturbing the existing `EmitReactorApiTxt` AfterBuild path on x64.
- **`Index_IsUpToDate` is the headline win.** It's exactly the regression guard the original issue (#416, `WindowSpec.Opacity` blind spot) asked for; PR-time stale-index detection is high value for low cost.
- **`[Collection("ConsoleTests")]`** is the right serialization choice for the FS-mutating regen arm — consistent with the repo's documented convention.
- **Edge cases handled correctly:**
  - `init`-only detection via `IsExternalInit` required custom modifier — standard, correct.
  - `IsSpecialName` filter excludes property/event accessors and operators.
  - `'<' / '>'` filter excludes compiler-generated members (`<Clone>$`, `<>c__DisplayClass…`).
  - `IDisposable.Dispose()` — implicit impls correctly surface (`DeclaredOnly` + implicit `public Dispose()` is declared on the type); explicit impls correctly hide (they are non-public).
  - Indexers excluded via `GetIndexParameters().Length == 0`.
  - Static classes (`IsClass && IsAbstract && IsSealed`) excluded from the Public-types section — they're already covered by Factories / Modifiers / Hooks.
- **No out-of-scope file changes** — verified the 8 modified files match `plan-1.md` exactly.

## Risk summary

- **Behavior regression risk for the 5 original sections:** ~zero. `ExistingSections_Unchanged` locks the byte-for-byte output, and the implementation moved verbatim.
- **Build risk on ARM64:** none introduced; the sibling classlib explicitly avoids the apphost path.
- **Downstream consumer risk (NuGet embed, `mur --api`):** low. The artifact is the same format and ~7× larger; worth confirming `mur` doesn't have a hard cap on the embedded blob.
- **Cross-platform test risk (LF vs CRLF):** moot — `TargetFramework=net10.0-windows10.0.22621.0` + `UseWinUI=true` means the test only runs on Windows, where both `StringBuilder.AppendLine` and `File.ReadAllText` see CRLF.
- **Worktree risk for `RepoRoot()`:** acceptable. The walk prefers the **nearest** ancestor containing `Reactor.slnx`, which for `.worktrees/agent-*` resolves to the worktree root and its own `skills/reactor.api.txt`. No accidental cross-worktree assertion.

PR #569, opened 2026-06-11
fix: pin MessagePack to 2.5.301 to address GHSA-hv8m-jj95-wg3x vulnerability

philnach:
The "Vulnerable packages" CI job was failing due to transitive dependencies on **MessagePack** versions 2.5.192 and 2.5.198, which carry a High severity vulnerability ([GHSA-hv8m-jj95-wg3x](https://github.com/advisories/GHSA-hv8m-jj95-wg3x)) — an LZ4 decompression out-of-bounds read that can cause `AccessViolationException` and process termination (DoS). The patched v2 release is 2.5.301.

## Root Cause

MessagePack was pulled in transitively through two chains:

- `GitHub.Copilot.SDK → StreamJsonRpc → MessagePack 2.5.198` — affecting `Reactor.Cli`, `DemoScriptTool`, and their downstream project references (`Reactor.Tests`, `Reactor.DocPipeline.Tests`, `Reactor.MurCheckGuardrail`)
- `Microsoft.VisualStudio.SDK → StreamJsonRpc → MessagePack 2.5.192` — affecting `Reactor.VsExtension`, `Reactor.VsExtension.Tests`, `Reactor.VsExtension.SdkTests`

## Changes

Pinned `MessagePack` explicitly to `2.5.301` in the four root projects that originate these transitive chains:

- **`src/Reactor.Cli/Reactor.Cli.csproj`** — alongside the existing `Nerdbank.MessagePack` pin; fixes all downstream consumers via the ProjectReference chain
- **`samples/apps/demo-script-tool/App/DemoScriptTool.csproj`** — same transitive chain via `GitHub.Copilot.SDK`
- **`src/vs-reactor/Reactor.VsExtension/Reactor.VsExtension.csproj`** — inside the `TF_BUILD != 'true'` condition block; fixes `Reactor.VsExtension.Tests` via its ProjectReference
- **`src/vs-reactor/Tests/Reactor.VsExtension.SdkTests/Reactor.VsExtension.SdkTests.csproj`** — inside the `TF_BUILD != 'true'` condition block (directly references `Microsoft.VisualStudio.SDK`, requiring its own pin)

PR #570, opened 2026-06-12
build: enable SourceLink and ship symbols for all packages

philnach:
## What

Enable [SourceLink](https://learn.microsoft.com/dotnet/standard/library-guidance/sourcelink) for the published packages so consumers can step directly into Reactor source while debugging, and ensure every binary package ships a symbol package (`.snupkg`).

## Why

The framework already published symbol packages, but the PDBs carried no SourceLink metadata, so debuggers couldn't map them back to source on GitHub. Additionally, `Microsoft.UI.Reactor.Advanced` shipped a `.nupkg` but no `.snupkg` at all (its csproj was missing the symbol properties that `Reactor` and `Devtools` already had), so it had no published symbols to debug with.

## Changes

**`Directory.Build.props`** — enable SourceLink for all packable projects:
- `PublishRepositoryUrl` + `EmbedUntrackedSources` (always on) so the package carries the repo URL and every PDB document resolves (including source-generator output).
- `ContinuousIntegrationBuild` + `DeterministicSourcePaths` (gated to `$(CI) == 'true'`) so CI builds normalize source paths to deterministic roots that match the committed source on GitHub, without leaking local paths. Gating to CI keeps local working-tree debugging intact. A comment clarifies that `$(CI)` is the universal "automated build agent" flag — set by **both** the PR-gate and the release/pack pipeline — so deterministic paths do apply to the signed release packages.

No `PackageReference` is needed: `Microsoft.SourceLink.GitHub` is bundled in the .NET SDK and auto-imported for GitHub-hosted repos (which also avoids having to seed the package into the internal build feed).

**`src/Reactor.Advanced/Reactor.Advanced.csproj`** — add `IncludeSymbols` + `SymbolPackageFormat=snupkg`, mirroring `Reactor` and `Devtools`, so Advanced now publishes a symbol package too.

`Microsoft.UI.Reactor.ProjectTemplates` is intentionally unchanged: it's a `PackageType=Template` package with no compiled assembly (`IncludeBuildOutput=false`), so it has no PDB to symbolicate.

## Verification

Built the official pipeline against these changes and inspected the resulting signed packages. All three binary packages — `Microsoft.UI.Reactor`, `Microsoft.UI.Reactor.Advanced`, and `Microsoft.UI.Reactor.Devtools` — produce a `.snupkg`, and `sourcelink test` passes on each PDB, with documents resolving to `raw.githubusercontent.com` at the build commit.

---

> Branch note: the SourceLink work originally lived on `ci/sourcelink`, but that branch was reused by an unrelated automated MessagePack change. To keep things single-purpose, SourceLink now lives here on a fresh `build/sourcelink` branch, and the MessagePack fix is isolated on `ci/sourcelink` / #569.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -52,6 +52,14 @@
         to deterministic roots so PDBs don't leak local paths (e.g. E:\repos\...) and
         match the committed source on GitHub. Gated to CI: turning them on locally
         rewrites paths and breaks stepping into your own working-tree builds.
+
+    NOTE on '$(CI)': despite the name, this is the universal "automated build agent"
+    flag (set by GitHub Actions, ADO, etc.), NOT "the CI/PR-gate pipeline". BOTH the
+    PR-gate (OneBranch.CI.Reactor.yml) AND the release/pack pipeline
+    (OneBranch.Official.Reactor.yml) set CI=true — so the deterministic-path arm below
+    DOES run for the signed release packages. "CI" here means "not a developer's
+    interactive working-tree build", which is exactly when we want deterministic,
+    source-linked PDBs.
   -->
   <PropertyGroup>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -40,6 +40,29 @@
   </PropertyGroup>
 
   <!--
+    SourceLink — lets consumers step into Reactor source while debugging. The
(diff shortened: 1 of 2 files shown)


codemonkeychris (dismissed):
rebase on main to address the vuln dependency issue

philnach:
@codemonkeychris rebased onto latest `main` (now includes the MessagePack `2.5.301` pin from #569). No conflicts — the branch is still just the 3 SourceLink commits on top of the fixed `main`, so the vuln dependency check should now pass. Ready for another look. 🙏

PR #640, opened 2026-06-23
test(selftest): gate docking effect-count asserts on body realization (#445)

### LINKED ISSUE #445: Residual flake: NativeDocking_DockContextHooksResolveOnRealMount fixture still occasionally fails under AOT after #442
## Summary

PR #442 (`fix(selftest): drain dispatcher after UpdateLayout in Harness.Render`) cut the flake rate of the five `NativeDocking_*` fixtures dramatically — main has been green for 8 consecutive CI runs since the merge — but did **not** fully eliminate the underlying race. Hit again on PR #444's CI:

> Job: **AOT Selftests** — run [26617293434](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/26617293434/job/78435707093)
>
> ```
> not ok DockHooks_Host_Resolved - assertion failed
> not ok DockHooks_Pane_TitleResolved - assertion failed
> not ok DockHooks_Pane_KeyResolved - assertion failed
> not ok DockHooks_IsActivePane_TrueWhenActive - assertion failed
> # Total failures: 4
> ```

PR #444's changes are doc-only (templates + doc apps under `docs/`) plus five `<snippet:...>` **comments** in `src/Reactor/Core/V1Protocol/` — byte-equivalent to no-ops for the C# compiler, cannot change runtime behavior. The failure pattern is identical to the original Mode-B cascade I diagnosed in PR #442.

A re-run of just the failed AOT job is in flight; statistically it will pass.

## Root cause (from PR #442 investigation)

The four `DockHooks_*` asserts probe text rendered inside a `Memo` component nested in a `DockManager` pane. Pane bodies render through a WinUI `TabView`, which realizes the selected tab's content presenter via Normal-priority dispatcher work scheduled by the *layout pass*. When the harness's post-`UpdateLayout` dispatcher drain doesn't cover that scheduled work in time, the probe runs against an empty subtree and every initial probe fails as a cascade.

PR #442's fix (`tests/Reactor.AppTests.Host/SelfTest/Harness.cs`):
- `await WaitForIdleAsync()` → `UpdateLayout()` → **Low-priority dispatcher yield** → `UpdateLayout()` → `Task.Delay(16)`

That sequence wins 99%+ of the time and is what flipped main green. The residual rate looks like a path where TabView's content realization is posted at *higher* than Normal priority on AOT (where the dispatcher is fast enough to drain Low without the realization being queued yet), or where the realization completes only after Loaded, which the Low yield doesn't gate.

## Proposed fix — defense-in-depth

Restore the per-fixture `pump-until` guards I removed from the five `NativeDocking_*` fixtures in PR #442. The pattern is the team's established shape for deferred-event flakes (#139, #149, #152, `fix/selftest-gotfocus-pump-until`):

```csharp
for (int i = 0; i < 10 && H.FindText(\"alpha-host:ok\") is null; i++)
    await Harness.Render();
H.Check(\"DockHooks_Host_Resolved\", H.FindText(\"alpha-host:ok\") is not null);
```

Fixtures that need the guard (sentinel probe):

| Fixture (file) | Sentinel probe |
|---|---|
| `TabGroupRendersToTabView` (`NativeDockingSmokeFixture.cs:94`) | `H.FindText(\"native-body-alpha\")` |
| `DockContextHooksResolveOnRealMount` (`NativeDockingSmokeFixture.cs:133`) | `H.FindText(\"alpha-host:ok\")` |
| `Composition_ContentMutationFlowsToActivePane` (`NativeDockingCompositionFixture.cs:25`) | `H.FindText(\"count=0\")` |
| `UseEffectCleanup_BodyRemovedOnPaneClose` (`NativeDockingReliabilityFixture.cs:195`) | `H.FindText(\"effect-body-p1\")` |
| `DynamicallyDockedComponentPage_WithOuterShellState` (`NativeDockingDynamicContentFixture.cs:~624`) | `FindControl<Button>(b => b.Name == \"PixDoc_ToolbarOpenWelcome\")` |
| `CompositionDrivenDocumentsRespectKeyedReconciliation` | `H.FindText(\"<first body sentinel>\")` |

Belt + suspenders: the harness fix carries the 99% case, the per-fixture pumps catch the residual 1% AOT timing window. Combined fail rate should round to zero.

## Validation plan

1. Apply the per-fixture pump-until guards to the six fixtures above (last one was missed in PR #442's local sweep — surfaced only on the second 10x run).
2. Re-run `tools/flake-loop.ps1 -Iterations 30 -Project tests/Reactor.SelfTests` locally; expect 30/30.
3. Land via PR; let CI exercise the AOT path over several green runs to confirm.

## Repro / observation surface

- Job pattern: AOT Selftests
- Affected fixtures: any `NativeDocking_*` whose initial probe targets text/control rendered through `DockManager` → `TabView` → `TabViewItem.ContentPresenter`
- Symptom: every initial assert in the fixture fails as a single cascade (count=4 / 8 / 1 depending on fixture)
- Frequency since #442 merged: 1 hit in 9 PR-attached runs (≤12%)

## Related

- PR #442 — original fix (Harness.Render drain)
- `fix/selftest-gotfocus-pump-until`, #139, #149, #152 — prior applications of the pump-until pattern

🤖 Filed via Claude Code

azchohfi:
## Summary

Closes #445.

Residual NativeAOT selftest flake in the `NativeDocking_*` fixtures. PR #442 (drain dispatcher after `UpdateLayout` in `Harness.Render`) cut the flake rate dramatically, and the later #502/#510 sweep generalized the per-fixture pump-until pattern into `Harness.WaitFor`. Five of the six fixtures flagged in #445 already place their `WaitFor` sentinel guard before their first realized-content assert.

The lone gap was **`NativeDocking_Reliability_UseEffectCleanup_BodyRemovedOnPaneClose`**: it probed `mountedCount == 1` / `cleanupCount == 0` *before* its `effect-body-p1` `WaitFor` guard. The nested component's `UseEffect` `OnMount` (which increments `mountedCount`) only fires once the pane body is committed behind the `TabView` content presenter. On the AOT dispatcher-timing window that realization can land a wave later than a single `Harness.Render` drains, cascading `Reliability_Effect_MountedOnce` against a count of `0`.

## Change

Reorder so the bounded pump-until body-realization guard runs **first**, matching the defense-in-depth pattern the other five NativeDocking fixtures already use. Test-side only — no framework source touched.

## Validation

- All 48 `NativeDocking_*` selftest fixtures pass, including the reordered fixture (`Reliability_Effect_BodyRendered` → `MountedOnce` → `NoCleanupBeforeClose` all `ok`).
- Focused flake loop: **30/30 clean** runs of the full NativeDocking selftest set on ARM64 (standard/non-AOT path — NativeAOT `link.exe` is unavailable on ARM64). Zero `not ok` across all runs.

Why the bounded pump also closes the AOT timing window: `Harness.WaitFor` re-queries the live tree after each `Harness.Render` pass until the sentinel resolves, so the count probe waits for the component's actual realization+effect regardless of whether the deferred `TabViewItem.ContentPresenter` realization was posted at Normal/Low/post-Loaded priority. The loop is bounded (25 passes) so a genuinely broken fixture still fails fast rather than hanging.

### DIFF
--- a/tests/Reactor.AppTests.Host/SelfTest/Fixtures/NativeDockingReliabilityFixture.cs
+++ b/tests/Reactor.AppTests.Host/SelfTest/Fixtures/NativeDockingReliabilityFixture.cs
@@ -219,10 +219,18 @@ internal static class NativeDockingReliabilityFixtures
             host.Mount(_ => managerEl);
             await Harness.Render();
 
-            H.Check("Reliability_Effect_MountedOnce", mountedCount == 1);
-            H.Check("Reliability_Effect_NoCleanupBeforeClose", cleanupCount == 0);
+            // Pump until the pane body realizes behind the TabView content
+            // presenter BEFORE probing the mount/cleanup counters. The
+            // nested component's OnMount effect — which increments
+            // mountedCount — only fires once the body is committed, and on
+            // the NativeAOT dispatcher-timing window that realization can
+            // land a wave later than Harness.Render drains. Gating the count
+            // asserts on the body sentinel closes the residual cascade where
+            // Reliability_Effect_MountedOnce probes a count of 0.
             H.Check("Reliability_Effect_BodyRendered",
                 await Harness.WaitFor(() => H.FindText("effect-body-p1") is not null));
+            H.Check("Reliability_Effect_MountedOnce", mountedCount == 1);
+            H.Check("Reliability_Effect_NoCleanupBeforeClose", cleanupCount == 0);
 
             var model = DockHostModelBridge.Get(managerEl);
             H.Check("Reliability_Effect_BridgeYieldsModel", model is not null);


azchohfi:
Closing without merging.

Investigation (prompted by review skepticism about a test-only reorder) showed this PR's change is **not load-bearing**:

- #445's documented `DockHooks_*` cascade was already resolved by #502 (commit 41bb2b42 — the `Harness.WaitFor` guard on the failing `alpha-host:ok` probe in `DockContextHooksResolveOnRealMount`) and #510 (swept the remaining NativeDocking probes). Both merged after #445 was filed.
- Fault injection (simulating the AOT dispatcher under-drain) proved the reliability fixture's `mountedCount` reaches 1 while the pane body is still absent from the visual tree (`mounted=1, bodyNull=True`). That counter is set at Reactor mount time — decoupled from the deferred TabView realization that `FindText` depends on — so gating the count asserts behind body realization does not change their outcome.
- `TwoPaneMountUpdateUnmount` (a static-audit candidate) isn't susceptible either: its DockSplit/FlexPanel bodies realize synchronously.

Full verification is on #445. No production or test change is warranted.

PR #738, opened 2026-06-27
perf(reconciler): SIMD-vectorize keyed-middle index shifts — recover #657 keyed-reconcile time, keep −30.9% alloc

azchohfi:
## What & why

Fix-forward for the keyed-reconcile **time** regression introduced by #657.

#657 shipped a **−30.8% keyed-list allocation** win, but its *correct* right-to-left LIS rewrite (which fixed the C1/H1 keyed-identity corruption) replaced H1's buggy O(1) survivor-index update with an **exact O(oldMidLen) shift-bookkeeping** pass per Insert/Move. Post-merge `/perf` on the keyed-list workload measured **reconcile +20.4%**, **diff +20.5%**, **rps −10.2%** (alloc still −30.9% ✅) — keyed-only (StocksGrid positional + all 17 micros flat). The cost lives entirely in those `oldRelToPanel` shift scans.

This PR **vectorizes those three shift loops** with `System.Numerics.Vector<int>` to recover the time while **keeping the −30.9% alloc win flat** (non-negotiable — main already carries it).

## Mechanism

`RunKeyedMiddleCore` keeps its `oldRelToPanel` survivor-position model exact by shifting a band of entries on every structural op. Each shift is a branchless *compare-and-conditional-±1* over an `int[]` — a textbook SIMD pattern. Three private helpers replace the scalar loops:

| Helper | Predicate (scalar) | Op |
|---|---|---|
| `BumpAtOrAfter(map,len,thr)` | `if (p >= thr) p++` | insertion shifts survivors ≥ anchor right |
| `DecrementWhereGtLe(map,len,lo,hi)` | `if (p > lo && p <= hi) p--` | left-to-right Move closes the gap over `(lo,hi]` |
| `IncrementWhereGeLt(map,len,lo,hi)` | `if (p >= lo && p < hi) p++` | right-to-left Move opens the gap over `[lo,hi)` |

Each applies the **EXACT** scalar predicate `Vector<int>.Count` lanes at a time (SSE/AVX2 on x64, NEON on ARM64) + a scalar remainder, and falls back to the verbatim scalar loop when `!Vector.IsHardwareAccelerated` or `len < width`. The masks exploit that `Vector.GreaterThan` et al. return all-ones (−1) in matching lanes, so `v - mask` adds 1 and `v + mask` subtracts 1 exactly where the predicate holds. Negative sentinels (−1 unrealized survivors) never satisfy any predicate → pass through untouched, matching scalar. Reads are bounded to `[0, len)` (`r ≤ len − width`), so the larger pooled tail is never touched.

## Behavior-preserving (hard requirement)

Only the **index oracle** is vectorized. The LIS computation, the move **set**, the move **order**, and patch-by-live-index are **bit-for-bit unchanged**. The corrected C1/H1 algorithm stays exactly as merged.

## Alloc-neutral (hard requirement)

Reuses the **same pooled `int[]`**; `Vector<int>` is a stack value type — zero new per-frame heap allocs. A new test asserts **0 B/iter** through the core via `GC.GetAllocatedBytesForCurrentThread`.

## Tests — `ChildReconcilerKeyedMiddleSimdTests.cs`

The correctness guard for a vectorized hot path is a **checked-in differential test**: it runs the production (SIMD) core and a hand-rolled **scalar reference** core over the same models and asserts the emitted **Mount/Move/Patch op stream + final cell model are identical**:

- **5000 random** keyed edit workloads (n≤80, churn 0–60%).
- **Steady-state ticks** at N=500 (pct 5/10/25) and N=2000 — the regressing workload, faithfully ported from `KeyedListSource`.
- **Large-N churn** at N=2000/4096/4099 — exercises full vector spans **+ remainder tails**.
- Explicit teeth: rotation `[a,b,c,d]→[b,c,d,a]` (the C1 repro), full reverse, interleave.
- The 4-invariant reorder oracle (length, prefix/suffix untouched, per-slot identity + content) on every case — independent ground truth, not just SIMD==scalar.
- Alloc-free assertion (0 B/iter).

The existing `ChildReconcilerKeyedMiddleCoreTests` (28 cases + rotation + dup teeth) now exercise the SIMD core through production and stay green unchanged.

## Why SIMD (and not Fenwick)

The fix-forward was originally scoped to an O(log N) Fenwick/BIT index oracle. I built a self-contained differential microbench (4 cores: array / Fenwick / SIMD / hybrid — all proven bit-for-bit identical) and **measured** it: the pure Fenwick is **slower at the N=500 target** (the `oldRelToPanel` array is L1-resident and the shift loops are branch-predictable sequential scans; Fenwick's log²N searches don't pay off until N≈2000, beyond Reactor's realistic non-virtualized keyed-list sizes). SIMD is **~65% faster at N=500** and faster at every realistic size, with the lowest correctness risk (a pure micro-opt of the already-proven in-main shift logic).

Bookkeeping-only microbench (µs/tick, ARM64 NEON 4-wide; x64 AVX2 runner is 8-wide → larger margin):

| workload | array | fenwick | **simd** |
|---|---|---|---|
| N=200 pct=10 | 7.7 | 19.7 (+158%) | 7.5 (−2%) |
| **N=500 pct=10** ★ | 16.0 | 21.7 (+35%) | **5.6 (−65%)** |
| N=500 pct=25 | 38.1 | 40.4 (+6%) | 11.2 (−71%) |
| N=1000 pct=10 | 52.3 | 50.5 (−3%) | 19.7 (−62%) |
| N=2000 pct=10 | 197.4 | 110.8 (−44%) | 70.5 (−64%) |

(Bench source archived in the session for traceability.)

## `/perf` success criteria (coordinator measures end-to-end)

Keyed **reconcile/diff substantially recovered** toward flat (a small residual is acceptable — the bench is bookkeeping-only and SIMD recovers ~62–65% of it on NEON, more on the AVX2 runner), **rps substantially recovered**, and keyed **alloc stays ~flat (−30.9% preserved)**. StocksGrid + all 17 micros flat.

## Gates (ARM64 local)

- `dotnet build src/Reactor/Reactor.csproj -c Release` = **0 warnings / 0 errors** (AOT/trim-clean).
- `dotnet test tests/Reactor.Tests` = **9899 passed / 0 failed / 64 skipped** (incl. the new differential suite + existing keyed teeth).

## Scope

`src/Reactor/Core/ChildReconciler.cs` (+101/−12) + the new test file only.

---

**DRAFT** — do not merge. Held for dual review + the `/perf` gate per the fix-forward plan.

### DIFF
--- a/src/Reactor/Core/ChildReconciler.cs
+++ b/src/Reactor/Core/ChildReconciler.cs
@@ -1,5 +1,6 @@
 using System.Buffers;
 using System.Collections.Concurrent;
+using System.Numerics;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Reactor.Core.Internal;
 using WinUI = Microsoft.UI.Xaml.Controls;
@@ -569,8 +570,7 @@ internal static class ChildReconciler
                 {
                     // Insertion at `anchor` shifts every tracked survivor at an
                     // index >= anchor one slot to the right.
-                    for (int r = 0; r < oldMidLen; r++)
-                        if (oldRelToPanel[r] >= anchor) oldRelToPanel[r]++;
+                    BumpAtOrAfter(oldRelToPanel, oldMidLen, anchor);
                     // The new control now occupies `anchor` and is the anchor for
                     // the next (left) item, so `anchor` itself is unchanged.
                 }
@@ -594,19 +594,13 @@ internal static class ChildReconciler
                     // between the two endpoints by one slot.
                     if (cur < to)
                     {
-                        for (int r = 0; r < oldMidLen; r++)
-                        {
-                            int p = oldRelToPanel[r];
-                            if (p > cur && p <= to) oldRelToPanel[r] = p - 1;
-                        }
+                        // Survivors in (cur, to] shift one left to close the gap.
+                        DecrementWhereGtLe(oldRelToPanel, oldMidLen, cur, to);
                     }
                     else
                     {
-                        for (int r = 0; r < oldMidLen; r++)
-                        {
-                            int p = oldRelToPanel[r];
-                            if (p >= to && p < cur) oldRelToPanel[r] = p + 1;
-                        }
+                        // Survivors in [to, cur) shift one right to open the gap.
+                        IncrementWhereGeLt(oldRelToPanel, oldMidLen, to, cur);
                     }
                     oldRelToPanel[oldRel] = to;
                     cur = to;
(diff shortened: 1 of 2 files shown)


azchohfi:
Closing — accepted outcome, not a defect.

This SIMD fix is **correct** (bit-for-bit vs a scalar reference over 5000 random + Vector-width remainder-tail cases; 3 independent review legs) and **alloc-neutral** (0 B/iter; #657's −30.9%/−98K keyed alloc win preserved). But its /perf shows **zero keyed-time recovery** (reconcile +0.8%, within-noise).

Empirical diagnosis proved why: the keyed-reconcile cost center is the **move issuance** (`IChildCollection.Move` = `UIElementCollection.RemoveAt+Insert` = O(|from−to|) native array-shift + COM), **not** the `oldRelToPanel` bookkeeping this PR vectorized (~0.4% of reconcile, so even eliminating it recovers nothing).

The +18.5% keyed-reconcile time #657 added is the **intrinsic price of correctness**: the C1/H1-correct keyed reorder structurally requires the more-expensive right-to-left anchor moves (avg array-shift 524 vs the prior buggy left-to-right's 433). The cheaper moves are provably wrong every tick — the decisive `lr-corr` experiment (correct-oracle + LIS-minimal + cheap L→R targets) is still mis-ordered 300/300. The pre-#657 speed was a bug artifact.

**Decision:** accept the trade — keep #657's −30.9% alloc win (the deliverable, unaffected) + the correctness; accept the ~0.77ms/render keyed-time cost on heavy reordering. This SIMD adds hot-path complexity for no measurable gain, so it is not merged. The differential test + the diagnosis data are archived. The only theoretical recovery lever (a non-LIS reconciler minimizing total array-shift *distance* rather than move *count*) is research-grade and out of scope — filed as a perf-backlog note.

azchohfi:
**Mechanism refinement (post-close, for the record).** My close note above attributed the cost to the more-expensive (longer-displacement) right-to-left moves. A follow-up experiment refined that: a **correct left-to-right mirror traversal (`la-anchor`)** exists with the same minimal move count and −20.6% *less* displacement (avg-shift 416 vs #657's 524), and when measured it recovers **zero** time (7.53 vs 7.57ms). So displacement is a red herring — the real cost is the **count** of WinUI `Move`/`Insert`/`Remove` COM ops, which LIS already minimizes and which every correctness-preserving traversal shares. #657 is on the correct-minimal-move floor regardless of direction; no correct in-scope variant beats it (la-anchor was measured — it's null). Decision unchanged. Full detail in #741.

Issue #671, opened 2026-06-25
perf: stabilize DataGrid cell/row modifier-handler delegates (.OnTapped/.OnPointerPressed) for the post-#665 skip path


azchohfi:
Follow-up to #663 / #669 (DataGrid per-render allocation + LINQ caching). **Deferred / gated — blocked on P1 FLAGSHIP-1 (#665). Do not implement until the conditions below are met.**

## Summary
The DataGrid attaches three **routed-input idiom-A _modifier_ handlers** (`.OnTapped` / `.OnPointerPressed`, dispatched via the `ModifierEventHandlerState` routed-input family) to its cells/rows, each built as a **fresh per-render closure** inside the static, virtualized `RenderRow`. Once **#665 (FLAGSHIP-1)** relaxes modifier-handler comparison to **per-slot reference-equality**, these per-render closures become the **only remaining blocker** to the Element-layer skip path for grid cells/rows. This issue tracks stabilizing those three delegates so cells/rows can hit the skip (Update-free) path.

## Why this is blocked on #665 (not actionable today)
`Element.ModifiersEqual` (src/Reactor/Core/Element.cs:1062-1067) currently treats **any** non-null handler modifier as an automatic inequality — both sides must be `null` to match:

```csharp
&& a.OnPointerPressed is null && b.OnPointerPressed is null
&& a.OnTapped is null && b.OnTapped is null
```

So on `main`/today, **every DataGrid cell with `.OnTapped` and every row with `.OnPointerPressed` already fails `ShallowEquals` → `CanSkipUpdate` returns false → full `Update` every render, regardless of closure identity.** Hand-stabilizing the delegates now buys nothing — the comparison short-circuits before delegate identity is ever consulted.

**#665 (FLAGSHIP-1)** relaxes this to per-slot `ReferenceEquals(a.OnTapped, b.OnTapped)`. _After_ #665 lands, a cell/row can skip **iff** its handler delegate is reference-stable across renders. Because the DataGrid rebuilds these closures every render, they'd fail `ReferenceEquals` and still block the skip — hence this follow-up. **This must NOT ship before #665.**

## In scope — affected handlers (exhaustive: 3 sites)
All are fluent routed-input modifier handlers (idiom-A) attached inside the static `RenderRow` (run by `VirtualList.renderItem`), all capturing load-bearing per-row state:

| # | Site | Modifier | Purpose | Captures |
|---|------|----------|---------|----------|
[...]

PR #754, opened 2026-06-29
perf: cache per-render arrays/LINQ in DataGrid (#669)

### LINKED ISSUE #663: perf: cache per-render arrays/LINQ in DataGrid
## Summary

On the data-grid stress workload (`tests/stress_perf/StressPerf.ReactorGrid` — the real `DataGridComponent` benchmark, 30 cols × 160 rows), the DataGrid rebuilds arrays and re-runs LINQ on **every render** even when the underlying sort/filter/column/width state is unchanged. These per-render allocations are first-order contributors to the C# vs Rust Renders/sec gap (C# ~4.1 vs Rust ~8.7) and the Avg Reconcile (~46ms vs ~7.9ms) / Avg Diff (~39ms vs ~7.1ms) / Avg Memory (~285MB vs ~190MB) deltas.

Scope: `src/Reactor/Controls/DataGrid/` only (`DataGridComponent.cs`, `DataGridState.cs`).

## Findings

| # | Location | Issue | Fix |
|---|----------|-------|-----|
| 123 | `DataGridComponent.cs:224` | `sortKey` built via `Select` + interpolation + `string.Join` every render | `UseMemo` keyed on `state.SortVersion` |
| 124 | `DataGridComponent.cs:183-188` | `state.Sorts.ToList()` + `Filters.ToList()` build a new `DataRequest` every render (and the hook's deps compare `Sort`/`Filters` list **references**, so pagination restarts every render) | `UseMemo` keyed on `SortVersion`/`FilterVersion`/`SearchQuery`/`pageSize` |
| 125 | `DataGridComponent.cs:428-448,893-900` | `new double[]` + `new string[]` + per-column `double.ToString` every render (header **and** rows) | cache `colWidths`/`gridColDefs` in state keyed on a column-layout version; return a stable cached `GridDefinition` so the reconciler skips re-applying ColumnDefinitions |
| 126 | `DataGridComponent.cs:517` | `new Element?[]` per row per render | **Investigated — not pooled.** `FilterChildren`'s fast path aliases the array into `GridElement.Children`, which the reconciler retains for the next diff (and VirtualList retains per realized row). Returning it to `ArrayPool` would alias a live array → corruption. The array is the element's backing store, not a transient buffer, so a defensive copy would negate the benefit. See PR notes. |
| 127 | `DataGridState.cs:130` | `Columns` getter does `Where + ToList` on **every** access | maintain `_visibleColumns` synced by Hide/Show with a dirty-flag rebuild |
| 128 | `DataGridState.cs:471,502,618,695` | `GetSortDirection`/`GetFilter`/`GetColumnWidth`/`PinColumn` use LINQ `FirstOrDefault`/`FindIndex` per column per render | `Dictionary` O(1) lookups kept in sync with the lists |
| 129 | `DataGridComponent.cs:313-315` | `new string[gridRow+1]` row-def arrays filled every render | static cached row-def arrays for the 4 combinations |
| 130 | `DataGridState.cs:539` | `SelectRange` materializes the entire `_rowKeyCache` as `List<RowKey>` on shift-click (100k rows) | index-based scan overload with no materialization |
| 131 | `DataGridComponent.cs:330-373,747-757` | Setters spread array rebuilt every render though `lostFocusWired` runs once; per-row validation LINQ + Join | `UseRef`-cached setter array; cached validation summary on state |

## Correctness

Behavior (sorting, filtering, pagination, column show/hide/pin/resize, range selection) must be **identical**. Every mutation path bumps the matching version counter / dirty flag so caches invalidate correctly:

- `_sorts` (ToggleSort) → `SortVersion`
- `_filters` (SetFilter/ClearFilter/ClearAllFilters) → `FilterVersion`
- `_columns`/`_hiddenColumns`/`_columnWidths` (Reorder/Hide/Show/Toggle/Pin/Resize) → `_columnVersion` + `_visibleColumnsDirty` + synced `_columnIndexByName`
- validation summary keyed on `_editingVersion`

## Expected impact

Removes, per render: 2× `string[ColCount]` + `double[ColCount]` + ~2×ColCount `double.ToString` (header + data layout), the `sortKey` join, and (hook path) the `DataRequest` + two `.ToList()`s; converts the per-column header/row lookups and the `Columns` getter from O(n)·LINQ to O(1); removes the shift-click `_rowKeyCache` materialization. The cached `GridDefinition` reference additionally lets the reconciler skip re-creating ColumnDefinitions for the header and every data row, which is expected to move Renders/sec, Avg Reconcile/Diff, and Avg Memory toward the Rust baseline.

## Validation

`dotnet build Reactor.slnx -c Release` + `dotnet test tests/Reactor.Tests`, plus new xUnit coverage for cache invalidation, O(1)-lookup parity, and `SelectRange` parity. Perf validated via `/perf` once that workflow merges.

azchohfi:
Closes #669. Closes #663. Touches #671 (per-render delegate identity).

Fresh-branch re-application of the #669 caching work, rebased cleanly onto **current `main`** (`56ee84a0`). The prior PR branch (`azchohfi-datagrid-perf-cache`) was based on the pre-`#746` tree; this branch carries the identical, already-PR-reviewed implementation on top of the merged `StressPerf.DataGrid` perf leg so `/perf` can measure it directly.

## Why this is measurable now
The merged `#746` leg (`StressPerf.DataGrid`) drives a real **30-col × 200-row** `DataGridComponent` whose full `Render()` runs every frame — the exact per-render path #669 targets. Baseline on `main` is ~**10.86 MB alloc/render**; the prior interleaved `/perf` on this implementation showed **alloc −7.2%**, **Gen0 −13.5%**, **renders/sec +20.5%** on the DataGrid leg.

## Approach (all edits confined to `Controls/DataGrid/**` + a new test)
**`DataGridState.cs`**
- **Version counters** (`SortVersion`/`FilterVersion`/`ColumnVersion`) bumped by every mutation path; no-op mutations don't bump (preserves `StateChanged` semantics).
- **O(1) dictionary lookups** (`_sortDirByField`/`_filterByField`/`_columnIndexByName`) replace per-column `FirstOrDefault`/`FindIndex` LINQ.
- **`Columns`** serves a version-keyed visible-column snapshot instead of `Where+ToList` every access.
- **`GetColumnLayout(...)`** caches per-column widths + a **single shared `GridDefinition`** (keyed on `ColumnVersion` + shape + column ref/count); header row and every data row reuse the one reference so the reconciler skips re-applying `ColumnDefinitions`.
- **`SelectRangeByKeyCache`** scans the internal row-key `string[]` by index instead of materializing a `List<RowKey>` on shift-click.

**`DataGridComponent.cs`**
- **Memoize the `DataRequest`** and the **sort key** with `UseMemo` keyed on the version counters (also fixes a latent spurious per-render pagination restart from fresh `Sorts.ToList()`/`Filters.ToList()` refs).
- **Root grid + data/header rows** build `GridElement`s directly from cached `GridDefinition`s (static root-row-def cache + shared column layout) instead of allocating `string[]` + `double[]` + per-column `double.ToString` each render. `FilterRowChildren` mirrors the private `Dsl.FilterChildren` child semantics exactly.
- **Stabilize the once-wired `LostFocus` setter array** and the **search-box `onChanged` delegate** in `UseRef`s, dropping per-render closure allocations (#671).

## Ownership boundary
No edits to `Core/Reconciler*.cs`, `Hooks/**`, `RenderContext.cs`, or `Theme.cs`. Public API surface unchanged (new members are `internal`).

## Validation
- `dotnet build src/Reactor/Reactor.csproj -c Release` — **0 warnings, 0 errors**.
- `dotnet test tests/Reactor.Tests -p:Platform=x64` — **9931 passed, 0 failed, 64 skipped**, including **18 `DataGridPerfCacheTests`** (cache invalidation on sort/filter/column/width change; O(1) lookups match the old LINQ; shared-layout reference stability; visible-columns cache; `SelectRangeByKeyCache` parity).

Draft pending the `/perf` comparison on the DataGrid allocation table + renders/sec.

### DIFF
--- a/src/Reactor/Controls/DataGrid/DataGridComponent.cs
+++ b/src/Reactor/Controls/DataGrid/DataGridComponent.cs
@@ -179,13 +179,18 @@ public class DataGridComponent<[DynamicallyAccessedMembers(DynamicallyAccessedMe
             var rowH = el.RowHeight ?? el.EstimatedRowHeight;
             var pageSize = Math.Max(50, (int)Math.Ceiling(2160.0 / rowH));
 
-            var request = new DataRequest
+            // Memoize the DataRequest on the sort/filter/search/pageSize identity. Rebuilding it
+            // every render allocated two fresh List copies (Sort/Filters) whose references then
+            // changed UseDataSource's deps every render, restarting pagination. Keyed on the
+            // version counters, the request and its list references stay stable until the
+            // underlying sort/filter/search actually change.
+            var request = UseMemo(() => new DataRequest
             {
                 PageSize = pageSize,
                 Sort = state.Sorts.Count > 0 ? state.Sorts.ToList() : null,
                 Filters = state.Filters.Count > 0 ? state.Filters.ToList() : null,
                 SearchQuery = state.SearchQuery,
-            };
+            }, state.SortVersion, state.FilterVersion, state.SearchQuery ?? string.Empty, pageSize);
 
             var resource = UseDataSource(
                 source,
@@ -221,7 +226,11 @@ public class DataGridComponent<[DynamicallyAccessedMembers(DynamicallyAccessedMe
 
         // Load data on mount and when sort changes (legacy path only — hook path
         // reacts to sort/filter changes through its own deps).
-        var sortKey = string.Join(",", state.Sorts.Select(s => $"{s.Field}:{s.Direction}"));
+        // Memoize the sort key so the Select + interpolation + Join only runs when sorts change,
+        // not on every render. It feeds the LoadDataAsync effect's deps below.
+        var sortKey = UseMemo(
+            () => string.Join(",", state.Sorts.Select(s => $"{s.Field}:{s.Direction}")),
+            state.SortVersion);
 
         UseEffect(() =>
         {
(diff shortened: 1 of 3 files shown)


azchohfi:
Closing as a duplicate. #669 is already covered by the original PR (#669, branch azchohfi-datagrid-perf-cache) with the already-reviewed implementation, update-branched onto post-#746 main with a completed /perf run. This branch was a redundant fleet spin-up (the original session was mis-read as frozen) and carries the same diff — closing to avoid a competing PR. Branch left un-merged for the coordinator to clean up.

Issue #327, opened 2026-05-18
Cross-container row memoization for VirtualList — opt-in API to collapse per-row reconcile cost


codemonkeychris:
## Background

After the `ItemsRepeater` recycling fix (#324), profiling the `DataSystemDemo → VirtualList (Variable Height)` page with a 10 s scroll capture (`reactor-scroll-2.etl`, post-#324) shows:

- **11,211 Reactor reconciles in 10.5 s = 1,069/sec** during scroll
- Per-reconcile cost: avg 334 µs, p95 370 µs, p99 768 µs, max 23.8 ms
- **Total reconcile time = 3.75 s = 35.7 % of trace wall clock**
- 0 GCs (allocation pressure already eliminated by #324)
- Almost all element-diffs (96 %) hit the `Update` path because each row has different content baked into it (`#{i:N0}` text, etc.); only ~4 % hit the existing `ShallowEquals` skip-path

The scroll-time reconciles are already as tightly scoped as they can be — one reconcile per realized row, not a full scene pass. The cost is unavoidable *given the current contract*: every time `ItemsRepeater` recycles a container, `ElementFactory<T>.GetElement` (`src/Reactor/Core/ElementFactory.cs:204-216`) rebuilds the row's `Element` tree from scratch via `_viewBuilder(item, index)`, then diffs the new tree against the recycled container's previous tree. Same `i` scrolling back into view → identical tree, fully rebuilt, fully diffed.

## What's *not* a fix

The existing `Memo(Func<RenderContext, Element>, params object?[] deps)` factory (`src/Reactor/Elements/Dsl.cs:777`) doesn't help here. `Memo` is **per-component-node** — when a container recycles from row 130 to row 47, the deps change, the memo invalidates, the tree re-renders. The memoization we'd want is **cross-container**: "row 47's tree is the same instance no matter which physical container is currently showing it."

Framework-level automatic memoization (cache `renderItem(i)` results inside `ElementFactory<T>` keyed by `ReactorRow.Key`) is also rejected as a path. It's too easy for `renderItem` to close over external state that the framework can't see — silent staleness would be worse than today's reliable-but-slow path.

## What developers can do today

[...]

PR #755, opened 2026-06-29
Cross-container row memoization for VirtualList — opt-in cacheRowsBy flag (Option B, #327)

azchohfi:
Alternative to #753 — same goal (issue #327), different shape. Opened as a **draft for perf/design comparison**; only one of the two should merge.

## What

A list-level, **opt-in** row-memoization flag on `VirtualList` — no new element type, no reconciler/factory changes:

```csharp
VirtualList(
    itemCount: count,
    renderItem: i => Border( /* … */ ),
    cacheRowsBy: i => items[i].Id.ToString(),   // null = current behavior (opt-in)
    estimatedItemHeight: 64)
```

New surface on `VirtualListElement`: `Func<int,string>? CacheRowsBy`, `int RowCacheCapacity = 128`; matching optional params on the `VirtualList(...)` factory.

## Why

Same root cause as #327/#753: on the VirtualList `int`-index path, `ElementFactory.BuildOrCache`'s `ReferenceEquals(item)` guard always misses (the index is re-boxed each recycle), so every container recycle rebuilds + re-diffs the row tree (~36% of scroll wall-clock). Returning the **same `Element` instance** per key makes `Element.ShallowEquals` short-circuit the per-row reconcile to a `CanSkipUpdate → null` skip.

## How

When `CacheRowsBy` is set, `VirtualListComponent` wraps its view builder with a per-instance **bounded LRU** (`RowMemoCache`, default cap 128) keyed by the selector. A hit returns the cached instance (reconciler skips); a miss builds once and stores. The cache lives in a `UseMemo` cell keyed on `(ItemCount, RenderItem identity, CacheRowsBy identity, fixedHeight, capacity)`, so any change — including a fresh `renderItem` closure — drops the whole cache and can't serve stale instances. Scroll-driven realize/recycle does not re-run `Render`, so the cache persists across container cycles (the hot path). `.Height(...)` is applied **before** caching so the cached instance stays reference-stable.

## Trade-offs vs #753 (Option A)

- **Pro:** zero core changes (no `Element.cs`/`Reconciler.cs` edits); declarative one-line call-site; lowest blast radius.
- **Con:** VirtualList-only (A also covers LazyVStack/HStack, ItemsView, DataGrid); whole-row only (A can memoize a sub-tree).

## Tests & docs

- 15 new headless cases in `VirtualListRowCacheTests` (cross-recycle effectiveness 480→24 rebuilds, same-instance/`CanSkipUpdate` proof, key-change-not-stale guard, LRU eviction, cache reset on count/closure change, default-path-unchanged, fixed-height wrapping order). `dotnet test tests/Reactor.Tests` green; core builds 0 warnings/0 errors (AOT/trim-clean).
- Guide updated (`collections.md.dt` → regenerated `collections.md`), documenting the flag **and** the user-land `Dictionary` escape hatch.

### DIFF
--- a/docs/_pipeline/templates/collections.md.dt
+++ b/docs/_pipeline/templates/collections.md.dt
@@ -330,6 +330,86 @@ return LazyVStack<Note>(notes, n => n.Id, (note, i) => Row(note));
 You then own the parts `Memo` handles for you: bound the dictionary (evict so
 it can't grow without limit) and drop or rebuild an entry when the data behind
 its key changes — the purity contract above, enforced by hand.
+### Cross-recycle row memoization (`cacheRowsBy`)
+
+Variable-height rows have a hidden scroll cost. With no `itemHeight`, every
+time `ItemsRepeater` recycles a container and re-realizes a row, the list
+rebuilds that row's element tree from `renderItem` and the reconciler diffs
+it against the previous tree. During a fast flick over tall rows that
+rebuild-and-diff can dominate scroll wall-clock. `cacheRowsBy` is an opt-in
+that collapses it:
+
+```csharp
+VirtualList(
+    itemCount: rows.Count,
+    renderItem: i => RenderRow(rows[i]),      // builds a multi-node subtree
+    getItemKey: i => rows[i].Id,
+    estimatedItemHeight: 64,
+    cacheRowsBy: i => rows[i].Id)             // ← opt in
+```
+
+When set, the list keeps a bounded **LRU cache of already-built row
+`Element` instances**, keyed by `cacheRowsBy(index)`. On a recycle for a key
+still in the cache it returns the *same* `Element` instance instead of calling
+`renderItem` again — the reconciler then short-circuits on `ReferenceEquals`
+and skips the entire per-row diff. The rebuild *and* the reconcile descent
+both go to ~zero on a cache hit.
+
+<!-- ai:caveat -->
+`cacheRowsBy` is a **purity assertion you are making**, not a transparent
+optimization. By opting in you promise that `renderItem` is a pure function of
+the inputs your key captures: for a given key, the element it produces must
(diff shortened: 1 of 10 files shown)


azchohfi:
Closing this in favor of #753 (Option A, `Memo<TKey>(key, factory)`), which merged to `main`. Keeping the branch here as a reference for the alternative design.

**Why A over B:** both collapse the fast-scroll per-row reconcile to a `ReferenceEquals` skip and both are big wins. A's cache lives one layer deeper (in `ElementFactory.BuildOrCache`, keyed by value), so it works for **every** `ElementFactory`-backed list — `VirtualList`, `LazyVStack`/`HStack`, `ItemsView`, `ItemsRepeater`, and **DataGrid** rows — and can memoize a sub-tree. B (`cacheRowsBy`) is a per-`VirtualList` flag: zero core changes and the lightest allocation, but VirtualList-only and whole-row only.

**Perf, measured on top of `main` (which already has A merged), all three arms in one process through the real `ElementFactory.BuildOrCache` recycle path — 9-node variable-height row, 1,000,000 recycles/arm, Release:**

| Arm | ns/recycle | bytes/recycle | vs baseline |
|---|--:|--:|--:|
| Baseline (no cache) | 316.7 | 4,568 | 1.0× |
| A — `Memo(i, …)` | 31.8 | 168 | 10.0× faster (27× less alloc) |
| B — `cacheRowsBy` | 41.0 | 0 | 7.7× faster (∞ less alloc) |

Both reach `CanSkipUpdate == true` (reconcile descent skipped). A is a touch faster per recycle; B allocates literally 0 bytes/recycle on a steady-state hit (lightest GC pressure). In practice they tie — A shipped for its broader reach.

The branch is rebased onto current `main` and left intact for anyone who wants the VirtualList-only, zero-allocation variant. Not merging.

PR #763, opened 2026-06-30
Add analyzer for event subscriptions inside .Set(...)

### LINKED ISSUE #762: Analyzer: warn on event subscriptions inside .Set(...) when a declarative On* modifier exists
## Summary

Reactor already exposes declarative `On*` event modifiers that are render-safe, but it is still easy to write event subscriptions imperatively inside `.Set(...)`:

```csharp
Border().Set(border =>
{
    border.Tapped += OnTapped;
    border.DragOver += OnDragOver;
    border.Drop += OnDrop;
});
```

That pattern is fragile because `.Set(...)` runs on mount and on later updates, so `+=` can accumulate duplicate handlers across renders.

For many of these events, Reactor already has declarative equivalents such as `.OnTapped(...)`, `.OnDragOver(...)`, `.OnDrop(...)`, `.OnKeyDown(...)`, `.OnPointerEntered(...)`, `.OnDoubleTapped(...)`, and `.OnSizeChanged(...)`.

## Why this seems like an analyzer gap

The current `.Set(...)` analyzer (`REACTOR_POOL_001`) only handles simple assignment expressions and does not inspect event subscription patterns inside block lambdas.

So code like this can slip through today even though a safer Reactor-native modifier exists.

## Proposed rule

Add a diagnostic for `.Set(...)` lambdas when all of the following are true:

1. The lambda subscribes to an event on the `.Set(...)` parameter via `param.Event += handler`.
2. Reactor exposes a matching declarative `On*` modifier for that event.
3. The usage is inside `.Set(...)`, not inside `.OnMount(...)` or another one-time imperative setup path.

Example:

```csharp
Border().Set(border => border.Tapped += OnTapped)
```

should warn and steer the caller toward:

```csharp
Border().OnTapped(OnTapped)
```

## Suggested scope for v1

- Flag direct `param.Event += handler` inside `.Set(...)` when a matching `OnEvent` modifier exists.
- Optionally offer a code fix for the simple single-handler case.
- Keep the rule narrowly scoped to the `.Set(...)` lambda parameter to avoid broad false positives.

## Non-goals / caution

- Do not flag `.OnMount(...)` patterns. One-time imperative hookup is still a valid escape hatch.
- `AddHandler(...)` probably needs separate treatment, especially when `handledEventsToo: true` changes semantics.
- Events with no Reactor declarative equivalent should not be flagged.
- Balanced unsubscribe/resubscribe patterns could be handled separately or at lower severity.

## Why this would help

This would catch a real behavioral bug pattern, not just a style preference. It would also reinforce Reactor's declarative event model in exactly the cases where the framework already provides a safe equivalent.

nsreehari:
## Summary

Adds `REACTOR_EVENT_001` for a specific Reactor footgun: subscribing to UI events inside `.Set(...)` when Reactor already provides a declarative `On*` modifier for the same event.

The problem is behavioral, not stylistic. `.Set(...)` can run again on later updates, so `param.Event += handler` can accumulate duplicate handlers across rerenders. The declarative modifier path is render-safe because the reconciler owns detach/reattach behavior.

## Scope

This PR intentionally stays narrow:

- warns on direct `.Set(fe => fe.Event += handler)` subscriptions on the `.Set(...)` lambda parameter
- covers the safe declarative event surface already exposed by `ElementExtensions.cs`
- does not flag `.OnMount(...)`
- does not flag unrelated/member-chain subscriptions
- keeps the code fix narrower than the analyzer: simple signature-compatible event cases auto-rewrite, while drag/drop-style abstractions warn but do not auto-rewrite

## Included changes

- add `SetEventSubscriptionAnalyzer` (`REACTOR_EVENT_001`)
- add `SetEventSubscriptionCodeFix` for the safe rewrite subset
- add focused analyzer tests for positive, negative, and code-fix behavior
- add a consistency test so the tracked event map cannot drift from `src/Reactor/Elements/ElementExtensions.cs`
- register the rule in `AnalyzerReleases.Unshipped.md`

## Validation

- `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter "SetEventSubscriptionAnalyzerTests|SetEventSubscriptionConsistencyTests"`

Closes #762.

### DIFF
--- a/src/Reactor.Analyzers/AnalyzerReleases.Unshipped.md
+++ b/src/Reactor.Analyzers/AnalyzerReleases.Unshipped.md
@@ -18,5 +18,6 @@ REACTOR_A11Y_003 | Microsoft.UI.Reactor.Accessibility | Warning | AccessibilityA
 REACTOR_REF_001 | Reactor.Reference | Warning | ReferenceCurrentReadAnalyzer - Use descriptor.Reference/binding.Reference instead of assigning ElementRef.Current to reference properties
 REACTOR_DSL_001 | Reactor.Dsl | Warning | MissingWithKeyAnalyzer - Dynamic list item missing .WithKey
 REACTOR_DOCK_001 | Reactor.Docking | Warning | OnLiveLayoutRoundTripAnalyzer - OnLiveLayoutChanged feeds the live layout back into state
+REACTOR_EVENT_001 | Reactor.Events | Warning | SetEventSubscriptionAnalyzer - .Set subscribes to an event that already has a declarative On* modifier
 REACTOR_POOL_001 | Reactor.Pool | Warning | PoolResetSetAnalyzer - .Set assigns to a property reset on pool return; use the surviving Reactor modifier
 REACTOR0050 | Reactor.Descriptor | Warning | OneWayClearValueAnalyzer - Optional<T> OneWay descriptor entries should provide dp: for ClearValue fallback
--- a/src/Reactor.Analyzers/SetEventSubscriptionAnalyzer.cs
+++ b/src/Reactor.Analyzers/SetEventSubscriptionAnalyzer.cs
@@ -1,0 +1,197 @@
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.UI.Reactor.Analyzers;
+
+/// <summary>
+/// REACTOR_EVENT_001: Detects <c>.Set(fe =&gt; fe.Event += handler)</c> patterns
+/// where Reactor already exposes a declarative <c>.OnEvent(...)</c> modifier.
+/// Imperative subscriptions inside <c>.Set(...)</c> can accumulate duplicate
+/// handlers as the element rerenders.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class SetEventSubscriptionAnalyzer : DiagnosticAnalyzer
+{
+    public const string DiagnosticId = "REACTOR_EVENT_001";
+
+    public static readonly IReadOnlyDictionary<string, string> EventModifiers =
+        new Dictionary<string, string>(System.StringComparer.Ordinal)
+        {
+            { "SizeChanged", "OnSizeChanged" },
+            { "PointerPressed", "OnPointerPressed" },
+            { "PointerMoved", "OnPointerMoved" },
+            { "PointerReleased", "OnPointerReleased" },
+            { "Tapped", "OnTapped" },
+            { "KeyDown", "OnKeyDown" },
+            { "PointerEntered", "OnPointerEntered" },
+            { "PointerExited", "OnPointerExited" },
+            { "PointerCanceled", "OnPointerCanceled" },
+            { "PointerCaptureLost", "OnPointerCaptureLost" },
+            { "PointerWheelChanged", "OnPointerWheelChanged" },
+            { "DoubleTapped", "OnDoubleTapped" },
+            { "RightTapped", "OnRightTapped" },
+            { "Holding", "OnHolding" },
+            { "KeyUp", "OnKeyUp" },
+            { "PreviewKeyDown", "OnPreviewKeyDown" },
+            { "PreviewKeyUp", "OnPreviewKeyUp" },
+            { "CharacterReceived", "OnCharacterReceived" },
+            { "AccessKeyDisplayRequested", "OnAccessKeyDisplayRequested" },
+            { "GotFocus", "OnGotFocus" },
+            { "LostFocus", "OnLostFocus" },
+            { "DragEnter", "OnDragEnter" },
+            { "DragOver", "OnDragOver" },
+            { "DragLeave", "OnDragLeave" },
+            { "Drop", "OnDrop" },
+        };
+
+    public static readonly IReadOnlyDictionary<string, string> CodeFixableEventModifiers =
+        new Dictionary<string, string>(System.StringComparer.Ordinal)
+        {
+            { "SizeChanged", "OnSizeChanged" },
+            { "PointerPressed", "OnPointerPressed" },
+            { "PointerMoved", "OnPointerMoved" },
+            { "PointerReleased", "OnPointerReleased" },
+            { "Tapped", "OnTapped" },
+            { "KeyDown", "OnKeyDown" },
+            { "PointerEntered", "OnPointerEntered" },
+            { "PointerExited", "OnPointerExited" },
+            { "PointerCanceled", "OnPointerCanceled" },
+            { "PointerCaptureLost", "OnPointerCaptureLost" },
+            { "PointerWheelChanged", "OnPointerWheelChanged" },
+            { "DoubleTapped", "OnDoubleTapped" },
+            { "RightTapped", "OnRightTapped" },
+            { "Holding", "OnHolding" },
+            { "KeyUp", "OnKeyUp" },
+            { "PreviewKeyDown", "OnPreviewKeyDown" },
+            { "PreviewKeyUp", "OnPreviewKeyUp" },
+            { "CharacterReceived", "OnCharacterReceived" },
+            { "AccessKeyDisplayRequested", "OnAccessKeyDisplayRequested" },
+            { "GotFocus", "OnGotFocus" },
+            { "LostFocus", "OnLostFocus" },
+        };
+
+    private static readonly LocalizableString Title =
+        "Use declarative event modifier instead of .Set subscription";
+
+    private static readonly LocalizableString MessageFormat =
+        "'{0}' subscribed inside '.Set(...)' may accumulate across renders. Use '.{1}(...)' modifier instead.";
+
+    private static readonly LocalizableString Description =
+        "Reactor reruns '.Set(...)' during later updates. Direct event subscriptions " +
+        "inside that lambda can accumulate duplicate handlers across renders. When " +
+        "Reactor already exposes a declarative On* modifier for the event, prefer " +
+        "that render-safe modifier instead.";
+
+    private static readonly DiagnosticDescriptor Rule = new(
+        DiagnosticId,
+        Title,
+        MessageFormat,
+        "Reactor.Events",
+        DiagnosticSeverity.Warning,
+        isEnabledByDefault: true,
+        description: Description);
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
+        ImmutableArray.Create(Rule);
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.EnableConcurrentExecution();
+        context.RegisterSyntaxNodeAction(AnalyzeInvocation, SyntaxKind.InvocationExpression);
+    }
+
+    private static void AnalyzeInvocation(SyntaxNodeAnalysisContext context)
+    {
+        var invocation = (InvocationExpressionSyntax)context.Node;
+
+        if (!TryGetSetInvocationData(invocation, out var setTarget, out var parameterName, out var assignment))
+            return;
+        if (assignment.Kind() != SyntaxKind.AddAssignmentExpression)
+            return;
+        if (assignment.Left is not MemberAccessExpressionSyntax leftAccess)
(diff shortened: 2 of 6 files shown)


src/Reactor.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -18,5 +18,6 @@ REACTOR_A11Y_003 | Microsoft.UI.Reactor.Accessibility | Warning | AccessibilityA
 REACTOR_REF_001 | Reactor.Reference | Warning | ReferenceCurrentReadAnalyzer - Use descriptor.Reference/binding.Reference instead of assigning ElementRef.Current to reference properties
 REACTOR_DSL_001 | Reactor.Dsl | Warning | MissingWithKeyAnalyzer - Dynamic list item missing .WithKey
 REACTOR_DOCK_001 | Reactor.Docking | Warning | OnLiveLayoutRoundTripAnalyzer - OnLiveLayoutChanged feeds the live layout back into state
+REACTOR_EVENT_001 | Reactor.Events | Warning | SetEventSubscriptionAnalyzer - .Set subscribes to an event that already has a declarative On* modifier
```
azchohfi, line 21: You register REACTOR_EVENT_001 here, but the analyzer reference table in `docs/_pipeline/templates/analyzer-architecture.md.dt` (around lines 106-126, where the sibling REACTOR_POOL_001 row lives) still doesn't list it. Can you add a row to that `.dt` template and run `mur docs compile` so the generated `docs/guide/analyzer-architecture.md` picks it up? The `reactor-build-and-check` cheat table only covers HOOKS/DSL/THEME/A11Y, so it doesn't need an entry.

src/Reactor.Analyzers/SetEventSubscriptionAnalyzer.cs
```diff
@@ -0,0 +1,197 @@
...
+            { "PreviewKeyDown", "OnPreviewKeyDown" },
+            { "PreviewKeyUp", "OnPreviewKeyUp" },
+            { "CharacterReceived", "OnCharacterReceived" },
+            { "AccessKeyDisplayRequested", "OnAccessKeyDisplayRequested" },
+            { "GotFocus", "OnGotFocus" },
+            { "LostFocus", "OnLostFocus" },
+            { "DragEnter", "OnDragEnter" },
+            { "DragOver", "OnDragOver" },
+            { "DragLeave", "OnDragLeave" },
+            { "Drop", "OnDrop" },
```
azchohfi: These four drag events are a different story from the routed events above. `.OnDragEnter/Over/Leave/Drop` take a Reactor `Action<DragTargetArgs>` and configure a `DropTargetConfig` (ElementExtensions.cs:494-529) — they're not a 1:1 replacement for a raw `DragEventHandler` subscription. So the warning fires and the message says "use `.OnDragOver(...)`", but there's no codefix and following the suggestion manually with the same handler won't compile and changes the semantics. I'd either drop these four from `EventModifiers`, or give them a tailored message that makes clear it's a drop-target API, not a straight event swap.

src/Reactor.Analyzers/SetEventSubscriptionAnalyzer.cs
```diff
@@ -0,0 +1,197 @@
...
+            { "RightTapped", "OnRightTapped" },
+            { "Holding", "OnHolding" },
+            { "KeyUp", "OnKeyUp" },
+            { "PreviewKeyDown", "OnPreviewKeyDown" },
+            { "PreviewKeyUp", "OnPreviewKeyUp" },
+            { "CharacterReceived", "OnCharacterReceived" },
+            { "AccessKeyDisplayRequested", "OnAccessKeyDisplayRequested" },
+            { "GotFocus", "OnGotFocus" },
+            { "LostFocus", "OnLostFocus" },
+        };
```
azchohfi: This is a hand-copy of `EventModifiers` minus the four drag entries, which is easy to let drift. Could we derive it instead — keep the single `EventModifiers` map plus a small `NonCodeFixableEvents` set for the drag exceptions, and have the codefix treat an event as fixable unless it's in that set? Then adding a new event only touches one place, and the exclusion is stated explicitly.

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionAnalyzerTests.cs
```diff
@@ -0,0 +1,301 @@
...
+    public static FakeElement OnCharacterReceived(this FakeElement el, Action handler) => el;
+    public static FakeElement OnAccessKeyDisplayRequested(this FakeElement el, Action handler) => el;
+    public static FakeElement OnGotFocus(this FakeElement el, Action handler) => el;
+    public static FakeElement OnLostFocus(this FakeElement el, Action handler) => el;
+    public static FakeElement OnDragEnter(this FakeElement el, Action handler) => el;
+    public static FakeElement OnDragOver(this FakeElement el, Action handler) => el;
+    public static FakeElement OnDragLeave(this FakeElement el, Action handler) => el;
+    public static FakeElement OnDrop(this FakeElement el, Action handler) => el;
+    public static FakeElement OnSizeChanged(this FakeElement el, Action handler) => el;
+}
```
azchohfi, line 90: These stubs make every modifier a zero-arg `OnXxx(this FakeElement, Action)`, so the codefix tests never check that the rewritten `.OnXxx(handler)` compiles against the real signatures (`Action<object, TArgs>`, and `Action<UIElement, ...>` for a couple). It lines up today, but one codefix test against the real Reactor types (or realistic typed stubs) would catch a future signature change that silently breaks the generated call.

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionAnalyzerTests.cs
```diff
@@ -0,0 +1,301 @@
...
+        var el = new FakeElement();
+        el.Set(fe => child.Tapped += OnTapped);
+    }
+}";
+
+        await new CSharpAnalyzerTest<SetEventSubscriptionAnalyzer, DefaultVerifier>
+        {
+            TestCode = source,
+        }.RunAsync(TestContext.Current.CancellationToken);
+    }
```
azchohfi, line 220: Two intended non-targets aren't pinned by a test: `-=` (the analyzer filters to AddAssignment) and multi-statement block lambdas (only single-statement blocks match). A couple of no-diagnostic tests — `fe.Tapped -= handler` and `fe => { Log(); fe.Tapped += handler; }` — would lock that behavior so a later refactor can't quietly start flagging them.

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionAnalyzerTests.cs
```diff
@@ -0,0 +1,301 @@
...
+        var el = new FakeElement();
+        {|REACTOR_EVENT_001:el.Set(fe => fe.DragOver += OnDragOver)|};
+    }
+}";
+
+        await new CSharpCodeFixTest<SetEventSubscriptionAnalyzer, SetEventSubscriptionCodeFix, DefaultVerifier>
+        {
+            TestCode = code,
+            FixedCode = code,
+        }.RunAsync(TestContext.Current.CancellationToken);
```
azchohfi, line 388: Only DragOver has the "fires but no codefix" test — DragEnter, DragLeave, and Drop are also non-codefixable but uncovered. Worth parameterizing this over all four so each keeps its warning-without-fix contract if the map ever changes.

azchohfi (commented):
Took a pass through the new `.Set(...)` event-subscription analyzer — nice work, it mirrors the REACTOR_POOL_001 pattern cleanly and the tests are thorough. No blockers from me. Two medium things worth a look (the drag-event handling and the missing analyzer-docs-table row) plus some low-severity test-coverage polish. Details inline.

nsreehari:
Pushed a follow-up pass addressing the maintainer comments on this PR:

- removed the drag/drop target events from `REACTOR_EVENT_001` so the rule no longer suggests `.OnDragEnter/Over/Leave/Drop` as if they were 1:1 raw-event replacements
- collapsed the duplicated event maps down to a single tracked map, with the code fix following that same source of truth
- added the missing analyzer table row in `docs/_pipeline/templates/analyzer-architecture.md.dt`
- strengthened the tests to pin intended non-targets (`-=`, multi-statement block lambdas, drag target events)
- fixed the misnamed block-body code-fix test and added a more realistic typed-signature code-fix test
- reworked the consistency test to use syntax inspection rather than string matching

Validation rerun after the patch:

- `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter "SetEventSubscriptionAnalyzerTests|SetEventSubscriptionConsistencyTests"`

I also ran `mur docs compile` to verify the analyzer docs template change; it completed successfully, but I did not include the unrelated generated image churn in this PR update.

src/Reactor.Analyzers/AnalyzerReleases.Unshipped.md
azchohfi: Confirmed in ccf127f1 - the `REACTOR_EVENT_001` row is in the `.dt` template now. Thanks!

src/Reactor.Analyzers/SetEventSubscriptionAnalyzer.cs
azchohfi: Resolved in ccf127f1 - the four drag events were dropped from the map entirely, so the rule no longer fires on them. Clean fix.

src/Reactor.Analyzers/SetEventSubscriptionAnalyzer.cs
azchohfi: Resolved in ccf127f1 - single `EventModifiers` map now, and the code fix reads from it. Looks good.

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionAnalyzerTests.cs
azchohfi: Resolved in ccf127f1 - the new `CodeFix_Rewrites_With_Realistic_Typed_Handler_Signature` test exercises the real `Action<object, TArgs>` shape. Exactly what I was after.

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionAnalyzerTests.cs
azchohfi: Resolved in ccf127f1 - `No_Diagnostic_For_Remove_Assignment` and `No_Diagnostic_For_Multi_Statement_Block_Lambda` pin both non-targets. Thanks.

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionAnalyzerTests.cs
azchohfi: Resolved in ccf127f1 - now covered by `No_Diagnostic_For_Drag_Target_Events` (analyzer-only), which reads cleaner than the old suppressed-code-fix assertion.

nsreehari:
Pushed one more small follow-up on this PR for the new Copilot review note:

- cached the parsed safe-modifier set in `SetEventSubscriptionConsistencyTests` so `ElementExtensions.cs` is read and parsed once per test process instead of once per test method

Validation rerun:

- `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter "SetEventSubscriptionAnalyzerTests|SetEventSubscriptionConsistencyTests"`

tests/Reactor.Tests/AnalyzerTests/SetEventSubscriptionConsistencyTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            .DescendantNodes()
+            .OfType<MethodDeclarationSyntax>()
+            .Where(method => method.Identifier.Text.StartsWith("On", StringComparison.Ordinal))
+            .Where(method => method.ReturnType is IdentifierNameSyntax { Identifier.Text: "T" })
+            .Where(method => MethodBodyContainsSameNamedModifierAssignment(method))
+            .Select(method => method.Identifier.Text)
+            .ToHashSet(StringComparer.Ordinal);
+
+        return modifiers;
+    }
```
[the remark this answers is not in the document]
azchohfi: Addressed in 20be2931 - `ReadSafeEventModifiers` now returns a `static Lazy<HashSet<string>>` (`ReadSafeEventModifiersCore`), so `ElementExtensions.cs` is read and parsed once per test process. None of the three tests mutate the shared set, so caching it is safe. Thanks!

azchohfi (approved):

azchohfi:
Thanks! Looking good!

Issue #675, opened 2026-06-25
perf/correctness: holistic audit of ApplyResourceOverrides + dirty-ancestor side-effects across the skip fast-path (positional + keyed) and the oldOverrides==null removal gate


azchohfi:
## Summary

Consolidated follow-up tracking **three related, pre-existing** items on the reconcile **skip fast-path** `ApplyResourceOverrides` surface, surfaced during the review passes for the COM-interop perf PR (#654 / #666) and the keyed-list-diff PR (#653 / #657): (a) a stale-key removal-gate gap, (b) a child-skip side-effect bypass, and (c) a missing behavior test for the shallow-skip path. (a) and (b) are latent correctness questions about *what side-effects the skip path is allowed to drop*; (c) is the coverage that would catch a regression in either. All three share a root area (`ApplyResourceOverrides` + dirty-ancestor handling on skip) and should be evaluated together across **all** skip paths.

> **All three are PRE-EXISTING on `origin/main` and were NOT introduced by #666 or #657/#30.** None should be "fixed" inside a behavior-preserving perf PR — hence this dedicated issue.

---

## Finding A — `ApplyResourceOverrides` skips stale-key removal when `oldOverrides == null`

`Reconciler.ApplyResourceOverrides` (`src/Reactor/Core/Reconciler.cs`, ~5016–5073). The managed-key **removal** block is gated:

```csharp
if (oldOverrides is not null && managed.Count > 0)   // ~5032
{
    // remove keys in `managed` no longer present in newOverrides
}
```

When `oldOverrides == null` **but** `managed.Count > 0` (i.e. this control already had Reactor-managed override keys applied via some prior path), the stale-key removal is skipped, so a previously-applied override key can persist in `fe.Resources` even though the current overrides no longer contain it.

**Pre-existing:** `origin/main` gates identically on `oldOverrides is not null`. The `&& managed.Count > 0` clause added in #26 is purely a no-op fast-path (`managed.Count == 0` ⇒ nothing to remove), so #666 neither introduces nor widens the gap. The open question is whether removal should instead be driven solely by `managed` vs `newOverrides` (as the apply path already is), independent of `oldOverrides` nullness, so stale keys that arrived via another path (e.g. pooling/reuse, or a skip-path interaction) are always reconciled away.

---

## Finding B — child-skip fast-path drops ThemeRef re-resolution + dirty-ancestor propagation (positional **and** keyed)

[...]

Issue #654, opened 2026-06-25
perf: coalesce COM-interop reads + allocations in reconcile path


azchohfi:
## Background

A perf scan measured a large C# vs Rust gap on the data-grid stress workload (`tests/stress_perf/StressPerf.ReactorOptimized`, a StocksGrid). The dominant gap is **Reconcile** (~46ms C# vs ~7.9ms Rust, ~5.8x), driven largely by the diff/patch path repeatedly crossing the COM/RCW boundary and allocating per element.

Each WinRT control access — `GetValue` on attached DPs, RCW type-tests (`is FrameworkElement`), `AutomationProperties` reads — is expensive, and the reconciler currently repeats several of these per element, per frame. This issue tracks coalescing those redundant reads and removing per-element allocations on the reconcile/patch path.

**Scope (files):** `src/Reactor/Core/Reconciler.cs`, `Reconciler.Update.cs`, `Reconciler.Mount.cs`, `Reconciler.KeyedItemsBinding.cs`, `AccessibilityScanner.cs` (plus minimal `Context`/`ContextScope`/`ComponentNode` plumbing the fixes require). Behavior must be preserved exactly — coalesced COM reads must observe the same values at the same times; unmount/cleanup ordering is unchanged; the accessibility scanner must emit identical findings.

## Flagship findings (attack the Reconcile gap directly)

- **#14** `Reconciler.Update.cs` — five separate `is FrameworkElement` type-tests on the same `target` per updated element. Coalesce to one `var targetFe = target as FrameworkElement;` and reuse.
- **#15** `Reconciler.cs CreateComponentRerender` — allocates a new `Action` closure every `ReconcileComponent` (per stateful row/cell per frame). Cache the wrapped delegate on the `ComponentNode`; recreate only when the `requestRerender` identity changes.
- **#16** `Reconciler.cs UnmountRecursive` — reads `ReactorAttached.StateProperty` (via `GetElementTag`) 5× on the same control per removed cell. Read once at the top, reuse.
- **#17** `Reconciler.cs UnmountAndCollect` — 5 independent `GetElementTag` COM calls on the same control. Read once, reuse.

## Other reconcile/patch findings

- **#18** `UpdateDefaultAutomationName` — `AutomationProperties.GetName` COM read for every captioned element past `ShallowEquals`. Skip the redundant `SetName` write when the value is unchanged (provenance prevents safely skipping the read).
[...]

PR #769, opened 2026-06-30
perf: re-validate #654 COM-interop coalescing on current main (double-check #666 no-op)

azchohfi:
## Why this PR exists

PR #666 implemented issue #654 ("coalesce COM-interop reads + allocations in the
reconcile path") and was **closed as an "honest no-op"** — the optimizations are
real and behavior-preserving, but their effect measured below `/perf` resolution
on the available workloads (no win, no regression), and the earlier skip-floor
`rps` regression was judged a macro-rps false positive.

This PR **re-applies #666's changes onto current `main`** purely to
**independently re-measure** via the authoritative CI `/perf` workflow and
confirm (or refute) that no-op verdict on today's baseline. **It is a measurement
vehicle and is not intended to merge as-is.**

## What's here

- A squash re-apply of `azchohfi-reconciler-interop-perf` (PR #666) onto current `main`.
- The only merge conflict was in `Reconciler.cs` `ApplyResourceOverrides`, resolved
  in favor of **main's #675 removal-guard fix** (drop the `oldOverrides is not null`
  gate, which leaked stale managed keys for pooled/Mount-path controls) while keeping
  the perf-#26 deferred-removal loop body. The other 7 files auto-merged.

## Local validation (ARM64 dev box, x64 build)

- Core `Microsoft.UI.Reactor.dll` compiles clean (0 C# errors).
- `dotnet test tests/Reactor.Tests`: **9999 passed, 0 failed** (incl. the 21 new
  `ReconcilerInteropPerfTests` — rerender-cache, `CurrentValueEquals`, scanner parity).
- `tests/Reactor.SelfTests` (live WinUI): **636 passed**; the lone batch failure was a
  shared-300s-budget timeout on `AsyncResource.Framerate.DataGridScroll`, which **passes
  in isolation** — an environmental timing flake, not a reconcile regression.

## Result — the no-op holds ✅

Ran CI `/perf` **twice** (interleaved paired-Δ, 12 reps each). Full write-up + two-run
table in [this comment](https://github.com/microsoft/microsoft-ui-reactor/pull/769#issuecomment-4849164833).

- **The "honest no-op" verdict holds** to harness resolution — reconcile/diff/throughput
  within noise on every macro workload, across both runs.
- The feared skip-floor `rps −6.6%` regression is a **confirmed false positive** (−0.4%, then 0.0%).
- The one real, **reproducible** effect is the small **keyed-list allocation win**
  (−2.4% bytes/render, tight CI) the author already acknowledged.
- A newer **DataGrid** macro workload showed a run-1 win (rps +3.9%, reconcile/diff −5.2%,
  all CIs excluding zero) that **did not reproduce** in the confirmatory run (every CI
  crossed back over zero). Directionally favorable, but below reproducible-significance —
  the confirmatory run avoided reporting a run-1 false positive.

**Recommendation: keep #666 closed.** The evidence doesn't support overturning the no-op.

Closes nothing. Refs #654, #666.

### DIFF
--- a/src/Reactor/Core/AccessibilityScanner.cs
+++ b/src/Reactor/Core/AccessibilityScanner.cs
@@ -117,6 +117,12 @@ public static partial class AccessibilityScanner
     // implementation and its dependency chain.
     private static IScanExtension? s_scanExtension;
 
+    // Reused per-thread scratch list for BuildContext's child extraction so a
+    // per-finding Where(...).ToList() is not allocated each call (perf #64).
+    // Cleared before and after use so it never pins Element references.
+    [ThreadStatic]
+    private static List<Element>? t_buildCtxChildren;
+
     /// <summary>
     /// Registration hook for a subsystem-specific scanner extension (e.g. the
     /// Charting accessibility checker). Installed once when the subsystem
@@ -175,7 +181,13 @@ public static partial class AccessibilityScanner
     {
         if (el is null or EmptyElement) return;
 
-        ctx.Push(el);
+        // Compute children once and reuse for both sibling-text collection
+        // (inside Push) and the recursion below. GetChildren does a full
+        // switch dispatch and (for the single-child arms) allocates a fresh
+        // array literal on every call, so calling it twice per element
+        // doubles that cost (perf #62).
+        var children = GetChildren(el);
+        ctx.Push(el, children);
 
         // Per-element checks
         CheckIconButton(el, ctx, findings);
@@ -196,7 +208,7 @@ public static partial class AccessibilityScanner
         CollectLandmark(el, ctx);
 
         // Recurse into children
-        foreach (var child in GetChildren(el))
+        foreach (var child in children)
             Walk(child, ctx, findings);
 
         ctx.Pop();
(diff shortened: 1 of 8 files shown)


azchohfi:
## Re-validation of #666 (the "honest no-op") — verdict: **the no-op holds**

I re-applied #666's changes onto current `main` (one conflict in `Reconciler.cs` `ApplyResourceOverrides`, resolved in favor of `main`'s #675 fix), validated correctness (0 compiler errors; 9,999 unit tests + the 21 new `ReconcilerInteropPerfTests` pass; 636 live-WinUI selftests pass), and ran the authoritative CI `/perf` **twice** (12 interleaved paired reps each).

The second run was deliberate: run 1 *appeared* to show a clear DataGrid win that contradicted the no-op call, and the DataGrid re-render is async / dispatcher-debounced (the harness itself flags those timing rows as "noisier"). A finding that contradicts the prior verdict needs to reproduce.

### Two-run comparison

| Workload / metric | Run 1 (`28480908384`) | Run 2 (`28483199746`) | Reproducibly significant? |
|---|---|---|---|
| StocksGrid headline — Reconcile / Diff | −1.1% / −1.6% ≈ | +0.8% / −0.5% ≈ | no-op (both) |
| Skip-floor `--percent 0` — rps (the feared −6.6%) | −0.4% ≈ | 0.0% ≈ | **no regression** (both) |
| **Keyed-list — alloc bytes/render** | **−2.3% [−2.5,−2.1] ✅** | **−2.4% [−2.6,−2.1] ✅** | **Yes — robust** |
| DataGrid — Renders/sec | +3.9% [+2.4,+5.5] ✅ | +2.3% [−0.5,+5.0] ≈ | **No — flips to noise** |
| DataGrid — Reconcile / Diff | −5.2% [−6.8,−3.7] ✅ | −3.4% [−7.3,+0.5] ≈ | **No** |
| DataGrid — Gen0 GC / 1k | −5.2% [−9.7,−0.7] ✅ | −1.8% [−6.5,+2.9] ≈ | **No** |
| DataGrid — alloc bytes/render | ~0% | −0.1% [−0.4,+0.2] | flat (both) |
| Flex/Yoga, micro-benchmarks M1–M14 | ≈ | ≈ | no-op (both) |

**Every DataGrid CI that excluded zero in run 1 crosses zero in run 2.** The point estimates stay directionally favorable (they never favor `main`), and DataGrid `alloc bytes/render` is flat in both runs — so there's a weak *directional* hint, but nothing that clears a reproducible-significance bar on the current harness.

### Verdict

- **The author's "honest no-op" call holds** to the resolution of the current harness. Reconcile/diff/throughput are within noise across both runs on every macro workload.
- The feared skip-floor `rps −6.6%` regression is a **confirmed false positive** (−0.4%, then 0.0%).
- The **one real, reproducible effect is the small keyed-list allocation win** (−2.4% bytes/render, tight CI, reproduced near-identically) — which the author had already acknowledged.
- The DataGrid workload's run-1 "win" was a **run-specific fluctuation near the noise floor** that did not survive the confirmatory run. Fittingly, the confirmatory pass prevented reporting a run-1 *false positive* — the same class of error (a phantom perf delta) that #654/#666 was about in the first place.

**Recommendation: keep #666 closed.** The evidence does not support overturning the no-op. If DataGrid delegate/alloc coalescing is worth revisiting later, the harness's DataGrid dispatcher-debounce noise should be reduced first so a ~2–4% effect can actually resolve run-over-run.

<sub>This draft PR (#769) was a measurement vehicle only — not intended to merge. Full per-run tables are in the sticky `/perf` comment above.</sub>

azchohfi:
Closing — measurement complete. Two interleaved CI \/perf\ runs confirm the #666 'honest no-op' verdict holds (see the verdict comment above). Branch \zchohfi-perf-coalesce-com-interop-reads\ is kept for reference; not deleting it.

Issue #761, opened 2026-06-30
Add a first-class external-store hook for subscribe/getSnapshot patterns


nsreehari:
## Summary

Reactor has strong primitives for local state and side effects, but it is missing a first-class hook for external event-backed stores.

When a store exposes a subscription mechanism and a current snapshot getter, consumers currently need to hand-roll the same pattern every time:

- subscribe in `UseEffect`
- trigger a rerender with `UseState` or `UseReducer`
- read the latest snapshot during render

That works, but it is boilerplate, easy to get subtly wrong, and there is no standard Reactor abstraction for it.

## Why the current hook surface is not enough

`UseObservable` covers `INotifyPropertyChanged`, and `UseEffect` plus `UseReducer` can bridge arbitrary stores, but that still leaves every app to reimplement the same subscribe/getSnapshot pattern.

This feels like a reusable abstraction that belongs at the framework level rather than repeated in every app.

## Example

A typical external store might look like this:

```csharp
public sealed class SomeStore
{
    public event Action? Changed;
    public AppState Snapshot => _state;
}
```

A Reactor consumer then ends up writing a custom bridge like this:

```csharp
TSnapshot UseStoreSubscription<TSnapshot>(
    Func<Action, Action> subscribe,
    Func<TSnapshot> getSnapshot)
{
    var (_, setRevision) = UseState(0);

    UseEffect(() =>
    {
        void OnChanged() => setRevision(Environment.TickCount);
        var unsubscribe = subscribe(OnChanged);
        return unsubscribe;
    });

    return getSnapshot();
}
```

That pattern is small, but it is the same small piece of infrastructure every app has to invent for itself.

## Proposed API

```csharp
TSnapshot UseExternalStore<TSnapshot>(
    Func<Action, Action> subscribe,
    Func<TSnapshot> getSnapshot,
    IEqualityComparer<TSnapshot>? comparer = null)
```

## Expected behavior

- read the current snapshot during render
- subscribe once for the hook slot
- re-read the snapshot on change notifications
- rerender only when the snapshot value changes, using the provided comparer or `EqualityComparer<TSnapshot>.Default`
- dispose the subscription on unmount

## Benefits

- a canonical pattern for arbitrary event-backed stores
- less repeated app-level boilerplate
- fewer cleanup and stale-subscription mistakes
- a clearer alternative to forcing rerenders manually
[...]

PR #772, opened 2026-07-01
Add UseExternalStore hook

nsreehari:
## Summary
- add `UseExternalStore<TSnapshot>` to `RenderContext`, with a matching `Component` convenience wrapper
- model the hook around the external-store pattern discussed in #761: read the current snapshot during render, own the subscription in an effect, and rerender only when the comparer reports a real snapshot change
- document the new hook in the public hooks guide and the hook internals guide
- add focused observable-hook tests covering subscription lifetime, equality suppression, custom comparer behavior, and resubscription when the store source changes

## Why this belongs in Reactor
Apps can build this pattern today from `UseEffect` plus `UseState` or `UseReducer`, but that repeats the same subscription bridge in every consumer. This hook makes that integration explicit and reusable at the framework layer for event-backed stores that need to feed render-time snapshots into Reactor components.

## Validation
- `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter ObservableHookTests`


nsreehari:
A concrete use case from my side is `demo-boards-winui-app`, where some app-level state sources behave like external stores: the component needs to subscribe to change notifications while also reading the latest snapshot during render.

That works today with a small custom hook built from `UseEffect` plus `UseState`/`UseReducer`, but it is the same subscription bridge logic each time rather than app-specific behavior. This PR is meant to make that integration a first-class Reactor pattern instead of pushing every consumer to rebuild the same adapter.

nsreehari:
Pushed a follow-up pass on this PR to address a likely correctness concern before review:

- closed the render-to-subscribe race in `UseExternalStore` by re-reading the snapshot immediately after the subscription is installed
- added a focused regression test that simulates a store change landing in that gap and verifies the next render sees the updated snapshot
- filled in the missing `UseExternalStore` hook reference surface in the docs

Validation rerun:

- `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter ObservableHookTests`

azchohfi:
Please, merge main, as there are some conflicts.

nsreehari:
Merged upstream/main into this branch to resolve the reported conflicts. I reran a focused validation slice locally before pushing: dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter ObservableHookTests\n\nGitHub has started a fresh CI run for the merge commit.

nsreehari:
Follow-up pushed in `460cdb1b` to address the remaining review notes:

- changed `UseExternalStore` to lazily initialize its `ExternalStoreState<TSnapshot>` ref, so we no longer allocate a fresh state object on every render
- added the missing `## External Stores` narrative to `docs/_pipeline/templates/hooks.md.dt` so the template matches the generated hooks guide

Validation rerun:

- `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter ObservableHookTests`

Also, the previously pending CI on this PR has now settled green.

src/Reactor/Core/RenderContext.cs
```diff
@@ -927,6 +927,59 @@ void handler(object? s, global::System.ComponentModel.PropertyChangedEventArgs e
         return source;
     }
 
+    /// <summary>
+    /// Subscribes to an external store that exposes a current snapshot getter.
+    /// Re-renders only when a change notification yields a different snapshot.
+    /// </summary>
+    public TSnapshot UseExternalStore<TSnapshot>(
```
azchohfi, line 953: `subscribe` is the effect's dependency (see the `}, subscribe);` on line 978), so its delegate identity is what decides whether we re-subscribe. A method group like `_store.Subscribe` stays stable across renders, but a capturing lambda (`onChanged => _store.Subscribe(onChanged)`) is a fresh delegate every render and will unsubscribe/resubscribe each time. That's an easy footgun and it isn't documented anywhere — can we call it out in the XML `<remarks>` and the guide so authors know to pass a stable delegate? It's the same guidance React gives for `useSyncExternalStore`.

src/Reactor/Core/RenderContext.cs
```diff
@@ -927,6 +927,59 @@ void handler(object? s, global::System.ComponentModel.PropertyChangedEventArgs e
...
+                if (changed)
+                    forceRender(revision => revision + 1);
+            }
+
+            var unsubscribe = subscribe(OnChanged);
+
+            // Close the render-to-subscribe race: if the store changed after the
+            // render-time snapshot read but before the subscription became active,
+            // re-read once immediately and schedule a rerender if needed.
+            OnChanged();
```
azchohfi, line 997: Since the effect re-runs whenever `subscribe` changes and immediately calls `OnChanged()` here, an unstable `subscribe` combined with a `getSnapshot` that returns a fresh, never-equal value each call (e.g. `() => items.ToArray()`) will spin: re-render → effect re-runs → this `OnChanged()` sees a "change" → `forceRender` → repeat. React guards this by throwing when `getSnapshot` isn't cached. At a minimum let's document that `getSnapshot` must return a cached/stable value; a dev-only warning when this immediate re-check trips on a freshly-installed subscription would be even better.

src/Reactor/Core/RenderContext.cs
```diff
@@ -927,6 +927,59 @@ void handler(object? s, global::System.ComponentModel.PropertyChangedEventArgs e
...
+    /// <summary>
+    /// Subscribes to an external store that exposes a current snapshot getter.
+    /// Re-renders only when a change notification yields a different snapshot.
+    /// </summary>
+    public TSnapshot UseExternalStore<TSnapshot>(
+        Func<Action, Action> subscribe,
+        Func<TSnapshot> getSnapshot,
+        IEqualityComparer<TSnapshot>? comparer = null)
+    {
+        var snapshot = getSnapshot();
```
azchohfi, line 961: Minor: null `subscribe`/`getSnapshot` currently blow up as an incidental `NullReferenceException` here (and later at `subscribe(OnChanged)`). A couple of `ArgumentNullException.ThrowIfNull(...)` at the top would give a clearer contract. The neighboring observable hooks don't validate their args either, so I'll leave it to you whether it's worth the consistency break.

tests/Reactor.Tests/ObservableHookTests.cs
```diff
@@ -47,6 +81,165 @@ public void RaiseAllPropertiesChanged()
...
+
+        ctx.BeginRender(() => rerenderCount++);
+        ctx.UseExternalStore(store.Subscribe, () => store.Snapshot, StringComparer.OrdinalIgnoreCase);
+        ctx.FlushEffects();
+
+        ctx.BeginRender(() => rerenderCount++);
+        ctx.UseExternalStore(store.Subscribe, () => store.Snapshot, StringComparer.OrdinalIgnoreCase);
+        ctx.FlushEffects();
+
+        store.SetSnapshot("alice");
```
azchohfi, line 186: Nice coverage overall — a few gaps worth closing while this is fresh:

- The custom-comparer test only checks the *equal* (no re-render) case. Add one where the comparer reports a difference (e.g. `Alice`→`Bob`) and assert exactly one re-render.
- The off-thread path is the whole reason for `threadSafe: true` + the `Gate` lock, but every test fires `Changed` synchronously on the test thread. A test that raises the change from `Task.Run` (await it, then assert one re-render / no exception) would actually exercise it.
- Nothing asserts that a stable method-group `subscribe` does *not* re-subscribe across renders — counting subscribe/unsubscribe calls in the store would lock that contract in.

docs/guide/reference/hooks/UseExternalStore.md
```diff
@@ -0,0 +1,15 @@
...
+> **Learn more:** [Hooks](../../hooks.md), [Effects](../../effects.md)
+
+## Summary
+
+Subscribes to an external store that exposes a current snapshot getter.
+Re-renders only when a change notification yields a different snapshot.
+
+## Featured in
+
+- [Hooks](../../hooks.md)
```
azchohfi: Missing trailing newline here — the sibling stubs (e.g. `UseObservable.md`) end with one. Also worth confirming whether these reference pages are produced by the docs pipeline; if they are, this should come from the generator rather than being hand-added.

skills/reactor.api.txt
```diff
@@ -889,6 +889,7 @@ RenderContext.UseEffect<T1>(Action effect, T1 d1) → void
 RenderContext.UseEffect<T1>(Func<Action> effectWithCleanup, T1 d1) → void
 RenderContext.UseElementFocus(FocusState state = FocusState.Programmatic) → ValueTuple<ElementRef, Action>
 RenderContext.UseElementRef<T>() → ElementRef<T>
+RenderContext.UseExternalStore<TSnapshot>(Func<Action, Action> subscribe, Func<TSnapshot> getSnapshot, IEqualityComparer<TSnapshot> comparer = null) → TSnapshot
```
azchohfi, line 901: This index is generated (`mur --regen-api`). The added line is correct and matches the plugin copy, so no harm here — but regenerating from source rather than hand-editing keeps it authoritative and avoids future drift.

azchohfi (changes requested):

azchohfi:
A couple of follow-ups from the review that don't map to a specific diff line, so noting them here:

- **CHANGELOG** — this adds a new public hook (`UseExternalStore<TSnapshot>`), but there's no entry under `## [Unreleased] → ### Added`. Worth a bullet in the same style as the existing ones, with the issue #761 cross-reference.
- **`reactor-getting-started` skill** — the key-hooks table in `plugins/reactor/skills/reactor-getting-started/SKILL.md` lists `UseObservable` but not `UseExternalStore`. They're the same class of subscribe-based hook, so a matching row there would keep the shipped agent kit in sync.

nsreehari:
Thanks for the detailed review @azchohfi — pushed `ed4d53c5` addressing the CHANGES_REQUESTED points:

**Docs / contract**
- Documented the two stability rules in the XML `<remarks>` on `UseExternalStore` and in the hooks guide (+ its `hooks.md.dt` template, kept in sync): `subscribe` must be a stable delegate (method group / `UseCallback`), and `getSnapshot` must return a cached value — called out the resubscribe footgun and the spin scenario, referencing React's `useSyncExternalStore` guidance.
- Added `ArgumentNullException.ThrowIfNull` for `subscribe`/`getSnapshot` so a null contract fails clearly instead of an incidental NRE.
- **CHANGELOG:** added an `## [Unreleased] → ### Added` bullet for `UseExternalStore<TSnapshot>` with the issue #761 cross-reference.
- **reactor-getting-started skill:** added a `UseExternalStore` row to the key-hooks table next to `UseObservable`.
- **Reference page:** fixed the missing trailing newline on `UseExternalStore.md` to match the sibling stubs. These pages are pipeline-generated (Section 10 marker-driven) and the Docs build + tier-drift checks are green, so the content is consistent with the generator; only the trailing newline was off.

**Tests** (`ObservableHookTests`)
- Custom comparer that reports a *difference* (`Alice`→`Bob` under `StringComparer.Ordinal`) asserting exactly one re-render.
- Off-thread change raised from `Task.Run` (awaited) to actually exercise the `threadSafe: true` reducer + `Gate` lock.
- Stable method-group `subscribe` does **not** re-subscribe across renders (asserts `SubscribeCount == 1`, `UnsubscribeCount == 0`).
- Null-argument validation.

Re the per-render allocation thread: already handled earlier in `460cdb1b` via the lazy `UseRef<ExternalStoreState<...>?>` init.

Validation: `dotnet test tests/Reactor.Tests/Reactor.Tests.csproj -p:Platform=x64 -m:1 --filter "FullyQualifiedName~ObservableHookTests"` → Passed 35 / Failed 0.

docs/guide/hooks-internals.md
```diff
@@ -273,7 +273,7 @@ If you find yourself wanting `UseRef` to trigger re-renders, you
 probably want `UseState` instead. `UseRef`'s value is an escape hatch
 for state that shouldn't drive UI.
 
-## UseObservable, UseResource, UseContext
+## UseObservable, UseExternalStore, UseResource, UseContext
```
azchohfi, line 268: This heading change — and the `UseExternalStore` ownership paragraph just below (~lines 303-308) — currently live only in this **generated** file. The source template `docs/_pipeline/templates/hooks-internals.md.dt` still reads `## UseObservable, UseResource, UseContext` with no `UseExternalStore` anywhere. Since the guide is assembled from the template, the next full `mur docs compile` will drop both edits.

(CI is green because the docs check only recompiles the templates touched in this PR — `hooks.md.dt` — not `hooks-internals`, so the drift is latent.)

Fix: mirror these two edits into `hooks-internals.md.dt` — the heading, plus the ownership paragraph after the `UseObservable` one — exactly like you already did for `hooks.md.dt`.

azchohfi: Went ahead and pushed this fix directly to the branch in `a44ac854` — added the heading + ownership paragraph to `hooks-internals.md.dt` so it matches the generated `hooks-internals.md`. Verified a local `mur docs compile --topic hooks-internals` reproduces the section byte-for-byte, so a future compile won't drop it. (Left the generated file untouched since it already had the correct content.)

azchohfi (approved):

Issue #162, opened 2026-05-06
Charts: double-announcement risk with custom LabelView / TickLabelView


codemonkeychris:
## Background

PR adding `PieChartElement<T>.LabelView(Func<T, PieSliceLayout, Element>)` and `ChartElement<T>.{X,Y}TickLabelView(Func<double, Element>)` lets callers replace the built-in `TextBlock` at a slice centroid / axis tick with any Reactor `Element` (including a custom `Component`).

As a stopgap, the chart wraps the returned element with `.AccessibilityView(AccessibilityView.Raw)` and `.OnMount(fe => fe.IsHitTestVisible = false)`, so the visual replacement is hidden from the UIA tree and pointer/focus events. Slice/axis a11y data still flows through the chart's existing `IChartAccessibilityData` contract.

## The risk

`AccessibilityView.Raw` is set on the **outer wrapper** the chart applies. A user-supplied `Element` may be a structured composite (e.g. a Reactor `Component` containing focusable children, a `TextBlock`, an icon, etc.) whose interior peers may still surface to assistive tech depending on how the platform composes peer trees under a `Raw`-marked parent. Concretely:

1. **Double-announcement** — the chart's own slice/point descriptor (`ChartPointDescriptor`) names the slice (e.g. \"Slice 5: 12 units\"), and a user `LabelView` containing the same string can produce a second peer that announces it again.
2. **Focusable children** — if the user's `Element` has interactive sub-elements, `IsHitTestVisible=false` on the outer FE prevents pointer hits but does not necessarily remove the inner peers from keyboard/focus traversal.
3. **`DataLabel` not auto-fed** — when a caller uses `LabelView` and skips the string `LabelAccessor`/`DataLabel`, the slice's accessible name falls back to `\"Slice {i+1}\"`, even though the visual clearly shows a meaningful label. Screen-reader users get a worse experience than sighted users.

## Proposed direction (separate from this PR)

- Walk the wrapper's peer subtree and force `AccessibilityView.Raw` on every descendant FE the chart owns, so only the chart's own peer surfaces.
- Provide an opt-in `LabelView(..., interactive: true)` overload that documents the user is responsible for a11y when they reach for it.
- When a `LabelView` is provided **without** a `DataLabel`/`LabelAccessor`, allow callers to supply an `accessibleName` projection (e.g. `LabelView(render, name: item => item.ToString())`) so the chart's own descriptor stays accurate.
[...]

PR #810, opened 2026-07-02
Add declarative IsHitTestVisible(bool) common element modifier

### LINKED ISSUE #809: Add declarative IsHitTestVisible(bool) modifier on the common element surface
## Summary

Reactor exposes declarative common-element modifiers for the sibling boolean/visual state properties on the shared element surface — `IsEnabled(bool)`, `IsTabStop(bool)`, `IsVisible(bool)`, and `Opacity(double)` — but there is **no** declarative modifier for `UIElement.IsHitTestVisible`. The only in-tree affordance is chart-scoped (`TightHitTest` on the charting surface); everything else must reach for the imperative `.Set(c => c.IsHitTestVisible = false)` escape hatch.

`IsHitTestVisible` is a first-class, universally-applicable `UIElement` property with exactly the same shape as the modifiers that already exist (a plain `bool` on every `FrameworkElement`). Its absence is an inconsistency in the common-modifier surface rather than a deliberate omission.

## Motivation

`Reactor.Community.ReactorFlow`'s MiniMap renders pointer-transparent overlay layers (the viewport rectangle and node thumbnails) that must never intercept pointer input destined for the interactive canvas underneath. Today ReactorFlow carries a small local `NonInteractive` helper whose entire remaining body is:

```csharp
element.Set(c => c.IsHitTestVisible = false)
```

This is the last imperative `.Set(...)` escape hatch in that component. A declarative `IsHitTestVisible(false)` modifier would let it drop the helper entirely and compose the property like any other element state, matching the idiom already used for `IsEnabled` / `IsVisible`.

## Proposed API

```csharp
public static T IsHitTestVisible<T>(this T element, bool isHitTestVisible = true) where T : Element;
```

Placed next to `IsTabStop`/`IsVisible` on the common element surface (`ElementExtensions.cs`), with a default of `true` to mirror the WinUI default and the existing boolean modifiers.

## Notes

- **Additive / back-compat.** New optional modifier; no existing signature or behavior changes.
- **Plumbing mirrors `IsTabStop` exactly.** A nullable `bool? IsHitTestVisible` field on `ElementModifiers` (unset ⇒ untouched), folded into `ModifiersEqual` and `Merge` identically to `IsTabStop`, and applied in the reconciler as `fe.IsHitTestVisible = value` only when set and changed.
- **Recycle-safety already covered.** `ElementPool` already clears `UIElement.IsHitTestVisibleProperty` on release (issue #162), so a recycled control cannot strand a stale `false` from a prior use.

nsreehari:
## Summary

Adds a declarative `IsHitTestVisible(bool)` modifier to the common element surface, filling the last gap among the sibling boolean/visual state modifiers (`IsEnabled`, `IsTabStop`, `IsVisible`, `Opacity`). Until now, opting an element out of pointer hit-testing required the imperative `.Set(c => c.IsHitTestVisible = false)` escape hatch.

Fixes #809.

## Motivation

`Reactor.Community.ReactorFlow`'s MiniMap renders pointer-transparent overlay layers (viewport rectangle, node thumbnails) that must not intercept pointer input destined for the interactive canvas underneath. Its last remaining `.Set(...)` escape hatch is a `NonInteractive` helper whose entire body is `element.Set(c => c.IsHitTestVisible = false)`. A declarative modifier lets that helper be dropped and the property composed like any other element state.

## What changed

Plumbing mirrors `IsTabStop` exactly:

- `src/Reactor/Core/Element.cs` — added `bool? IsHitTestVisible` to `ElementModifiers`, folded into `ModifiersEqual` and `Merge`.
- `src/Reactor/Elements/ElementExtensions.cs` — added `public static T IsHitTestVisible<T>(this T el, bool isHitTestVisible = true) where T : Element` next to `IsTabStop`, with matching XML-doc style.
- `src/Reactor/Core/Reconciler.cs` — applies `fe.IsHitTestVisible = value` only when set and changed, next to where `IsTabStop` is applied.
- `tests/Reactor.Tests/ElementExtensionsCoverageTests.cs` — added `IsHitTestVisible_Sets_Modifier`, mirroring `IsEnabled_Sets_Modifier`.
- `skills/reactor.api.txt` + `plugins/reactor/skills/reactor-dsl/references/reactor.api.txt` — regenerated API index (`UPDATE_API_INDEX=1`).

Additive and back-compat: new optional modifier defaulting to `true` (the WinUI default); no existing signature or behavior changes. Recycle-safety is already covered — `ElementPool` clears `UIElement.IsHitTestVisibleProperty` on release (issue #162), so a recycled control cannot strand a stale `false`.

## Validation

Built and tested on win-x64:

```
UPDATE_API_INDEX=1 dotnet test tests/Reactor.Tests \
  --filter "FullyQualifiedName~Tooling.ApiIndexGeneratorTests.Index_IsUpToDate" -r win-x64
# → regenerated both reactor.api.txt copies; Passed 1

dotnet test tests/Reactor.Tests \
  --filter "FullyQualifiedName~ElementExtensionsCoverageTests|FullyQualifiedName~Tooling.ApiIndexGeneratorTests" \
  -r win-x64 --no-build
# → Passed 63, Failed 0 (incl. IsHitTestVisible_Sets_Modifier and Index_IsUpToDate)
```

Build succeeded with 0 errors.


src/Reactor/Core/Reconciler.cs
```diff
@@ -3832,6 +3832,9 @@ internal void ApplyModifiers(FrameworkElement fe, ElementModifiers? oldM, Elemen
         if (m.IsTabStop.HasValue && m.IsTabStop != oldM?.IsTabStop)
             fe.IsTabStop = m.IsTabStop.Value;
 
+        if (m.IsHitTestVisible.HasValue && m.IsHitTestVisible != oldM?.IsHitTestVisible)
+            fe.IsHitTestVisible = m.IsHitTestVisible.Value;
```
azchohfi, line 3848: This only writes when `HasValue`, so it never resets. If an element sets `.IsHitTestVisible(false)` and a later render drops the modifier, the update path hands `ApplyModifiers` an empty `ElementModifiers` (`Reconciler.Update.cs:239-240`) and this write is skipped — the control stays click-through. The `IsDragRegion`/`AutomationName` handlers above have a clear-on-unset branch; this one (like `IsTabStop`) doesn't. Pooled reuse is fine since `ElementPool` already clears the DP, but an in-place set→unset on a retained control isn't.

```suggestion
        if (m.IsHitTestVisible.HasValue && m.IsHitTestVisible != oldM?.IsHitTestVisible)
            fe.IsHitTestVisible = m.IsHitTestVisible.Value;
        else if (!m.IsHitTestVisible.HasValue && oldM?.IsHitTestVisible.HasValue == true)
            fe.ClearValue(UIElement.IsHitTestVisibleProperty);
```

Would also be worth a selftest fixture covering the set→unset round-trip against a live control — the current unit test only asserts the modifier record, not the reconciler write.

nsreehari: Addressed in `8c57e707`: removing the modifier now clears `UIElement.IsHitTestVisibleProperty`, and `ModifierEvent_ModifierClearResets` covers the live false-to-unset round trip. I also merged current `upstream/main` in `232309fb` so the coverage workflow has its required `tests/coverage/ci` scripts. Focused validation: `dotnet run --project tests/Reactor.AppTests.Host -p:Platform=x64 -- --self-test --filter "ModifierEvent_ModifierClearResets"` exited successfully. The focused `Reactor.Tests` run reached the Reactor build but is currently blocked locally by a TLS handshake downloading `@github/copilot-win32-x64`; CI will provide the clean-environment result. Please resolve this thread if the reset coverage addresses the finding.

azchohfi: Verified in `8c57e707` — the `ClearValue(UIElement.IsHitTestVisibleProperty)` branch is exactly right, and `ModifierClear_HitTestVisibleRestored` exercises the live false-to-unset round trip against a real control, which was the gap I was worried about. The local `Reactor.Tests` TLS/download flake is environmental, not your change — CI will cover it. Resolving. Thanks for the quick turnaround!

PR #812, opened 2026-07-02
fix(reconciler): re-render context consumers behind reference-stable child skips (#811)

### LINKED ISSUE #811: Bug: pure context changes can be lost behind reference-stable child skip paths
## Summary

A pure context change can be lost when a reference-stable child subtree is skipped by `ChildReconciler` before reconciliation descends into the context consumer.

## Concrete repro

I hit this while adding Tier-2 UI coverage for ReactorFlow:

- ReactorFlow repo: https://github.com/nsreehari/ReactorFlow
- Repro surface: the controls overlay reads `ctx.Interactive` from context to label the lock button as `Lock interactivity` or `Unlock interactivity`.
- On click, the underlying surface state does change: panning freezes as expected.
- But the overlay never relabels, and a second click cannot unlock.

Observed behavior:

- First click engages the lock.
- The button keeps the stale `Lock interactivity` label instead of switching to `Unlock interactivity`.
- A second click dispatches the stale action shape and re-locks instead of unlocking.

Expected behavior:

- A context-consuming descendant should re-render when its consumed context changes, even if an ancestor re-emits that descendant as the same element instance.

## Why this appears to be Reactor core

The relevant flow appears to be:

- `ChildReconciler.UpdateCommonChild(...)` can early-return on `Element.CanSkipUpdate(oldEl, newEl)`.
- `Element.CanSkipUpdate(...)` delegates to `ShallowEquals(...)`.
- `Element.ShallowEquals(...)` returns `true` immediately on `ReferenceEquals(a, b)`.
- `Reconciler.HasConsumedContextChanged(...)` exists, but it is only consulted once reconciliation has already descended into a component or memo node.

That means a reference-stable child subtree can be skipped before the context-consumer invalidation check is ever reached.

## Minimal conceptual repro

1. A parent component provides context and renders a caller-supplied child/overlay.
2. The caller creates that child element once and reuses the same element instance across renders.
3. Inside that subtree, a component or memo consumer reads the provided context and derives visible UI from it.
4. The parent updates only the provided context value.
5. The parent re-renders, but the child element remains reference-stable.
6. Child reconciliation skips the subtree, so the context consumer never re-renders.

Actual result:

- The consumer UI remains stale.

Expected result:

- The consumer UI updates because consumed context changed.

## Notes

This seems adjacent to the dirty-ancestor / skip-path correctness class, but it is specifically about pure context invalidation being lost behind child-skip fast paths.

I am not prescribing the fix, but the skip-path contract likely needs to preserve context-consumer invalidation semantics for reference-stable descendants instead of treating structural equality alone as sufficient.

nsreehari:
## Bug: pure context changes lost behind reference-stable child skip paths

Fixes #811.

### Root cause
A `UseContext` consumer nested inside a **reference-stable** subtree stopped
re-rendering on pure context changes. When a parent re-rendered with a changed
provider value but returned the *same* child `Element` instance, the
reconciler's shallow-equality fast paths short-circuited **before** descending
into the consumer, so `HasConsumedContextChanged` was never consulted. The
consumer kept rendering the stale context value, and event handlers it created
captured that stale value.

The skip was reachable at two independent layers, so a single-site fix was
insufficient (verified: removing either guard re-breaks the repro).

### Fix
Both skip fast paths are now context-aware:

- **`Reconciler.Update.cs`** — before the element-level `ShallowEquals` skip,
  decline the skip when a consumed context value changed in the subtree.
- **`ChildReconciler.cs` (`UpdateCommonChild`)** — before the per-child
  `CanSkipUpdate` skip, run the same check and fall through to a real update.

Both are gated by the new `ContextScope.HasActiveValues` so no subtree walk
happens when no provider is active, and are driven by
`Reconciler.HasConsumedContextChangedInSubtree(control)`, which checks the
control's `ComponentNode` and recurses via `ForEachReactorChildControl`.

### Tests
New self-test regression fixture
`Issue811ContextConsumerSkipFixtures.ReferenceStableChildSkip_ContextConsumerRerenders`
(registered in `SelfTestFixtureRegistry`) covering mount, provider-toggle
re-render, and current-vs-stale context use from an event handler.

### Validation
- Focused host selftest `Issue811_ReferenceStableChildSkip_ContextConsumerRerenders`: **8/8** checks pass.
- `Reactor.Tests` slice (`ChildReconcilerStructuralSkip`, `ContextSystemSelfHost`, `MemoizationSelfHost`, `ComponentModelIntegration`): **56 passed / 0 failed**.
- Full `Reactor.SelfTests` batch: **passed**.


src/Reactor/Core/Reconciler.Update.cs
```diff
@@ -93,11 +93,16 @@ public sealed partial class Reconciler
...
+        if (HasActiveContextValues)
+            subtreeContextChanged = HasConsumedContextChangedInSubtree(control);
+
         if (Element.ShallowEquals(oldEl, newEl)
             && Element.ModifiersEqual(oldModifiers, modifiers)
             && oldEl.HasCallbacks == newEl.HasCallbacks
             && !ForceRenderThroughWrapper(newEl)
-            && !IsOnDirtyAncestorPath(control))
+            && !IsOnDirtyAncestorPath(control)
+            && !subtreeContextChanged)
```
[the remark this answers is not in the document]
azchohfi: Good catch, fixed in f7f2485d. I folded the walk into the skip predicate as the last `&&` term: `!(HasActiveContextValues && HasConsumedContextChangedInSubtree(control))`. Now short-circuiting means the recursive subtree walk only runs after the cheap structural gates (ShallowEquals/ModifiersEqual/etc.) all pass, so a non-skippable element never pays for it, and it's skipped entirely when no provider is on the stack.

Note this only covered the element-level Update path. The per-child arm in ChildReconciler already gates the walk behind `CanSkipUpdate` + `HasActiveContextValues` before the `children.Get(i)` fetch, so it's fine. The deeper follow-up is that the gate keys off `HasActiveValues` (any provider present) rather than whether a provided value actually changed this render, so stable subtrees still get walked even when no context changed. Worth a follow-up using a context version signal.

src/Reactor/Core/ChildReconciler.cs
```diff
@@ -213,9 +213,17 @@ private static void UpdateCommonChild(
...
         var newEl = newChildren[i];
+        UIElement? existingControl = null;
         if (Element.CanSkipUpdate(oldEl, newEl)
             && !reconciler.ForceRenderThroughWrapper(newEl))
         {
+            if (reconciler.HasActiveContextValues)
+            {
+                existingControl = children.Get(i);
+                if (reconciler.HasConsumedContextChangedInSubtree(existingControl))
+                    goto perform_update;
```
azchohfi: This context-aware skip decline only lives here in the positional `UpdateCommonChild` arm, so the fix is incomplete for keyed lists. The other skip sites in this file don't have it:

- keyed prefix skip (~L313) and keyed suffix skip (~L362) still `continue` on `CanSkipUpdate` with no `HasConsumedContextChangedInSubtree` check, so a keyed stable child (grid rows, keyed list items) that consumes context still goes stale on a provider change — the exact #811 bug via the keyed path;
- the `ChildDiffHints` fast-path (~L111-143) skips untouched indices wholesale and only calls `UpdateCommonChild` for changed indices, so it bypasses the check too.

The rule now exists in three shapes (this `goto perform_update`, the keyed `continue`, and the `&&` chain in Reconciler.Update.cs). I'd pull it into a single `ShouldDeclineSkip(oldEl, newEl, control)` helper used by all four sites — that closes the keyed/hint gaps in one place and lets this arm drop the one-off `goto`. Please add a keyed selftest variant to lock it in.

src/Reactor/Core/ContextScope.cs
```diff
@@ -48,5 +48,7 @@ internal T Read<T>(Context<T> context)
         return context.DefaultValueBoxed;
     }
 
+    internal bool HasActiveValues => _stack.Count > 0;
```
azchohfi, line 51: `HasActiveValues` gates the subtree walk on *any* provider being on the stack, not on whether a provided value actually changed this render. With a provider near the app root (the common case), every reference-stable skip now pays a recursive `HasConsumedContextChangedInSubtree` walk even when no context changed — a real cost on the hot skip-floor.

There's already a `_version` here that bumps on every push/pop. Consider deriving a cheaper change signal from it (or a per-node "consumed-context version at last render") so stable subtrees are only walked when a provided value actually changed, rather than whenever any provider is present.

src/Reactor/Core/Reconciler.cs
```diff
@@ -1908,6 +1910,27 @@ private bool HasConsumedContextChanged(ComponentNode node)
...
     }
 
+    internal bool HasConsumedContextChangedInSubtree(UIElement control)
+    {
+        if (_componentNodes.TryGetValue(control, out var node)
+            && HasConsumedContextChanged(node))
+            return true;
+
+        bool found = false;
+        ForEachReactorChildControl(control, child =>
```
azchohfi, line 1920: This walk is only as complete as `ForEachReactorChildControl`, which covers Panel / ItemsRepeater / Border / ScrollViewer / UserControl / ContentControl but not `SplitView` (Pane/Content) or `Viewbox` (Child) — both real Reactor child hosts. A context consumer under one of those inside a reference-stable wrapper will be missed and stay stale, so the fix has a blind spot.

Better to drive traversal off the descriptor child metadata / `GetCurrentChild` accessors so every V1 child-hosting strategy is covered, rather than an ad-hoc control-type switch. (Unmount is unaffected — it uses strategy-driven teardown — but this walk and the nav-lifecycle traversals that reuse the helper share the gap.)

tests/Reactor.AppTests.Host/SelfTest/Fixtures/Issue811ContextConsumerSkipFixtures.cs
```diff
@@ -0,0 +1,85 @@
...
+            var host = H.CreateHost();
+            host.Mount(ctx =>
+            {
+                var (interactive, setInteractive) = ctx.UseState(true);
+
+                return VStack(
+                        TextBlock(interactive ? "surface:on" : "surface:off"),
+                        Button("Toggle interactive", () => setInteractive(!interactive)),
+                        stableOverlay)
+                    .Provide(InteractiveCtx, interactive);
```
azchohfi, line 63: Solid repro for the positional/unkeyed case. To cover the rest of the fix, worth adding three variants:

1. **Keyed:** put `.Key(...)` on `stableOverlay` so it routes through the keyed prefix/suffix skip arms (currently uncovered — see the ChildReconciler comment).
2. **Nested:** wrap the overlay in a reference-stable non-component container (a stable `Border`/`ScrollViewer`) so the recursive walk in `HasConsumedContextChangedInSubtree` is actually exercised — here the consumer is a direct child, one level deep.
3. **Negative:** re-render from unrelated state with the provider value unchanged and assert `probe.RenderCount` does not increase, so the coarse `HasActiveValues` gate can't regress into over-rendering.

azchohfi (commented):
## Review summary — context-consumer re-render fix (#811)

Thanks for tackling this. The core mechanism — declining the skip when a consumed context changed, via `HasConsumedContextChangedInSubtree` — is sound, and the primary case (a reference-stable component sibling under `.Provide(...)`) is fixed and covered by the new selftest.

My main concern is that **the fix is incomplete for keyed lists**, which I'd treat as a blocker. Details are in the inline comments; summarizing here:

### Blocking
- **Keyed + hint skip paths bypass the new check.** The context-aware decline only lives in the positional `UpdateCommonChild` arm and the element-level shallow skip. The keyed prefix (`ChildReconciler.cs` ~L313) and suffix (~L362) arms, and the `ChildDiffHints` fast-path (~L111–143), still skip on `CanSkipUpdate` with no context check — so a **keyed** stable child (grid rows, keyed list items) that consumes context still goes stale on a provider change. That's the same #811 bug via the keyed path. I'd centralize the decision into one `ShouldDeclineSkip(oldEl, newEl, control)` helper shared by all four skip sites — that closes the gap in one place and lets the positional arm drop the one-off `goto`.

### Non-blocking
- **Traversal blind spot.** `ForEachReactorChildControl` doesn't cover `SplitView` (Pane/Content) or `Viewbox` (Child), both real Reactor child hosts — consumers under them inside a stable wrapper are missed.
- **Coarse gate / perf.** The walk is gated on `HasActiveValues` (any provider present), not on whether a value actually changed, so stable subtrees get walked speculatively on every skip while a provider is active. A `_version`-derived change signal would avoid it. (The earlier hot-path ordering nit from the Copilot review is already handled in `f7f2485d`.)
- **Test coverage.** Add keyed, deeper-nested (stable non-component container), and negative (unchanged-context ⇒ no re-render) selftest variants.
- **CHANGELOG.** No `### Fixed` entry for #811 (user-visible bug).

Happy to pair on the `ShouldDeclineSkip` refactor if useful.

src/Reactor/Core/ChildReconciler.cs
nsreehari: Addressed in 1e6ef077 (now on head 22a51088 after merging current main). The context-aware decline is centralized in ShouldDeclineSkip(control) and is now applied to the keyed prefix/suffix paths and ChildDiffHints early-out, while preserving the cheap predicates ahead of the subtree walk. Added keyed and hinted-range live fixtures. Validation after the main merge: dotnet build tests/Reactor.AppTests.Host/Reactor.AppTests.Host.csproj -p:Platform=x64; dotnet run --project tests/Reactor.AppTests.Host -p:Platform=x64 --no-build -- --self-test --filter "Issue811_" (3 groups, 19 assertions, 0 failures).

tests/Reactor.AppTests.Host/SelfTest/Fixtures/Issue811ContextConsumerSkipFixtures.cs
nsreehari: Added the keyed regression plus a ChildDiffHints regression in 1e6ef077 (head 22a51088). Both exercise the newly covered skip paths alongside the original positional fixture. Post-merge validation: dotnet build tests/Reactor.AppTests.Host/Reactor.AppTests.Host.csproj -p:Platform=x64; dotnet run --project tests/Reactor.AppTests.Host -p:Platform=x64 --no-build -- --self-test --filter "Issue811_" (3 groups, 19 assertions, 0 failures). The nested and unchanged-context variants remain useful non-blocking follow-ups, but are outside this blocker fix.

PR #855, opened 2026-07-09
Add app-mark icon to the app template and fix FontIconSource NaN crash

### LINKED ISSUE #854: [Feature] Give the default app template a title-bar app-mark icon
### Problem

The first thing anyone sees when they try Reactor is the app that `dotnet new reactorapp` scaffolds (the `WinUIApp-CSharp` template). The good news: it already opens with a **Mica** backdrop, an **extended/modern title bar** showing the app title, and a small interactive `Hello, {name}!` greeting. So most of the modern-Windows-11 baseline is already there.

The one thing missing is an **app-mark icon** in the title bar. A modern Windows 11 caption almost always carries a small app mark on the left next to the title; without one the title bar looks a touch unfinished, and the template misses a natural chance to show authors the `TitleBar.Icon(...)` API.

### Proposed solution

Give the template's title bar a placeholder app-mark icon — a Segoe Fluent Icons *CircleRing* glyph (`\uEA3A`) — with a comment pointing authors at a bundled `.ico` for when they add their own. That is the **whole** template change; the greeting, Mica backdrop, and extended title bar already ship on `main` and are left exactly as-is.

```csharp
var titleBar = TitleBar("Company.ReactorApp1")
    .Icon(FontIcon("\uEA3A", "Segoe Fluent Icons")) // CircleRing placeholder mark
    .Flex(shrink: 0);
```

### Additional context — a framework bug this surfaced

Adding that one icon line immediately crashed the scaffolded app. Placing a `FontIcon` with **no explicit size** into an icon-source slot (`TitleBar.Icon(...)`) threw `ArgumentException` from `TitleBar.set_IconSource` ("Value does not fall within the expected range" / E_INVALIDARG): `ResolveIconSource` was building the `FontIconSource` with `FontSize = NaN`. Reactor caught the exception and painted the error into the window instead of the app — so today a naive "just add a title-bar icon" doesn't actually work.

This is a latent framework bug independent of the template (it fires for any sizeless `FontIcon` in an `IconSource` slot — `TitleBar`, `TabView`, …). It is fixed in the same PR (#855) with a selftest regression.

### Alternatives considered

- **Ship a real `.ico` asset** in the template so the title bar and taskbar share one icon. Deferred: it adds binary assets, an `Assets/` folder, and an `<ApplicationIcon>` to every scaffold. A font glyph is zero-asset and trivially replaceable — the template comment points authors at `.Icon("ms-appx:///Assets/AppIcon.ico")` for when they add their own.
- **Leave the title bar iconless.** Less polished, and it skips a natural teaching moment for the `TitleBar.Icon(...)` API.

### Confirmation

- [x] I have searched existing issues and specs for prior discussion of this idea.

niels9001:
## Summary

Gives the app scaffolded by `dotnet new reactorapp` a **title-bar app-mark icon** — the one bit of modern-title-bar polish the template was missing — and fixes a framework bug in icon-source resolution that adding the icon surfaced.

The template already shipped a Mica backdrop, an extended/modern title bar, and an interactive `Hello, {name}!` greeting. This PR leaves all of that untouched and **only adds the icon**.

## Linked issue

Fixes #854

## What changed

**Template** — `tools/Templates/templates/WinUIApp-CSharp/App.cs` (one line + a comment)

Add a placeholder app-mark icon to the title bar: a Segoe Fluent Icons *CircleRing* glyph (`\uEA3A`). A comment points authors at `.Icon("ms-appx:///Assets/AppIcon.ico")` for when they add a real asset.

```diff
-        var titleBar = TitleBar("Company.ReactorApp1").Flex(shrink: 0);
+        var titleBar = TitleBar("Company.ReactorApp1")
+            .Icon(FontIcon("\uEA3A", "Segoe Fluent Icons"))
+            .Flex(shrink: 0);
```

**Framework fix** — `src/Reactor/Core/V1Protocol/IconResolver.cs`

Adding that icon hit a runtime crash. Reactor caught the exception and painted it into the window instead of rendering the app — so "just add a title-bar icon" didn't actually work before this fix.

- **What happened:** `ResolveIconSource(IconData)` built the `FontIconSource` for the icon-source slot with an object initializer that always set `FontSize = fi.FontSize ?? double.NaN` (and `FontFamily = null!` when unspecified). A `FontIconSource` whose `FontSize` is `NaN` is rejected by `TitleBar.set_IconSource` with `ArgumentException: Value does not fall within the expected range` (E_INVALIDARG). So **any** `FontIcon` placed in an `IconSource` slot (`TitleBar`, `TabView`, …) *without* an explicit size threw. The parallel `IconElement` path (`CreateFontIcon`) was already correct — it only assigns `FontFamily`/`FontSize` when specified — which is why the bug stayed hidden (existing icon tests always passed an explicit size and never assigned the result to a real control).
- **How it's fixed:** replaced the initializer with a new `CreateFontIconSource(fi)` helper that mirrors `CreateFontIcon` — `Glyph` always, `FontFamily`/`FontSize` only when specified — leaving them at their control defaults otherwise.

**Test**

- Add selftest regression `PrivMount_TitleBarFontIconNoSize` (`ReconcilerBigCoverageFixtures.cs`): resolve a no-size `FontIcon` and assign it to a real `TitleBar.IconSource`, asserting no `ArgumentException`. Fails without the fix, passes with it.

## Test plan

- [x] `dotnet build Reactor.slnx` — clean.
- [x] Selftest: `dotnet run --project tests\Reactor.AppTests.Host -p:Platform=x64 -- --self-test --filter "PrivateMountHotPaths"` → `ok PrivMount_TitleBarFontIconNoSize`, `# Total failures: 0`.
- [x] Manually scaffolded and ran the template app: modern title bar now shows the CircleRing app mark + title over the Mica surface, no error panel.

## Risk / breaking changes

- **Template:** one added icon line; scaffolded apps now show an app mark in the title bar. No public API change, and nothing else about the template changes.
- **Framework fix is a strict bug fix.** `FontIcon`s that specified a size are unchanged; ones that omitted a size now correctly inherit the control's default size instead of throwing. No regression for callers that already worked.

### DIFF
--- a/tests/Reactor.AppTests.Host/SelfTest/Fixtures/ReconcilerBigCoverageFixtures.cs
+++ b/tests/Reactor.AppTests.Host/SelfTest/Fixtures/ReconcilerBigCoverageFixtures.cs
@@ -1746,6 +1746,27 @@ internal static class ReconcilerBigCoverageFixtures
             };
             H.Check("PrivMount_IconSources", iconSources.Take(6).All(i => i is not null));
 
+            // Regression: a FontIcon with NO explicit FontSize must resolve to a
+            // FontIconSource that TitleBar.IconSource accepts. The IconSource path
+            // used to force FontSize = double.NaN, which TitleBar.set_IconSource
+            // rejects with ArgumentException ("Value does not fall within the
+            // expected range") — the exact failure the modernized app template hit
+            // via TitleBar(...).Icon(FontIcon("\uEA3A", "Segoe Fluent Icons")).
+            // Assign to a real TitleBar on the UI thread to lock the fix in.
+            var barFontIconNoSize = IconResolver.ResolveIconSource(
+                new FontIconData("\uEA3A", "Segoe Fluent Icons"));
+            bool titleBarAcceptedFontIcon;
+            try
+            {
+                _ = new WinXC.TitleBar { IconSource = barFontIconNoSize };
+                titleBarAcceptedFontIcon = true;
+            }
+            catch (ArgumentException)
+            {
+                titleBarAcceptedFontIcon = false;
+            }
+            H.Check("PrivMount_TitleBarFontIconNoSize", titleBarAcceptedFontIcon);
+
             var rows = new[] { new KeyRow("a"), new KeyRow("b"), new KeyRow("c") };
             int selected = -1;
             IReadOnlyList<KeyRow>? multi = null;
--- a/tests/Reactor.AppTests.Host/SelfTest/Fixtures/ReconcilerBigCoverageFixtures.cs
+++ b/tests/Reactor.AppTests.Host/SelfTest/Fixtures/ReconcilerBigCoverageFixtures.cs
@@ -1752,20 +1752,28 @@ internal static class ReconcilerBigCoverageFixtures
             // rejects with ArgumentException ("Value does not fall within the
             // expected range") — the exact failure the modernized app template hit
             // via TitleBar(...).Icon(FontIcon("\uEA3A", "Segoe Fluent Icons")).
-            // Assign to a real TitleBar on the UI thread to lock the fix in.
-            var barFontIconNoSize = IconResolver.ResolveIconSource(
-                new FontIconData("\uEA3A", "Segoe Fluent Icons"));
-            bool titleBarAcceptedFontIcon;
-            try
+            // The parallel null-FontFamily slot was coerced the same way (FontFamily
+            // = null!), so cover it too. For each case assert the resolver produced a
+            // real FontIconSource (not null / some other type) AND that a real
+            // TitleBar accepts the assignment on the UI thread.
+            static bool TitleBarAcceptsFontIconSource(IconData data)
             {
-                _ = new WinXC.TitleBar { IconSource = barFontIconNoSize };
-                titleBarAcceptedFontIcon = true;
+                if (IconResolver.ResolveIconSource(data) is not WinXC.FontIconSource src)
+                    return false;
+                try
+                {
+                    _ = new WinXC.TitleBar { IconSource = src };
+                    return true;
+                }
+                catch (ArgumentException)
+                {
+                    return false;
+                }
             }
-            catch (ArgumentException)
-            {
-                titleBarAcceptedFontIcon = false;
-            }
-            H.Check("PrivMount_TitleBarFontIconNoSize", titleBarAcceptedFontIcon);
+
+            H.Check("PrivMount_TitleBarFontIconNoSize",
+                TitleBarAcceptsFontIconSource(new FontIconData("\uEA3A", "Segoe Fluent Icons"))
+                && TitleBarAcceptsFontIconSource(new FontIconData("\uEA3A", FontFamily: null, FontSize: 18)));
 
             var rows = new[] { new KeyRow("a"), new KeyRow("b"), new KeyRow("c") };
             int selected = -1;
--- a/tools/Templates/templates/WinUIApp-CSharp/App.cs
+++ b/tools/Templates/templates/WinUIApp-CSharp/App.cs
@@ -27,7 +27,12 @@ class App : Component
     {
         var (name, setName) = UseState("World");
 
-        var titleBar = TitleBar("Company.ReactorApp1").Flex(shrink: 0);
+        // App-mark icon in the title bar: a Segoe Fluent Icons glyph (U+EA3A,
+        // "CircleRing") used as a placeholder. Swap it for a bundled asset once
+        // you add one, e.g. .Icon("ms-appx:///Assets/AppIcon.ico").
+        var titleBar = TitleBar("Company.ReactorApp1")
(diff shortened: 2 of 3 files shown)


azchohfi (commented):
## Automated PR review — app-mark icon + FontIconSource NaN fix

Ran the repo's multi-dimensional `pr-review` skill (security, correctness, api-ergonomics, alternative-solution, test-coverage, docs-and-samples, packaging) plus a cross-model check (GPT-5.5, high reasoning) over `origin/main...HEAD`.

**Verdict: looks good to merge.** The fix is clean and the right shape — extracting `CreateFontIconSource` to mirror the existing `CreateFontIcon` and only assigning `FontFamily`/`FontSize` when specified is exactly how the `IconElement` path already works, so the two projections are now consistent. No security, correctness, API, packaging, or alternative-solution concerns surfaced.

Good call adding a real-`TitleBar` regression selftest that asserts the assignment no longer throws — that's a non-vacuous guard (it fails if `FontSize = double.NaN` is restored). Cross-check note: the fixture doesn't force a layout/render pass, so it relies on `TitleBar.set_IconSource` validating synchronously at assignment (which matches the app mount path and the reported failure) — worth keeping in mind if that WinUI behavior ever changes.

Only two minor, non-blocking notes (inline below):

- **M1 (medium, test-coverage):** the new test only isolates the missing-`FontSize` half of the change; the null-`FontFamily` branch isn't exercised on its own.
- **L1 (low, docs):** the template comment names glyph `U+EA3A` as "CircleRing", which independent sources dispute.

Summary: **Critical: 0  High: 0  Medium: 1  Low: 1**

tests/Reactor.AppTests.Host/SelfTest/Fixtures/ReconcilerBigCoverageFixtures.cs
```diff
@@ -1746,6 +1746,27 @@ public override async Task RunAsync()
...
 
+            // Regression: a FontIcon with NO explicit FontSize must resolve to a
+            // FontIconSource that TitleBar.IconSource accepts. The IconSource path
+            // used to force FontSize = double.NaN, which TitleBar.set_IconSource
+            // rejects with ArgumentException ("Value does not fall within the
+            // expected range") — the exact failure the modernized app template hit
+            // via TitleBar(...).Icon(FontIcon("\uEA3A", "Segoe Fluent Icons")).
+            // Assign to a real TitleBar on the UI thread to lock the fix in.
+            var barFontIconNoSize = IconResolver.ResolveIconSource(
+                new FontIconData("\uEA3A", "Segoe Fluent Icons"));
```
azchohfi: **M1 (medium, test-coverage):** This regression case pins the `FontSize`-unset path, but it passes a non-null `FontFamily` (`"Segoe Fluent Icons"`), so the *other* half of the change — the `if (fi.FontFamily is not null)` branch in `CreateFontIconSource` — is never isolated. The pre-fix code also set `FontFamily = fi.FontFamily is null ? null! : ...`, so a null family was a distinct (and also potentially control-rejected) path.

Consider an adjacent check with a null family, e.g. `new FontIconData("\uEA3A", fontFamily: null, fontSize: 18)` assigned to a fresh `TitleBar`, so a future regression in the FontFamily branch can't slip through. Not a blocker.

tools/Templates/templates/WinUIApp-CSharp/App.cs
```diff
@@ -27,7 +27,12 @@ public override Element Render()
     {
         var (name, setName) = UseState("World");
 
-        var titleBar = TitleBar("Company.ReactorApp1").Flex(shrink: 0);
+        // App-mark icon in the title bar: a Segoe Fluent Icons glyph (U+EA3A,
```
azchohfi: **L1 (low, docs):** The comment labels `U+EA3A` as "CircleRing", but that glyph name doesn't hold up — the Segoe Fluent Icons mapping for `U+EA3A` is reported elsewhere as something else (e.g. "ContactInfo"/"Movies" depending on source), so the name is at best ambiguous. Since the code only depends on the raw codepoint, safest to either drop the name and call it a "placeholder glyph", or confirm it against the official Segoe Fluent Icons list. Purely cosmetic.

niels9001:
Thanks for the review! Addressed the nits in fcf221ac:

- **M1 / @Copilot (test-coverage):** The selftest now asserts `ResolveIconSource` returns a real `FontIconSource` (not just that `TitleBar.IconSource` assignment doesn't throw), and adds a second case with a **null `FontFamily`** (`new FontIconData("\uEA3A", FontFamily: null, FontSize: 18)`) so the `if (fi.FontFamily is not null)` branch of `CreateFontIconSource` is exercised in isolation. Both cases are checked via a shared `TitleBarAcceptsFontIconSource` helper.
- **L1 (docs):** Dropped the disputed `"CircleRing"` name for `U+EA3A` in the template comment — it's now just a "placeholder Segoe Fluent Icons glyph".

Verified: `PrivMount_TitleBarFontIconNoSize` → `ok`, `# Total failures: 0`.

PR #899, opened 2026-07-16
Pin /perf PR head to the trigger-time commit and gate the sticky comment

azchohfi:
## What

Two adjustments to `.github/workflows/perf-compare.yml` so a `/perf` run always benchmarks — and comments on — exactly the commit that triggered it.

### 1. Pin the PR head at trigger time
- New early **Pin PR head SHA** step resolves the PR head via the API and records it, running *before* the `.NET` setup + PerfLib steps rather than resolving the head live several minutes later.
- **Resolve PR + create PR worktree** now consumes that pinned SHA, fetches `pull/N/head`, and re-verifies the fetched head still equals the pinned SHA (`-cne`) before creating the build worktree. If the branch moved on after the trigger, the step stops with an actionable message instead of building a different commit.
- For comment triggers, a head whose commit post-dates the `/perf` comment is rejected (best-effort on the commit-date lookup; the pinned-SHA re-verification is the hard check either way). `workflow_dispatch` keeps working unchanged.

### 2. Only publish a comment the benchmark actually produced
- **Post / update sticky comment** now publishes `perf-out/comment.md` only when the benchmark step''s `outcome == success` (the trusted script writes that file as the final action of a clean run). On any non-success it synthesizes the neutral fallback, overwriting whatever is on disk, so a leftover or incomplete file is never posted.

### Docs
- `tests/stress_perf/ci/README.md` updated to describe the pinned-head resolution and the success-gated comment.

## Notes
- A degraded/partial run (previously posted a low-confidence table) now shows the neutral fallback pointing at the run log + `perf-logs` artifact.

## Validation
- Workflow YAML parses; the three changed PowerShell step bodies parse cleanly.
- Simulated the pin/equality/gate decision points (timezone-aware date comparison, exact SHA match, publish-vs-fallback matrix incl. the skipped-step case).

### DIFF
--- a/.github/skills/perf-compare/SKILL.md
+++ b/.github/skills/perf-compare/SKILL.md
@@ -175,6 +175,8 @@ git worktree remove ../perf-main
 - This skill is the **local** equivalent of the `/perf` PR workflow
   (`.github/workflows/perf-compare.yml`); both use the same scripts and render
   the same comparison. Use this before pushing; use `/perf` for the
-  reviewer-visible comment on the PR.
+  reviewer-visible comment on the PR. The workflow publishes that rendered
+  comparison only when the benchmark run completes successfully; on a failed or
+  incomplete run it posts a neutral fallback comment linking the run log instead.
 - It is unrelated to the **startup** perf harness under `tests/startup_perf/`
   (ETW/WPR TTFP/TTI measurement) — do not invoke that here.
--- a/.github/workflows/perf-compare.yml
+++ b/.github/workflows/perf-compare.yml
@@ -297,7 +297,11 @@ jobs:
         shell: pwsh
         env:
           GH_TOKEN: ${{ github.token }}
-          PR_NUMBER: ${{ steps.resolve.outputs.pr_number }}
+          # Resolve from the event, not steps.resolve — a fetch/head-mismatch throw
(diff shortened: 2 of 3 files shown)


Hemantxk (approved):
I see the CI.yml also captures SHA, but it captures it through HEAD_SHA: ${{ github.event.pull_request.head.sha }}, which seems to fetch right hash. Just curious would it be easier to follow the similar approach here?

azchohfi:
@Hemantxk thanks! That works in `ci.yml` because it's triggered by `pull_request`, so the event payload carries the PR object and `github.event.pull_request.head.sha` is populated.

`perf-compare.yml` runs on `issue_comment` (the `/perf` comment) and `workflow_dispatch`, and neither payload includes `github.event.pull_request` — `issue_comment` only has `github.event.issue`/`github.event.comment`, and `workflow_dispatch` only has `inputs.pr_number` — so `github.event.pull_request.head.sha` would be empty here. The `gh api repos/${{ github.repository }}/pulls/$pr --jq .head.sha` call in the "Pin PR head SHA" step does the equivalent trigger-time capture for a comment/dispatch trigger. We keep it on `issue_comment` on purpose so the perf scripts always run from the trusted default branch.

PR #939, opened 2026-07-29
Never write FlyoutPlacementMode.Auto to a WinUI FlyoutBase (fixes the Flyout default-placement process kill)

azchohfi:
## Symptom

Clicking **Show Flyout** on the gallery's **Dialogs and Flyouts → Flyout** page terminated the process — a `0xC000027B` stowed exception carrying `System.ArgumentException: The parameter is incorrect`. Because it is a fail-fast, there is no catchable error; the app just disappears.

## Root cause

Quoted from the bug-hunt report — this chain was established with `cdb` + public symbols by the reporting session, **not** re-derived here:

> 2. `cdb` + public symbols resolved that address to `DirectUI::FlyoutBase::ShowAtCore+0x42` — the instruction directly after `call DirectUI::FlyoutBase::ValidateAndSetParameters`.
> 3. Disassembling `ValidateAndSetParameters` shows its placement switch accepts **0–12** and falls through to `mov edi,80070057h` (E_INVALIDARG) for anything else. `FlyoutPlacementMode.Auto == 13`.
> 4. Disassembling `GetEffectivePlacement` shows that with no cached override it returns the raw `Placement` property, so `Auto` reaches the validator unresolved.
> 5. A/B in a single build on a single page: card 1 pinned to `Placement.Top` opens normally; card 2 left at the default `Auto` crashes.

What I verified myself, and how:

- **From the published API reference:** `FlyoutPlacementMode.Auto == 13`, and `FlyoutBase.Placement`'s own default is **`Top`** — WinUI itself never puts `Auto` on that DP.
- **Reproduced in-process:** see the revert-proof output below.

Reactor defaults `Placement` to `Auto` on its flyout element records and then assigned it to the live WinUI flyout unconditionally. The correct guard already existed — but only for `MenuFlyout` (`Reconciler.cs`, comment: *"Auto can cause assertions on MenuFlyout"*).

## What changed

New internal choke point `src/Reactor/Core/FlyoutPlacement.cs`:

```csharp
internal static bool ShouldApply(FlyoutPlacementMode placement) => placement != FlyoutPlacementMode.Auto;

internal static void Apply(FlyoutBase flyout, FlyoutPlacementMode placement)
{
    if (!ShouldApply(placement)) return;
    if (flyout.Placement != placement) flyout.Placement = placement;
}
```

Routed through it:

| File | Element | Why |
|---|---|---|
| `V1Protocol/OverlayLifecycle.cs` ×3 | `FlyoutElement` | **crashed** |
| `Reconciler.cs` ×2 | `ContentFlyoutElement` (`.WithFlyout()`, `.WithContextFlyout()`, button `flyout:` slots) | **crashed** |
| `Reconciler.cs` ×2 | `MenuFlyoutContentElement` | already guarded; folded in so the helper is the only write site |

Object-initializer writes became post-construction `Apply` calls placed **before** `ApplySetters`, so a caller's `.Set(f => f.Placement = ...)` still wins.

### `CommandBarFlyout` — same bug, guarded by #943 (now merged)

**Correction to an earlier revision of this PR.** It previously said `CommandBarFlyout` was "deliberately not guarded" because it "resolves `Auto` itself and auto-positions". **That was wrong.** `CommandBarFlyout` is affected by exactly this crash. The reason it never crashed is that **it never opened**: `MountCommandBarFlyout` installed the flyout as `FlyoutBase.AttachedFlyout` metadata that nothing ever called `ShowAttachedFlyout` on, so the placement validator was never reached. A latent crash masked by a separate defect reads exactly like a working code path.

**I confirmed this independently rather than taking it on report.** I had also written that it was unobservable "because `CommandBarFlyout` cannot be opened from its target" — that was wrong too. The broken wiring only blocked the *target/click* path; `ShowAt` can be called directly. Doing so, with a control to prove the harness itself works:

```
# probe[Top]  flyout=found placement=Top
# probe[Top]  ShowAt returned, IsOpen=True          ← control: a valid placement opens fine
# probe[Auto] flyout=found placement=Auto
# probe[Auto] THREW ArgumentException: Value does not fall within the expected range
```

Same exception as the `Flyout` crash this PR fixes. So `CommandBarFlyout` retains `Auto` on the DP quite happily and throws the moment it is shown — "the DP accepts it" and "it can be shown" are different claims, and only the second one matters. That probe was a one-off measurement and is not checked in: asserting "this throws" would be wrong now that #943 has landed, and #943 owns the positive assertion.

**[#943](https://github.com/microsoft/microsoft-ui-reactor/pull/943) has since merged**, fixing the wiring and guarding those three sites via `Reconciler.ApplyFlyoutPlacement`. So the affected sites are partitioned across two changes:

| Change | Guards | Helper |
|---|---|---|
| **#939 (this one)** | `FlyoutElement`, `ContentFlyoutElement`, `MenuFlyoutContentElement` | `FlyoutPlacement.Apply` |
| **#943 (merged)** | the three `CommandBarFlyout` sites | `Reconciler.ApplyFlyoutPlacement` |

This branch is rebased onto that merge and leaves those three sites **byte-identical to `main`**.

**The two were verified to compose — before #943 landed, and again after.** Ahead of the merge I simulated #943 locally and found **three** of my tests would have failed, so whichever PR merged second would have gone red:

```
No_Flyout_Placement_Write_Bypasses_FlyoutPlacement   [FAIL]   ← my scan flagged #943's helper as a bypass
CommandBarFlyout_Is_Deliberately_Left_Unguarded      [FAIL]   ← asserted the exemption was non-empty
not ok FlyoutPlacement_CommandBarFlyout_KeepsAuto             ← asserted the DP stays Auto
```

The first was the serious one: my architecture test would have failed on #943's *correct* helper, because `flyout.Placement = placement` inside `ApplyFlyoutPlacement` matched the "no raw writes outside the choke point" rule. Fixed by teaching the scan to recognise `ApplyFlyoutPlacement` as a legitimate choke point — which also exposed a real blind spot, since the matcher had not recognised `ClearValue(FlyoutBase.PlacementProperty)` as a placement write at all. The other two tests asserted invariants that #943 correctly invalidates and were removed. **Against the real merged #943, all 22 unit tests and all 9 flyout fixtures pass.**

**The helpers now agree — I converged on `ClearValue`.** They previously differed on an `explicit → Auto` update: this one skipped the write and retained the last explicit value, #943's cleared the DP. I had flagged that for a follow-up, but [#952](https://github.com/microsoft/microsoft-ui-reactor/issues/952) turns it from tidiness into correctness: skipping leaves the previous placement behind as a **local** DP value, and in WinUI precedence a local value outranks a `Style` setter — so an app returning to `Auto` could never get its styled placement back, and "no opinion" would silently mean "whatever I last said". That is the same precedence defect #952 documents for the common modifiers. `FlyoutPlacement.Apply` now clears the property for `Auto`, so both helpers behave identically and the update fixture asserts the transition lands on the platform default (read from a fresh `WinUI.Flyout`, not hard-coded). Folding the two into a single helper is mechanical and left as a follow-up, since it would mean rewriting call sites that landed with #943 — tracked as [#953](https://github.com/microsoft/microsoft-ui-reactor/issues/953), whose *behavioural* half this PR already implements; only the structural de-duplication remains.

### Design choice

**Chosen: guard the write.** `Auto` stays a meaningful value in the Reactor API for the flyout types that can honour it, and this mirrors the guard that already existed for `MenuFlyoutContentElement`.

**Rejected: change the element defaults to `Top`.** It changes observable public API defaults that user code can read via `element.Placement`, for no behavioural gain.

**Adopted (after initially rejecting it): `ClearValue(FlyoutBase.PlacementProperty)` on `Auto`.** I first rejected this, partly arguing it "would wipe any local value a control's own constructor had set" — measuring the platform showed that is false (`Flyout`, `MenuFlyout` and `CommandBarFlyout` all default to `Top`, with no constructor-set local value), so on mount clearing and skipping are equivalent. I then kept skipping on the weaker ground that it matched the pre-existing MenuFlyout guard. #952 settled it: on an `explicit → Auto` update, skipping leaves a stale **local** value that outranks a `Style` setter. Clearing is now what the helper does, and the update fixture asserts it.
**Rejected: a `REACTOR_*` analyzer** warning on `Flyout(...)` without an explicit placement — the report is explicit that it "would paper over R1; fix the framework instead".

No public API changed (`FlyoutPlacement` is `internal`), so no `mur --regen-api` / `reactor.api.txt` churn. No `docs/guide/` topic documents flyout placement, so no template recompile. No analyzer rule added, so **no merge-ordering hazard** from the `TreatWarningsAsErrors`-in-Release interaction — and `dotnet build Reactor.slnx -c Release` is clean on this branch rebased onto `main` **after** #930 and #932.

## Tests

**Unit — `tests/Reactor.Tests/FlyoutPlacementGuardTests.cs` (22 tests)**
- Full-enum table: `ShouldApply` true for the 13 values WinUI's validator accepts, false only for `Auto`; plus `(int)Auto == 13` and "`Auto` is the only value outside 0..12" pinned directly.
- **Roslyn structural scan** over every `src/Reactor/**/*.cs`: the only site writing a WinUI flyout `Placement` must be `FlyoutPlacement.cs`. Matches member assignments, `new ...Flyout { Placement = ... }` initializers, and `SetValue`/`SetCurrentValue(FlyoutBase.PlacementProperty, ...)`.
- The `CommandBarFlyout` carve-out is **scoped by enclosing method name**, not file or line, so it survives the in-flight rework of those methods and cannot leak to other writes in the same file. `Scanner_Scopes_The_Exemption_To_The_Enclosing_Method` proves two *identical* writes are classified differently purely by which method they sit in; `CommandBarFlyout_Is_Deliberately_Left_Unguarded` pins the asymmetry against a future "consistency" cleanup.
- Anti-vacuity: the scanner must parse >100 files, must find the helper's own write, and must flag a synthetic bypass of all three shapes while ignoring near-misses (element-record initializers, `PreferredPlacement`, reads).

**Selftests — `FlyoutPlacementFixtures.cs` (9 fixtures, 44 checks)**, registered in both `AllFixtures` and the `Create()` switch. The `Flyout`/`ContentFlyout`/fresh-create fixtures actually **open** the flyout — the only tier that catches this crash. Every mount/create path carries two arms: a *"never `Auto`"* check (fails if the bug returns) and a *differential explicit-placement* check (fails if the `Apply` call is deleted outright, which the structural scan cannot see).

Covering `Flyout(...)`, `.WithFlyout(ContentFlyout(...))`, `.WithContextFlyout(MenuItems(...))`, `DropDownButton`/`SplitButton` flyout slots, `UpdateFlyoutElement`'s fresh-create branch, the `Auto → Left → Auto` update sequence (with a render witness so step 2 can't pass trivially), and the inverted `CommandBarFlyout` case above.

### Revert-proof — re-run against the final test set, on the rebased base

**1. Restore one raw write** (`OverlayLifecycle.MountFlyout` writes `Placement = flyEl.Placement` again):

```
Microsoft.UI.Reactor.Tests.FlyoutPlacementGuardTests.No_Flyout_Placement_Write_Bypasses_FlyoutPlacement [FAIL]
Failed!  - Failed:     1, Passed:    21, Skipped:     0, Total:    22

1..9
not ok FlyoutPlacement_Default_DpIsNotAuto - assertion failed
not ok 1 FlyoutPlacement_Flyout_DefaultPlacement_Opens_CRASH - ArgumentException: Value does not fall within the expected range.
System.ArgumentException: Value does not fall within the expected range.
not ok FlyoutPlacement_Update_MountedNotAuto - assertion failed
# Total failures: 3
```

That `ArgumentException` is the original crash, reproduced in-process (thrown out of `ButtonAutomationPeer.Invoke()` when the fixture clicks the button).

**2. Delete an `Apply` call outright** (`CreateFlyoutFromElement`, ContentFlyout arm):

```
1..9
not ok FlyoutPlacement_ContentFlyout_CreateAppliesExplicit - assertion failed
not ok FlyoutPlacement_SplitButton_CreateAppliesExplicit - assertion failed
# Total failures: 2

Passed!  - Failed:     0, Passed:    22, Skipped:     0, Total:    22
```

Note the unit suite passes 22/22 here: a *deletion* is invisible to a scan that looks for un-routed writes. That is exactly why each path also carries a differential explicit-placement assertion in the selftests. Both mutations were reverted and the tree verified clean afterwards.

## Verification

- `dotnet build Reactor.slnx -c Release` — **succeeded** (the config that turns warnings into errors).
- `dotnet test tests/Reactor.Tests` — **12600 passed, 0 failed**.
- Full selftest suite — **1350 fixtures, 0 failures**.
- **Platform assumption measured, not assumed.** `Platform_FlyoutBase_PlacementDefault` constructs each flyout type on the UI thread and reads the DP before anything writes it:

  ```
  # measured Flyout.Placement default           = Top
  # measured MenuFlyout.Placement default       = Top
  # measured CommandBarFlyout.Placement default = Top
  ```

  This is the assumption the whole fix rests on — "leave the DP alone" is only safe while the DP's own default is inside the validator's accepted `0..12`. Without this check, a future SDK defaulting `Placement` to `Auto` would silently disarm the guard while every other test here stayed green, since they all assert `!= Auto` against a DP that would by then *be* `Auto`. The values are emitted as TAP comments so an SDK bump shows up as a log diff rather than a silent pass.
- **Gallery, real mouse input.** Verified twice; the second run used the shared input mutex and an `AttachThreadInput` foreground force, because ~10 sibling agent sessions were driving their own gallery instances concurrently and real input is a single global resource. Dedicated instance, every `winapp ui` call scoped to its HWND, retry on any "not in the foreground" refusal:

  | Page | Click | Result |
  |---|---|---|
  | Flyout → "Show Flyout" | `✅ (926, 630)` | `Pane`/`className=Flyout` "This is flyout content!" at (801, 520) — **above** the button, i.e. `Top`. Process alive. Screenshot confirms. |
  | Flyout → "Cart Options" | `✅ (928, 896)` | `Empty Cart` at y=779, `isOffscreen: false`. Process alive. |
  | CommandBarFlyout → "Show Commands" | `✅ (950, 630)` | No crash. No commands surface — separate finding, below. |

  These gallery runs were performed before the final rebase; the product change they exercise is unchanged since.

## Review

Ran the repo's `.github/skills/pr-review` skill across all 8 dimensions. security / correctness / alternative-solution / packaging clean; the 5 medium+low findings were all **confirmed by the multi-model cross-check** on a different model family, and all fixed. Two further *suppressed low-confidence* Copilot comments (scan memoization, and a fixture comment that overstated what the suite does) were also fixed.

## ⚠️ Known gap for the merging human: the later commits are unreviewed

Copilot reviewed this PR three times and returned no comments each time — but **all three reviews predate a rebase, so none of them covers the branch as it now stands.** The commits they examined no longer exist: the rebase onto merged #943 replayed this work under new SHAs and orphaned the old ones, which is why the review entries point at objects unreachable from the head.

Reviewed content, by commit subject (subjects survive a rebase, SHAs do not):

- ✅ *Never write FlyoutPlacementMode.Auto to a WinUI FlyoutBase*
- ✅ *Address pr-review findings: docs, CHANGELOG, scanner and fixture coverage*
- ✅ *Record how the FlyoutBase internals in FlyoutPlacement were established*
- ✅ *Address the two low-confidence Copilot review comments*

**Not reviewed by anything but CI:**

- *Do not guard CommandBarFlyout's placement — Auto is correct there* — reverts the CommandBarFlyout guard so those sites stay byte-identical to `main`
- *Measure the FlyoutBase.Placement platform default instead of assuming it* — adds the selftest pinning the measured platform default
- *Correct the CommandBarFlyout rationale and make the guard compose with #943* — corrects documentation that stated a since-disproved rationale, and stops the structural scan flagging #943's helper as a bypass
- *Rebase onto merged #943 and update docs that described it as in-flight*
- *Converge the placement guard on ClearValue, matching ApplyFlyoutPlacement*

**Why it cannot be closed by re-requesting.** All three reviews are anchored to commits the rebase orphaned — verified: their `commit_id`s are `5cd7058b`, `205c5169` and `aa543b1a`, none reachable from the current head. Seven re-requests via `POST /pulls/939/requested_reviewers` were accepted (HTTP 200) without producing a further round, and the documented fallback `gh pr edit --add-reviewer "copilot-pull-request-reviewer[bot]"` fails with *"Could not resolve user with login"* (a known SAML/GraphQL issue in this org).

The same pattern was observed across every rebased PR in this batch, and another session traced it to review threads carrying a frozen `original_commit_id` that a rebase orphans. I could not confirm that mechanism on this PR, because it has zero review comments — so treat the mechanism as reported rather than verified here; what *is* verified is the orphaned anchors and that no review has landed since the rebase. Either way, further re-requests are not expected to recover it.

Stating it rather than counting a clean review as covering work it predates. CI is green on the head, the full unit and selftest suites pass locally, and the unreviewed delta is a revert *toward* `main` plus test and documentation changes — but a reviewer may reasonably want to look at those commits specifically.

## Out of scope — related findings deliberately untouched

- **`CommandBarFlyout` never opens from its target.** It is installed with `SetAttachedFlyout`, so clicking does nothing. Owned by another session, which is reworking `MountCommandBarFlyout`/`UpdateCommandBarFlyout`; this PR keeps those lines byte-identical to `main` so that work applies cleanly.
- **`FlyoutElement.IsOpen` is dead for `Button`/`SplitButton` targets.** `MountFlyout` installs the flyout into `Button.Flyout` via `SetFlyoutOnControl`, then calls `ShowAttachedFlyout`, which reads the *attached* slot — so `IsOpen` never opens anything for a button target. Queued as a separate follow-up; untouched here.
- A `.Placement(...)` fluent modifier for `FlyoutElement` doesn't exist (callers use `with { Placement = ... }`). Adding one is new public API requiring `mur --regen-api` and both `reactor.api.txt` copies — deliberately left out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,22 @@ Conventions for contributors:
 
 ### Fixed
 
+- **`Flyout(...)` no longer terminates the process when opened at its default
+  placement.** Reactor's flyout elements default `Placement` to
+  `FlyoutPlacementMode.Auto`, and that value was written straight onto the WinUI
+  `FlyoutBase.Placement` dependency property. WinUI's show-time validator
+  (`FlyoutBase::ValidateAndSetParameters`) only accepts placements `0..12`, so
+  `Auto` (13) failed with `E_INVALIDARG` and the resulting stowed
+  `ArgumentException` killed the app the first time the flyout was shown —
+  reproducible from the gallery's Dialogs and Flyouts → Flyout page. `Auto` now
+  means "Reactor has no opinion — let the platform decide" and is expressed by
+  leaving the dependency property at WinUI's own default (`Top`) instead of
+  writing `Auto` to it. Applies to `Flyout(...)`, `ContentFlyout(...)` /
+  `.WithFlyout(...)` / `.WithContextFlyout(...)`, `MenuItems(...)` and
+  `CommandBarFlyout(...)`. Explicit placements are unaffected; a flyout whose
+  placement changes from an explicit value back to `Auto` keeps the last
+  explicit value rather than resetting.
+
 - **`NavigationView` pane state no longer desyncs when the control moves its own
   pane (issue #916).** `IsPaneOpen` could be written but had no change
   notification, so a light dismiss or an adaptive display-mode change on resize
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,11 +103,11 @@ Conventions for contributors:
   its own documented default of `Top` and WinUI repositions from there. Applies
   to `Flyout(...)`, `ContentFlyout(...)` / `.WithFlyout(...)` /
   `.WithContextFlyout(...)` and `MenuItems(...)`. `CommandBarFlyout(...)` is
-  deliberately unchanged: it resolves `Auto` itself and positions automatically,
-  so suppressing its write would have pinned it to `Top`. Explicit placements are
-  unaffected everywhere; a guarded flyout whose placement changes from an
-  explicit value back to `Auto` keeps the last explicit value rather than
-  resetting.
+  affected by the same defect but is fixed separately, alongside the change that
+  makes it open from its target at all — until that lands it never reaches the
+  validator, because nothing ever shows it. Explicit placements are unaffected
+  everywhere; a guarded flyout whose placement changes from an explicit value
+  back to `Auto` keeps the last explicit value rather than resetting.
 
 - **`NavigationView` pane state no longer desyncs when the control moves its own
   pane (issue #916).** `IsPaneOpen` could be written but had no change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,14 +98,16 @@ Conventions for contributors:
   (`FlyoutBase::ValidateAndSetParameters`) only accepts placements `0..12`, so
   `Auto` (13) failed with `E_INVALIDARG` and the resulting stowed
   `ArgumentException` killed the app the first time the flyout was shown —
-  reproducible from the gallery's Dialogs and Flyouts → Flyout page. `Auto` now
-  means "Reactor has no opinion — let the platform decide" and is expressed by
-  leaving the dependency property at WinUI's own default (`Top`) instead of
-  writing `Auto` to it. Applies to `Flyout(...)`, `ContentFlyout(...)` /
-  `.WithFlyout(...)` / `.WithContextFlyout(...)`, `MenuItems(...)` and
-  `CommandBarFlyout(...)`. Explicit placements are unaffected; a flyout whose
-  placement changes from an explicit value back to `Auto` keeps the last
-  explicit value rather than resetting.
+  reproducible from the gallery's Dialogs and Flyouts → Flyout page. Reactor now
+  leaves the dependency property alone when the element says `Auto`, so it keeps
+  its own documented default of `Top` and WinUI repositions from there. Applies
+  to `Flyout(...)`, `ContentFlyout(...)` / `.WithFlyout(...)` /
+  `.WithContextFlyout(...)` and `MenuItems(...)`. `CommandBarFlyout(...)` is
+  deliberately unchanged: it resolves `Auto` itself and positions automatically,
+  so suppressing its write would have pinned it to `Top`. Explicit placements are
+  unaffected everywhere; a guarded flyout whose placement changes from an
+  explicit value back to `Auto` keeps the last explicit value rather than
+  resetting.
 
 - **`NavigationView` pane state no longer desyncs when the control moves its own
   pane (issue #916).** `IsPaneOpen` could be written but had no change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,7 +107,8 @@ Conventions for contributors:
   open from its target at all — before that it never reached the validator,
   because nothing ever showed it. Explicit placements are unaffected
   everywhere; a guarded flyout whose placement changes from an explicit value
-  back to `Auto` keeps the last explicit value rather than resetting.
+  back to `Auto` returns to that default, rather than leaving a stale local value
+  that would outrank a `Style` setter.
 
 - **`NavigationView` pane state no longer desyncs when the control moves its own
   pane (issue #916).** `IsPaneOpen` could be written but had no change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,10 +102,10 @@ Conventions for contributors:
   leaves the dependency property alone when the element says `Auto`, so it keeps
   its own documented default of `Top` and WinUI repositions from there. Applies
   to `Flyout(...)`, `ContentFlyout(...)` / `.WithFlyout(...)` /
-  `.WithContextFlyout(...)` and `MenuItems(...)`. `CommandBarFlyout(...)` is
-  affected by the same defect but is fixed separately, alongside the change that
-  makes it open from its target at all — until that lands it never reaches the
-  validator, because nothing ever shows it. Explicit placements are unaffected
+  `.WithContextFlyout(...)` and `MenuItems(...)`. `CommandBarFlyout(...)` was
+  affected by the same defect and is guarded by the companion fix that made it
+  open from its target at all — before that it never reached the validator,
+  because nothing ever showed it. Explicit placements are unaffected
   everywhere; a guarded flyout whose placement changes from an explicit value
   back to `Auto` keeps the last explicit value rather than resetting.
 
--- a/src/Reactor/Core/Element.cs
+++ b/src/Reactor/Core/Element.cs
@@ -5271,6 +5271,12 @@ public record FlyoutElement(
 ) : Element
 {
     public bool IsOpen { get; init; }
+    /// <summary>
+    /// Preferred position of the flyout relative to its target. Defaults to
+    /// <see cref="FlyoutPlacementMode.Auto"/>, which means "no opinion — let WinUI decide";
+    /// Reactor deliberately leaves <c>FlyoutBase.Placement</c> untouched in that case
+    /// because WinUI's own show-time validator rejects <c>Auto</c>.
+    /// </summary>
     public FlyoutPlacementMode Placement { get; init; } = FlyoutPlacementMode.Auto;
     public Action? OnOpened { get; init; }
     public Action? OnClosed { get; init; }
@@ -5290,6 +5296,12 @@ public record FlyoutElement(
 /// </summary>
 public record ContentFlyoutElement(Element Content) : Element
 {
+    /// <summary>
+    /// Preferred position of the flyout relative to its target. Defaults to
+    /// <see cref="FlyoutPlacementMode.Auto"/>, which means "no opinion — let WinUI decide";
+    /// Reactor deliberately leaves <c>FlyoutBase.Placement</c> untouched in that case
+    /// because WinUI's own show-time validator rejects <c>Auto</c>.
+    /// </summary>
     public FlyoutPlacementMode Placement { get; init; } = FlyoutPlacementMode.Auto;
 }
 
@@ -5299,6 +5311,12 @@ public record ContentFlyoutElement(Element Content) : Element
 /// </summary>
 public record MenuFlyoutContentElement(MenuFlyoutItemBase[] Items) : Element
 {
+    /// <summary>
+    /// Preferred position of the flyout relative to its target. Defaults to
+    /// <see cref="FlyoutPlacementMode.Auto"/>, which means "no opinion — let WinUI decide";
+    /// Reactor deliberately leaves <c>FlyoutBase.Placement</c> untouched in that case
+    /// because WinUI's own show-time validator rejects <c>Auto</c>.
+    /// </summary>
     public FlyoutPlacementMode Placement { get; init; } = FlyoutPlacementMode.Auto;
 }
 
@@ -6681,6 +6699,12 @@ public record CommandBarFlyoutElement(
     /// The target normally opens the flyout on click without this.
     /// </summary>
     public bool IsOpen { get; init; }
+    /// <summary>
+    /// Preferred position of the flyout relative to its target. Defaults to
+    /// <see cref="FlyoutPlacementMode.Auto"/>, which means "no opinion — let WinUI decide";
+    /// Reactor deliberately leaves <c>FlyoutBase.Placement</c> untouched in that case
+    /// because WinUI's own show-time validator rejects <c>Auto</c>.
+    /// </summary>
     public FlyoutPlacementMode Placement { get; init; } = FlyoutPlacementMode.Auto;
     internal Action<WinUI.CommandBarFlyout>[] Setters { get; init; } = [];
 }
--- a/src/Reactor/Core/Element.cs
+++ b/src/Reactor/Core/Element.cs
@@ -6713,15 +6713,15 @@ public record CommandBarFlyoutElement(
     public bool IsOpen { get; init; }
     /// <summary>
     /// Preferred position of the flyout relative to its target. Defaults to
-    /// <see cref="FlyoutPlacementMode.Auto"/>.
+    /// <see cref="FlyoutPlacementMode.Auto"/>, meaning "no opinion — let the platform decide".
     /// </summary>
     /// <remarks>
-    /// <c>Auto</c> is currently written through to <c>CommandBarFlyout.Placement</c> rather
-    /// than being guarded like the other flyout elements. That is a merge-boundary artifact,
-    /// not a statement that <c>CommandBarFlyout</c> tolerates <c>Auto</c> — it does not. The
-    /// value simply never reached WinUI's show-time validator, because the flyout was
-    /// installed as <c>AttachedFlyout</c> metadata that nothing ever opened. The companion
-    /// change that fixes that wiring also guards these sites.
+    /// These sites are guarded by <c>Reconciler.ApplyFlyoutPlacement</c> rather than by
+    /// <c>FlyoutPlacement.Apply</c>, because they are owned by the change that fixed
+    /// <c>CommandBarFlyout</c> never opening from its target. Both prevent <c>Auto</c> from
+    /// reaching WinUI's show-time validator, which rejects it; they differ only on an update
+    /// back to <c>Auto</c>, where this one clears the DP so it returns to the platform
+    /// default rather than retaining the last explicit value.
     /// </remarks>
     public FlyoutPlacementMode Placement { get; init; } = FlyoutPlacementMode.Auto;
     internal Action<WinUI.CommandBarFlyout>[] Setters { get; init; } = [];
--- a/src/Reactor/Core/Element.cs
+++ b/src/Reactor/Core/Element.cs
@@ -5273,9 +5273,13 @@ public record FlyoutElement(
     public bool IsOpen { get; init; }
     /// <summary>
     /// Preferred position of the flyout relative to its target. Defaults to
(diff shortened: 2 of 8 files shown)


azchohfi:
### ⚠ Merge-time conflict with the sibling flyout PR — please read before merging either

**#939 and #943 make contradictory assertions about `CommandBarFlyout` placement.** Whichever merges second will go red, and **the change that would fail is the correct one**.

| | #939 | #943 |
|---|---|---|
| `CommandBarFlyout` + `Auto` | written through, guard deliberately excluded | never written; DP cleared via `ClearValue` |
| pinned by | `CommandBarFlyout_Is_Deliberately_Left_Unguarded` (unit), `CommandBarFlyout_KeepsAuto_Unguarded` (selftest, + 2 registry sites) | `CbfAuto_PlacementNotAuto` |

**The two PRs' evidence is not actually contradictory — they measure different things:**

- #939's fixture proves WinUI **accepts** `Auto` on the `CommandBarFlyout` DP. True.
- #943 proves that **opening** a flyout whose `Placement` is `Auto` fail-fasts in `ShowAtCore` — reproduced live: APPCRASH `0xc000027b`, WER `P7=80070057` (E_INVALIDARG), faulting module `Microsoft.UI.Xaml.dll`. Also true.

**#939's fixture reads the DP after mount and never opens the flyout**, so it cannot support the invariant it asserts. On `main` a `CommandBarFlyout` never opens at all (that is the bug #943 fixes), which is why `Auto` was never validated and why it looked tolerated.

**Falsifiable in one line:** add a call to `CommandBarFlyout_KeepsAuto_Unguarded` that actually *opens* the flyout. If `Auto` were resolved it would open; it fail-fasts.

**Suggested resolution:** drop the `CommandBarFlyout` `Auto` assertions from #939 and keep #943's guard. Please do **not** resolve it by removing #943's guard or inverting `CbfAuto_PlacementNotAuto` — that would make a guard pass by weakening it.

*Context: this conflict was caused by contradictory guidance from the coordinating session, not by either PR's author. #939's own plan had flagged exactly this risk before being talked out of it.*

---

**Added — the cleanest refutation of the "you would lose auto-placement" objection**, from #943's session:

> You are correct that an *unset* `CommandBarFlyout` placement resolves to `Top`. **That is exactly what the guard produces** — it uses `ClearValue`, which restores the DP default. "Cleared" and "unset" are the same state.
>
> The step that does not follow is "therefore remove the guard". Reactor's **element** default is `Auto` (13), and pre-fix Reactor *wrote 13 onto the DP*. So the DP was never unset — it held the one value `ValidateAndSetParameters` rejects. **Removing the guard restores that write. There is no working auto-placement to lose, because `Top` is what unset means here.**

Measured, with the value recorded in the assertion names:

```
ok     CbfAuto_PlacementNotAuto[Top]      <- guard present: flyout opens
ok     CbfReset_DefaultIsNotAuto[Top]     <- new CommandBarFlyout().Placement
not ok CbfAuto_PlacementNotAuto[Auto]     <- guard removed
not ok CbfAuto_OpenedWithoutFailFast      <- ...and it does not open
```

So #939's `FlyoutPlacement_CommandBarFlyout_KeepsAuto` asserts `Placement == Auto` — pinning the DP to 13, which is the *opposite* of the unset/`Top` state that is actually correct.

---

### ⚠ Correction to the resolution above — "delete the assertions" is too broad

Verified in the fixture: `CommandBarFlyout_KeepsAuto_Unguarded` contains **four** checks, and only **one** conflicts.

| Line | Check | After #943 |
|---|---|---|
| `:347` | `FlyoutPlacement_CommandBarFlyout_TargetMounted` | unaffected — **keep** |
| `:350` | `FlyoutPlacement_CommandBarFlyout_MountAttached` | unaffected — **keep** (its helper already reads `button.Flyout ?? GetAttachedFlyout(button)`, so it survives the slot move) |
| `:351` | `FlyoutPlacement_CommandBarFlyout_KeepsAuto` | **the only conflict — invert to `!= Auto`, or drop** |
| `:354` | `FlyoutPlacement_CommandBarFlyout_ExplicitStillApplies` | still true — **keep**; explicit placement lands verbatim under #943, independently proven by its `CbfReset_ExplicitPlacementApplied` |

**Minimal correct reconciliation:**
1. Delete the unit test `CommandBarFlyout_Is_Deliberately_Left_Unguarded()` in `tests/Reactor.Tests/FlyoutPlacementGuardTests.cs`.
2. Invert **or drop only** the `KeepsAuto` check at `FlyoutPlacementFixtures.cs:351`.
3. **Leave the fixture and both `SelfTestFixtureRegistry.cs` entries alone.**

Deleting the whole fixture would discard three valid, passing assertions *and* require the repo's two-place registration removal (`AllFixtures` **and** the `Create()` switch) — easy to half-do, and it fails confusingly when you do.

---

## ✅ RESOLVED — no action needed at merge time

Verified on **#939 head `277c73cc`**: the `CommandBarFlyout_KeepsAuto` and
`CommandBarFlyout_Is_Deliberately_Left_Unguarded` assertions have been **removed**, and **#943's guard is intact** at `14e5fc6f`.

Both PRs are `MERGEABLE` with zero failing checks and zero unanswered comments. **These two no longer conflict and can be merged in either order.**

Leaving the analysis above as the record of what the conflict was and how it was settled.

azchohfi:
Confirming resolved, with two updates since your verification — the branch has moved twice and both changes strengthen the conclusion.

**Verified at head `3f2ef771`** (you checked `277c73cc`):

1. **Rebased onto merged #943.** The rebase conflicted in exactly the partitioned places — `OverlayLifecycle.cs` ×3 and `Element.cs` — all resolved in #943's favour. Those three `CommandBarFlyout` sites are byte-identical to `main`, so #943's guard is untouched.
2. **Converged the two helpers on `ClearValue`.** They previously differed on an `explicit → Auto` update: `ApplyFlyoutPlacement` cleared the DP, `FlyoutPlacement.Apply` skipped the write. #952 turns that from tidiness into correctness — skipping leaves the previous placement as a **local** value, which outranks a `Style` setter, so an app returning to `Auto` could never recover its styled placement. `FlyoutPlacement.Apply` now clears too. Tracked as #953, whose behavioural half this implements; only the structural de-duplication remains.

**On your "delete the assertions is too broad" correction — I did remove the whole fixture, deliberately.** Your reasoning was right at the time; what changed is that #943 has since landed and covers all three checks you wanted preserved, with stronger versions:

| Check you flagged to keep | Covered by #943 |
|---|---|
| `TargetMounted` | `CbfWire_TargetMounted` |
| `MountAttached` | `CbfWire_FlyoutInButtonSlot` (and `CbfWire_MountNotAttached`) |
| `ExplicitStillApplies` | `CbfReset_ExplicitPlacementApplied` |

Keeping mine would have duplicated another change's territory with weaker assertions — mine read the DP after mount, #943's actually open the flyout, which is the distinction your comment is built on. **And I did the two-place removal properly**, since you flagged it as easy to half-do: grepping `SelfTestFixtureRegistry.cs` for the fixture name returns nothing from either `AllFixtures` or the `Create()` switch.

**On the falsifiable check you proposed** — I ran it rather than taking it on report, with a control to prove the harness works:

```
# probe[Top]  ShowAt returned, IsOpen=True          <- control: valid placement opens
# probe[Auto] THREW ArgumentException: Value does not fall within the expected range
```

Same exception as the `Flyout` crash this PR fixes. It confirms your framing exactly: the DP *accepts* `Auto`, and *showing* it throws — two different claims, only the second bearing on the guard. That probe is not checked in, since asserting "this throws" became wrong the moment #943 landed.

Current state: `MERGEABLE`/`CLEAN`, 18 checks passing / 2 skipped, 0 unanswered comments, 0 behind `main`.

One thing worth your attention separately: `Issue487_SV2ParkedAtBottom` and `Issue487_SV2ClampObserved` fail **consistently on a clean `main` worktree** with no PR changes present — not a flake, and unowned as far as I can tell.

Issue #919, opened 2026-07-28
[Bug] Got error message when click expand item in Reactor.TestApp


nufikq:
### What happened?

I've put the error log and step to reproduce to make it clear. I've got error message when click expand item in Reactor.TestApp. I'm sorry if this description is still unclear. I don't know what to say more.

### Steps to reproduce

1. Build as said in [docs](https://microsoft.github.io/microsoft-ui-reactor/getting-started/)

```
git clone https://github.com/microsoft/microsoft-ui-reactor.git
cd microsoft-ui-reactor
./bootstrap.ps1

dotnet new reactorapp -n MyApp
cd MyApp
dotnet run
```

2. Close Visual Studio then open file project `Reactor.slnx`
3. Run project `WordPuzzle (samples\apps\wordpuzzle\WordPuzzle)`, fix bug with GPT 5.4 mini, then play it until congratulation message.
4. Run project `Reactor.TestApp`. Select `DataGrid`, Demo: `Advanced Features (Phase 3-5)`, then try click one of the triangle for expand item in the list.
5. You'll get the error.

### Reactor version / commit

0.1.0-preview.12

### Platform

x64

### .NET SDK version

10.0.302

### Windows version

Windows 11 25H2 10.0.26200 Build 26200

### Windows App SDK version

2.1.3

### Logs / stack trace

```text
System.InvalidCastException: Unable to cast object of type 'Microsoft.UI.Xaml.Controls.Grid' to type 'Microsoft.UI.Reactor.Layout.FlexPanel'.
   at Microsoft.UI.Reactor.Core.V1Protocol.V1HandlerAdapter`2.Update(Element oldEl, Element newEl, UIElement control, Action requestRerender, Reconciler reconciler) in C:\Users\MSI\source\repos\microsoft-ui-reactor\src\Reactor\Core\V1Protocol\V1HandlerAdapter.cs:line 62
   at Microsoft.UI.Reactor.Core.Reconciler.Update(Element oldEl, Element newEl, UIElement control, Action requestRerender) in C:\Users\MSI\source\repos\microsoft-ui-reactor\src\Reactor\Core\Reconciler.Update.cs:line 186
   at Microsoft.UI.Reactor.Core.Reconciler.ReconcileImperative(Element oldElement, Element newElement, UIElement existingControl, Action requestRerender) in C:\Users\MSI\source\repos\microsoft-ui-reactor\src\Reactor\Core\Reconciler.cs:line 1567
[...]

PR #940, opened 2026-07-29
Document and test the edge-triggered contract for native-mutated declared bools (R7)

azchohfi:
## TL;DR

**R7 as written is overstated — I could not reproduce it, so this PR ships no behaviour change.**
A natively-dismissed `InfoBar` *does* recover: a declared `false → true` transition re-opens it. What
was actually missing is a **stated rule** and a **test that pins it**. This PR adds both, plus an
audit of a real inconsistency the investigation uncovered.

Rebased onto `main` after #930 merged. **The diff is purely additive — 512 insertions, 0 deletions.**
No existing assertion, allowlist or guard was touched, weakened, or relaxed.

---

## Symptom as reported

> R7 — `InfoBar.IsOpen` drifts after a native close and never recovers
>
> The user closes an `InfoBar` with its built-in ✕. WinUI sets `IsOpen = false` on the control.
> Reactor's declared `IsOpen` is still `true`, so on the next render old == new and the updater
> skips the write — **the bar can never be shown again without remounting.**

The ask was to treat `InfoBar.IsOpen` as a **controlled** property: when the declared value is `true`
and the live control has drifted to `false`, re-assert it on the next render.

## Root cause — accurate version

`InfoBarElement.IsOpen` is a generator-auto-mapped `OneWay` prop. `OneWayPropEntry.Update`
diffs **old element vs new element**, never the live control:

```csharp
var nv = _get(newEl);
if (!_comparer.Equals(_get(oldEl), nv)) _set(ctrl, nv);
```

That much is true. But the conclusion drawn from it is not — with declared `false → true` the
comparison is **unequal**, so it **does** write. I confirmed the code path is unobstructed:

- `InfoBarElement` has no arm in `Element.ShallowEquals`'s switch (falls to `_ => false`), so
  `Reconciler.Update`'s shallow-skip arm never fires for it.
- `DescriptorHandler.Update` has no entry-level skip; `props[i].Update(...)` always runs.

## Evidence — I ran this, against a real WinUI `InfoBar`

Throwaway selftest fixture, since deleted. Stable across 4 runs:

| Step | declared | live after | `OnClosed` count |
|---|---|---|---|
| A mount | `true` | `True` | 0 |
| B native ✕ (`bar.IsOpen = false`) | `true` | `False` | **1** |
| C re-render, **same** declared | `true` | `False` | 1 |
| D `true → false` | `false` | `False` | 1 |
| E `false → true` | `true` | **`True`** | 1 |

Row **E** is the recovery R7 says is impossible. Row **C** — the only non-recovering case — is the
*defined* behaviour of an edge-triggered value, not a bug.

## The rule this PR writes down

> A declared `bool` that the native control can also mutate is **edge-triggered**: the element
> declares a **transition**, not a mirror. Reactor writes the control only when the declared value
> *changes*. Wire the control's change callback (`OnClosed` / `OnPaneOpenChanged`) to keep app state
> in sync with a native dismissal.

## Why I did *not* implement the requested fix

I planned it (a `ReassertingOneWayPropEntry` + a public `ControlDescriptor.ReassertingOneWay(...)`),
then withdrew it. `InfoBarElement.IsOpen`, `SplitViewElement.IsPaneOpen` and
`NavigationViewElement.IsPaneOpen` **all default to `true`**. Live re-assert would make every one of
them written without a change callback **undismissable** — the user dismisses, and the next unrelated
`setState` anywhere in that component brings it back. That is a worse, more visible regression than
the reported bug, and it would land on every existing caller.

"It's the React controlled-input contract" is a fair principle, but it does not survive a default of
`true`.

## What changed

| File | Change |
|---|---|
| `src/Reactor/Core/Element.cs` | XML doc on `InfoBarElement.IsOpen`, `TeachingTipElement.IsOpen`, `SplitViewElement.IsPaneOpen`, `NavigationViewElement.IsPaneOpen` (+ `InfoBarElement.IsClosable`). **Doc comments only.** |
| `docs/_pipeline/templates/extending-reactor-controls.md.dt` | New *"Values the native control mutates on its own"* section |
| `docs/guide/extending-reactor-controls.md` | Regenerated via `mur docs compile --pages extending-reactor-controls` |
| `tests/Reactor.AppTests.Host/SelfTest/Fixtures/IsOpenEdgeTriggeredFixtures.cs` | New selftests against real `InfoBar` / `TeachingTip` |
| `tests/Reactor.AppTests.Host/SelfTest/SelfTestFixtureRegistry.cs` | Registration (both `AllFixtures` and the `Create()` switch) |
| `tests/Reactor.Tests/InfoBarIsOpenEdgeTriggeredTests.cs` | New headless descriptor-wiring guards |

`mur --regen-api` reports `reactor.api.txt unchanged` for **both** `skills/reactor.api.txt` and its
`plugins/reactor/...` mirror — no public API surface change.

## Tests

The two failure modes point in opposite directions, so both halves are pinned:

- **`RisingEdgeReopens`** fails if the rising-edge write is ever lost — the recovery path.
- **`SameDeclaredValueDoesNotReopen`** fails if the engine is ever changed to re-assert against the
  live control — a standing guard against exactly the regression described above.

### Proof they are not vacuous — verbatim output

There is no product change here, so the usual "revert the fix, watch the test fail" does not apply.
I mutation-tested instead. All three mutations re-run **after the rebase**; all reverted afterwards
(`git diff HEAD -- src/` is empty).

**Baseline (unmutated):** `# Total failures: 0` — 9 InfoBar + 5 TeachingTip assertions pass.

**MUTATION-1** — `OneWayPropEntry.Update` never writes:

```
not ok IsOpenEdge_InfoBar_RisingEdgeReopens - assertion failed
not ok IsOpenEdge_InfoBar_ProgrammaticCloseClosesOpenControl - assertion failed
not ok IsOpenEdge_TeachingTip_RisingEdgeOpens - assertion failed
not ok IsOpenEdge_TeachingTip_RisingEdgeReopensAfterDismiss - assertion failed
# Total failures: 4
```

**MUTATION-2** — `OneWayPropEntry.Update` always writes (i.e. the re-assert "fix" R7 asked for):

```
not ok IsOpenEdge_InfoBar_SameDeclaredValueDoesNotReopen - assertion failed
not ok IsOpenEdge_InfoBar_FallingEdgeOnClosedControlRaisesNoCallback - assertion failed
not ok IsOpenEdge_InfoBar_RisingEdgeRaisesNoCallback - assertion failed
not ok IsOpenEdge_TeachingTip_SameDeclaredValueDoesNotReopen - assertion failed
# Total failures: 4
```

**MUTATION-3** — `InfoBarElement.IsOpen` default flipped to `false`:

```
InfoBarIsOpenEdgeTriggeredTests.IsOpen_IsBoundThroughAPlainOneWayEntry [FAIL]
InfoBarIsOpenEdgeTriggeredTests.IsOpenEntry_CannotObserveTheLiveControl [FAIL]
InfoBarIsOpenEdgeTriggeredTests.IsOpen_DefaultsToTrue [FAIL]
Failed!  - Failed: 3, Passed: 0, Skipped: 0, Total: 3
```

The two mutations fail **disjoint** assertion sets, which is the point: neither direction of
regression can slip through.

> **The "always writes" mutation earned its keep:** it caught a vacuity bug in my own first draft. I
> had passed the *same element instance* as both old and new to `rec.UpdateChild`, and
> `Element.ShallowEquals` opens with `if (ReferenceEquals(a, b)) return true;` — so the entire update
> was skipped and the assertions passed for the wrong reason. Fixed by constructing a fresh element
> per simulated render, which is what a real render loop does.

## Audit — Reactor has three different rules here, not one

Found while verifying the above. **Reported, not fixed.**

| Prop | Element default | Update rule | Where |
|---|---|---|---|
| `Expander.IsExpanded` | `Optional<bool>` **unset** | live re-assert | `ExpanderHandler.cs:50` |
| `Popup.IsOpen` | `false` | live re-assert | `OverlayLifecycle.cs:401` |
| `InfoBar.IsOpen` | **`true`** | edge, both directions | auto `OneWay` |
| `TeachingTip.IsOpen` | `false` | edge, both directions | auto `OneWay` |
| `SplitView.IsPaneOpen` | **`true`** | edge, both directions | auto `OneWay` |
| `NavigationView.IsPaneOpen` | **`true`** | edge, both directions | auto `OneWay` |
| `ContentDialog.IsOpen` | `false` | **rising edge only — no falling edge** | `OverlayLifecycle.cs:48` |
| `Flyout.IsOpen` | `false` | **rising edge only — no falling edge** | `OverlayLifecycle.cs:215` |

The two re-asserting rows are defensible and the docs now say why: `Expander` is `Optional<bool>` so
authority is opt-in, and `Popup` defaults to `false`, which is also where a light-dismissed popup
rests — so its re-assert is a no-op in the direction that would fight the user.

**Two genuine defects fall out of this table, both left for separate PRs:**

1. **`ContentDialog.IsOpen` / `Flyout.IsOpen` have no falling edge.** Declaring `IsOpen = false`
   does not close them — the update only handles `n.IsOpen && !o.IsOpen`.
2. **A Reactor-mounted `TeachingTip` never actually presents.** Declaring `IsOpen: true` on the
   first render leaves it closed, and even after a rising edge that flips the property, WinUI raises
   **no `Closed` event** on the way back down — while an identically-parented raw
   `WinUI.TeachingTip` does present and does raise `Closed`. This likely affects the gallery's
   TeachingTip page. This is also why `TeachingTipEdgeTriggered` has no `OnClosed` oracle: a callback
   count there would assert the absence of that presentation defect rather than the edge contract.
   The fixture's doc comment records this.

## Note for the gallery fix (sibling session)

The gallery's *Reset* button needs the `OnClosed` wiring **and** something more:
`setShowClosable(true)` when the value is already `true` never re-renders at all —
`RenderContext.MakeStateSetter` bails on an equal value. The framework change alone would not have
fixed that button.

## Validation

Re-run after rebasing onto `main` (post-#930):

- `dotnet build Reactor.slnx` — 0 errors
- `dotnet test tests/Reactor.Tests` — `Failed: 0, Passed: 12527, Skipped: 64, Total: 12591`
- targeted selftests — `# Total failures: 0`
- `dotnet test tests/Reactor.SelfTests` — 2 failures, **identical to the `origin/main` baseline**
  measured on the same machine by reverting to a clean tree (`BackdropTransparent_Apply` + a fixture
  timeout); `BackdropTransparent_Apply` passes in isolation, so it is order/timing dependent and
  pre-existing, not from this branch
- `mur docs compile --validate-only` — validation passed; unrelated drifted snippet churn in
  `docs/guide/architecture-overview.md` (regenerated from PR #919's source) was reverted so it is
  not in this diff
- `.github/skills/pr-review` across all 8 dimensions on the final commit: 2 findings, both fixed
  (commits 2 and 3). Multi-model cross-check (GPT-5.4, high reasoning) returned 0 findings and
  independently confirmed the core judgement.

### One red `Selftests` run in this PR's history — proven flake

Commit `273864e3` initially failed `Selftests` on `Issue717_PinReleaseDoesNotClobberAuthorMinHeight`.
That is not one of this PR's fixtures. Four independent pieces of evidence, rather than an assertion:

1. `Selftests` was **green on the immediately preceding head** `4f1eee88`.
2. The **only** difference between those two commits is
   `tests/Reactor.Tests/InfoBarIsOpenEdgeTriggeredTests.cs` — a headless xUnit file that
   `Reactor.AppTests.Host.csproj` does not reference, so the selftest host never compiles it.
3. The fixture passes **3/3 locally** on the failing commit.
4. **Re-running the job unchanged on the identical commit went green.**

## Merge-ordering

No constraint. This PR adds no analyzer rule and changes no behaviour, so it cannot trip the
Release `TreatWarningsAsErrors` gate that #932 introduces.

## Open question for the reviewer

If you'd rather have the behaviour change anyway, the safe version is **not** live re-assert — it is
migrating these four props to `Optional<bool>` so authority is opt-in, which is a breaking change to
their defaults and deserves its own spec.


azchohfi:
### Addressing the suppressed low-confidence review comment (`InfoBarIsOpenEdgeTriggeredTests.cs:142`)

Copilot's review body reported *"generated no new comments"* but carried a collapsed
**Comments suppressed due to low confidence (1)** block, which does not appear in the
`pulls/940/comments` API. Replying here since a suppressed block cannot be replied to inline.

> `GetPrivateField()` only searches the concrete runtime type. If the descriptor entry
> implementation later moves a private field (like `_get`) into a base type, this helper
> returns null and the "locates by behaviour" test will start failing for a refactor that
> doesn't change the contract.

**Taken — it was right, and it matters more here than usual.** This PR's whole argument is
that you pin semantics so a future refactor cannot silently break them. A guard that goes
red on a *behaviour-preserving* refactor is the opposite failure: it cries wolf, and
crying-wolf guards are how allowlists and assertions get weakened.

`GetField(name, Instance | NonPublic)` does not walk the hierarchy, and
`BindingFlags.FlattenHierarchy` does not help — it is documented not to apply to private
fields — so an explicit `BaseType` walk is the reliable form. Now consistent with
`InstanceFieldNames()`, which I fixed for the mirror-image reason in `273864e3` (that one
could have gone *silently green*; this one would have gone *falsely red*).

**The widening cannot mask a real regression.** The entry is still identified by
**behaviour** — it must read `IsOpen` and not `IsClosable` — and a miss returns `null`,
so `ReadsIsOpen` returns `false` and the `Assert.Single` in `FindIsOpenEntry` still fails
loudly. Fail-loud, not fail-silent.

Added `ReflectionHelpers_FindPrivateFieldsDeclaredOnABaseType` so the hierarchy walk cannot
be quietly reverted, using a local type whose private field is declared on the base — the
exact shape a future descriptor refactor would produce.

**Mutation-checked, per the "make sure it still fails" bar:**

| Mutation | Result |
|---|---|
| `GetPrivateField` reverted to single-level | **only** `ReflectionHelpers_FindPrivateFieldsDeclaredOnABaseType` fails |
| `InfoBarElement.IsOpen` default flipped to `false` | the **3 contract guards** still fail; the helper test correctly still passes |

Full unit suite green: `Failed: 0, Passed: 12528`.

PR #978, opened 2026-07-31
Fix TeachingTip declared IsOpen: true on first render never opening (#949)

### LINKED ISSUE #949: [Bug] TeachingTip declared IsOpen: true on its first render never opens (mount-path only; click-driven open works)
### What happened?

A `TeachingTip` whose **first render** declares `IsOpen: true` never opens. The mount-time write *is* issued and then silently lost.

> **⚠️ Scope, so this isn't closed as non-reproducible in thirty seconds:** a **click-driven `false → true` open works correctly**. The gallery's TeachingTip page presents fine when you click its button — verified independently with a real click (`TitleTextBlock`, `IsOffscreen=False`). **This issue is about the mount path only.**

### The reconciliation — three observations that look contradictory but aren't

| Path | Behaviour | Cause |
|---|---|---|
| Mount with declared `IsOpen: true` | **does not present** | **this issue** |
| `false → true` after mount (e.g. a button) | presents correctly | working as intended |
| Gallery *"(Title Only)"* card | can never present | separate gallery bug — `TeachingTipPage.cs:39-41` declares a tip with no `IsOpen` and no trigger |

### ⚠️ If you are here to fix the gallery's "(Title Only)" TeachingTip card — read this first

`TeachingTipPage.cs:39-41` declares a tip with no `IsOpen` and no trigger, so it never appears. **The obvious one-line fix — adding `IsOpen = true` — lands straight on this bug and still will not present.** You will apply it, see nothing, and most likely conclude the card is broken some other way.

The correct fix is the shape the working "Basic TeachingTip" card on the same page already uses: start closed and toggle after mount.

```csharp
var (show, setShow) = UseState(false);          // starts CLOSED — this is the load-bearing part
...
Button("Show tip", () => setShow(true)),
TeachingTip("Did you know?", "…") with { IsOpen = show, OnClosed = () => setShow(false) }
```
### ✋ It is *not* "edge-triggered semantics have no edge at mount"

That was the first explanation offered and it is wrong — recording it here so this doesn't get triaged as by-design:

- `OneWayPropEntry.Mount` writes **unconditionally** (`_set(ctrl, _get(el))`). There is no diff at mount, so the declared value is always applied.
- **`InfoBar` mounts open correctly** under the identical semantics — `IsOpenEdge_InfoBar_MountAppliesDeclaredOpen` (added in #940) passes. If "no edge at mount" were the cause, InfoBar would fail too.
- A descriptor entry bisect shows the write landing and *then* being lost:

```
Bisect_08_OneWayPropEntry[…, Boolean]                   False -> True     ← IsOpen written, lands
Bisect_09_OneWayPropEntry[…, TeachingTipPlacementMode]  True  -> False    ← gone
Bisect_finalLive=False
```

### Root cause — pinned

**On an *unparented* `TeachingTip`, any subsequent property write after `IsOpen = true` drops the pending open. Parented, it is harmless.** Reactor's `Mount` configures the control before it is parented, and writes `IsOpen` *before* its remaining property entries — so a mount-time open is always discarded.

Isolated against raw `WinUI.TeachingTip`, no Reactor involved:

| Parented when opened? | Subsequent property write? | Result |
|---|---|---|
| no | **no** | opens ✔ |
| no | **yes** | **lost** ✘ |
| yes | yes | opens ✔ |

```
ProbeN14_MountTimeOpen__preParent=False__afterParentAndDwell=False                ← Reactor
ProbeN14_RawMountTimeOpen__preParent=True__afterParentAndDwell=True               ← raw, same configure-then-parent order
ProbeN14_RawUnparented_OpenThenWrite__afterOpen=True__afterWrite=False__afterParentAndDwell=False
ProbeN14_RawParented_OpenThenWrite__afterParentAndDwell=True
```

Raw WinUI with the *same* configure-then-parent order works, so this is Reactor-side: it is the extra property writes that follow `IsOpen` while the control is still unparented. `PreferredPlacement` is not special — it is simply *a subsequent write*.

### Steps to reproduce

```csharp
// First render declares the tip open. It never appears.
return VStack(
    TextBlock("hello"),
    TeachingTip("Welcome!", "Guidance text.") with { IsOpen = true });
```

Contrast — this works:

```csharp
var (show, setShow) = UseState(false);   // starts closed
return VStack(
    Button("Show", () => setShow(true)),          // rising edge — presents correctly
    TeachingTip("Welcome!", "Guidance.") with { IsOpen = show, OnClosed = () => setShow(false) });
```

Selftest form (what I actually ran):

1. `rec.Mount(new TeachingTipElement("t") { IsOpen = true })`
2. Add the returned control to a parented `Grid`, `await Harness.Render(300)` twice
3. Read `tip.IsOpen` → **`False`**
4. Repeat with a bare `new WinUI.TeachingTip { Title = "t", IsOpen = true }` → **`True`**

### Suggested fix direction

Write `IsOpen` **last** in the TeachingTip mount sequence, or defer it until after the control is parented (a `Loaded` hook). Confirm no echo loop with `Closed` / `OnClosed`.

### Relationship to #948

Same family — both are gaps in getting a declared value onto the control. #948 is a **falling edge that is never issued**; this is a **mount write that is issued and then dropped**. Together with the eight-property audit in #940 they make the case for standardising the native-mutable declared bools.

### Correction to an earlier version of this report

I first reported this as *"a Reactor-mounted TeachingTip never presents"*, including the rising-edge path, on the basis that no `Closed` event fired on the way down. **That was my own test error and is withdrawn.** My diagnostic opened the tip via a predicate that returned as soon as the property read `true` — i.e. with essentially zero message pumping — then closed it immediately, while the raw control it was compared against got an explicit 300 ms dwell. With the dwell equalised the Reactor rising edge presents and fires `Closed` (`ReactorRisingEdge_WITH_dwell__open=True__closedEvents=1`), and the zero-dwell control reproduces the original artifact exactly. Only the mount-path half survives.

### Reactor version / commit

184bf5cdd79b807f21d69297e4b8a90ea7356d64

### Platform

x64

### .NET SDK version

10.0.302

### Windows version

Windows 11 Enterprise Insider Preview build 26310

### Windows App SDK version

2.1.3 (WinUI 2.1.0)

### Logs / stack trace

N/A — no exception. `TeachingTip.IsOpen` simply reads `false` after mount.

### Confirmation

- [x] This bug reproduces against the current `main` branch.

azchohfi:
Fixes #949.

## The bug

A `TeachingTip` whose **first render** declares `IsOpen: true` never opened. The mount-time write *was* issued — `OneWayPropEntry.Mount` writes unconditionally — and then silently lost.

Reactor, like XAML, fully configures a control **before** handing it to its parent:

```
DescriptorHandler.Mount   prop loop (IsOpen written here), EnsureSubscribed, ApplySetters
V1HandlerAdapter.Mount    children strategy (Content / HeroContent), AfterChildrenMount
Reconciler.Mount          ApplyModifiers (Margin, ...)
   ...the caller then adds the control to its parent   <-- parenting is LAST
```

On an **unparented** `TeachingTip`, WinUI discards a pending open as soon as anything else is written to the control. Parented, it's harmless. So the declared open was always thrown away by a later write. I confirmed by measurement that `PreferredPlacement` is the write that discards it — matching the descriptor bisect in the issue, and correcting its "PreferredPlacement is not special" note for the post-mount case.

Ordering the entry last inside the descriptor is **not** sufficient: setters, the content slots and common modifiers all run outside the descriptor's reach.

## The fix

1. `[WrapManual("IsOpen")]` takes `IsOpen` off the generator's auto-map.
2. A bespoke `.Imperative` entry defers a declared `true` to the control's `Loaded` event — the first moment the tip is parented into a live tree and an open is guaranteed to stick. A declared `false` writes directly and disarms any pending open.
3. Updates stay **edge-triggered**, so the contract locked in by #940 holds: a re-render carrying the same declared `true` still does not re-assert against a natively dismissed tip.

Arm state lives on the control's existing `TeachingTipEventPayload` via the `ReactorState` attached-DP store — the documented home for per-element state. A side `ConditionalWeakTable` would be keyed per RCW, so two RCWs over the same native tip would see different arms.

### New: `ControlDescriptor.OnUnmount` / `.WithUnmount(...)`

An arm could otherwise outlive the mount that created it: unmount before `Loaded`, re-parent, and the tip popped open on behalf of a mount that no longer existed. The engine already dispatched unmount to hand-coded handlers, but `DescriptorHandler` never forwarded it — descriptors had **no teardown seam at all**.

Because that dispatch is tag-gated, declaring the hook also makes `DescriptorHandler` force the control's `ReactorState` at mount. Without it the hook would fire for callback-bearing elements of a type and silently not for callback-free ones.

## One behavior note for authors

A `.Set(t => t.IsOpen = false)` can no longer override a declared mount-time `true`, because the deferred write lands after the setter pass. Documented on `TeachingTipElement.IsOpen`, in the guide, and pinned by a test so it can't drift.

## Also fixed

The gallery's **"TeachingTip (Title Only)"** card declared a tip with no `IsOpen` and no trigger, so it could never appear. The issue warns that the obvious one-line `IsOpen = true` fix lands straight on this bug; now that mount-open works it would present once and — being edge-triggered with nothing wired to `OnClosed` — never come back. It now uses the state-driven shape its sibling card already used.

## Testing

New selftest fixture `IsOpenEdge_TeachingTipMountOpen` (10 checks) plus headless `DescriptorUnmountHookTests`. Every assertion was mutation-checked — none can pass vacuously:

| Probe | Fails |
|---|---|
| bypass the deferral | `DeclaredOpenOnFirstRenderOpens`, `PresentsAPopup`, `PresentedTipRaisesClosedOnTeardown`, `RisingEdgeBeforeLoadOpensOnLoad`, `DeclaredOpenWinsOverSetter`, `ReloadDoesNotReopenAfterNativeDismiss` |
| `Cancel` no-op | `FallingEdgeBeforeLoadCancelsPendingOpen` |
| mount always arms | `DeclaredClosedStaysClosed` |
| unmount hook disabled | `UnmountBeforeLoadCancelsPendingOpen` (selftest) + `DescriptorHandler_Unmount_...` (unit) |

`PresentedTipRaisesClosedOnTeardown` is a real presentation oracle: WinUI raises `Closed` only for a tip that genuinely presented, so a "set the DP and hope" fix would not satisfy it. Presentation itself is waited for via a popup appearing on the `XamlRoot`, counted from a pre-open baseline, rather than assuming a fixed dwell.

**13,000 unit tests pass. The full selftest suite (1378 fixtures) passes with zero failures.**

## Review

Ran the repo's `pr-review` skill (8 dimensions + a GPT-5.6 cross-check on a different model family). Security, correctness and packaging came back clean; the other findings are addressed in the second commit. Two are deliberately **not** taken:

- Extracting a shared `Loaded`-deferral helper across the four existing near-duplicates in the repo — real, but a refactor beyond this fix's scope.
- `ReloadDoesNotReopenAfterNativeDismiss` is named for what it actually pins. A probe that disabled the arm's self-cleanup did *not* fail it, because WinUI doesn't re-invoke the spent hook there, so claiming it tested that cleanup would have been false advertising.

## Environment caveats

`mur docs compile` and `mur --regen-api` can't run in this environment — `Reactor.Cli`'s `GitHub.Copilot.SDK.targets` npm download is TLS-blocked. I hand-applied a verified byte-identical docs block (confirmed no template-only constructs would transform) and used the supported in-process `UPDATE_API_INDEX=1` path for the two API index copies. The ReactorGallery search index was regenerated and is unchanged.


azchohfi:
Re: the suppressed review comment on `TeachingTipPage.cs:53` — the displayed snippet still using `show` / `setShow`.

Declining this one deliberately, because the gallery's displayed snippets are **minimal teaching examples, not verbatim copies of the card code**, and that is the established convention on this page and elsewhere.

All three cards on this page already do it, including the two I did not touch:

| Card | Code uses | Snippet shows |
|---|---|---|
| Basic TeachingTip (pre-existing) | `showTip` / `setShowTip`, button `"Show Teaching Tip"` | `show` / `setShow`, button `"Show Tip"` |
| TeachingTip (Title Only) (this PR) | `showTitleOnlyTip` / `setShowTitleOnlyTip` | `show` / `setShow` |
| Targeting another subtree (pre-existing) | `showTargetedTip` / `setShowTargetedTip` | `show` / `setShow` |

`ContentDialogPage.cs` follows the same pattern. The "Basic" snippet also shortens the subtitle text, so the abstraction is intentional rather than drift.

Changing only the card in this PR would make it the one inconsistent card on the page; changing all three is unrelated churn in a bugfix PR. The `pr-review` docs-and-samples pass reached the same conclusion independently ("snippets are simplified/abstracted but consistent with the actual code above them").

Happy to normalise the whole page in a follow-up if the team would rather the snippets be verbatim — that is a page-wide (arguably gallery-wide) editorial decision, not a property of this fix.

azchohfi:
Correcting my earlier reply — I was wrong to decline this, and I have now fixed it in `2d953e8a`.

My reasoning was that all three cards on this page (plus `ContentDialogPage.cs`) use generic `show` / `setShow` in the displayed snippet, so matching only the card I touched would make it the inconsistent one.

That read the convention as settled when it is actively being removed. #974 is landing a CI-gated lint whose rule is *"every lowerCamelCase name a snippet uses must exist in the code beside it"*, and its description names this exact file's `show` / `setShow` as genuine drift it fixes. So the snippet would have failed the new gate, and in the meantime a reader copying it gets a name that does not compile — which is the actual harm the reviewer was pointing at.

The snippet now uses `showTitleOnlyTip` / `setShowTitleOnlyTip` and the real button label. I deliberately did **not** touch the two sibling cards or `ContentDialogPage.cs`: #974 already rewrites 23 cards across 19 pages, and duplicating that here would just create a conflict.

azchohfi:
### Addressed: suppressed finding in review 4831445806 (`IsOpenEdgeTriggeredFixtures.cs` popup baseline)

That review raised, in its body rather than as a thread, that `popupsBeforeOpen` was captured before a render pass, so `parent.XamlRoot` could still be null and the baseline would read `0` regardless of any already-open popups. **Fixed in `3c497a2b`** — with one correction to the finding and one generalisation beyond it.

**The specific exploit path does not reproduce.** I instrumented `OpenPopupCount` at the exact call site and measured the root rather than reasoning about it: `H.SetContent(parent)` populates `XamlRoot` synchronously, so `rootNull=False` at the baseline. The stated sequence (null root at `:317`, first pump at `:321`) is not what the fixture actually does.

**The shape it describes is real anyway, and that is the part worth fixing.** The helper returned `0` for both *"I measured and found none"* and *"I could not measure"*, and those are not the same fact. The direction matters: a `0` baseline makes `OpenPopupCount(tip) > popupsBeforeOpen` **easier** to satisfy, so a leftover popup from an earlier fixture could satisfy it with no tip presented — while the comment directly above claimed immunity to exactly that. A reader trusting the comment would not re-derive it. Being unreachable today is a property of the current call order, not of the helper.

**Fix — convert the class, not the instance.** `OpenPopupCount` now returns `int?` and yields `null` on an unavailable `XamlRoot`; both call sites treat `null` as a failed check, never as a value. A new check `MountOpen_TeachingTip_PopupBaselineIsMeasurable` asserts the baseline was actually taken, so an unmeasurable one fails loudly instead of quietly lowering the bar. A render pass is pumped before the baseline as well — belt and braces, since it is what the finding assumed.

**Mutation-checked, because this is precisely an oracle that could have been passing for a reason other than the one claimed.** With the mount-open deferral bypassed, `MountOpen_TeachingTip_DeclaredOpenOnFirstRenderPresentsAPopup` reddens (6 checks fail in total). It is measuring the fix, not the environment.

This is the same fail-open pattern tracked in #992 — a *could-not-measure* status crossing a boundary that cannot carry it and arriving as a legitimate value.

azchohfi:
### Addressed: suppressed finding in review 4833897099 (`IsOpenEdgeTriggeredFixtures.cs:379`)

Fixed in `1771c378` — and the finding understated itself, so the fix is wider than what was asked.

The review flagged `MountOpen_TeachingTip_ControlMounted` as inconsistent with the surrounding `*_FixtureMounted` guards. Correct. But mapping every guard in the file to the sub-fixture it actually guards shows **two** names failing the same way, not one:

```
MountOpen_TeachingTip_Mounted                  <- MountOpenPresents          (generic)
MountOpen_TeachingTip_ControlMounted           <- MountClosedStaysClosed     (generic)
MountOpen_TeachingTip_CancelFixtureMounted     <- FallingEdgeBeforeLoad...
MountOpen_TeachingTip_UnmountFixtureMounted    <- UnmountBeforeLoad...
MountOpen_TeachingTip_RisingEdgeFixtureMounted <- RisingEdgeBeforeLoad...
MountOpen_TeachingTip_SetterFixtureMounted     <- SetterCannotOverride...
```

Both are now renamed to name their sub-fixture: `_PresentsFixtureMounted` and `_StaysClosedFixtureMounted`.

**Why this isn't a style nit.** These guards sit on a `H.Check(..., false); return;` path, so when one fires the fixture stops and the TAP trailer is the only evidence a triager gets. A bare `_ControlMounted` says a precondition failed but not *which* — and attribution is most of what the name is for. Six guards in one class means the generic ones are actively ambiguous, not merely terse.

**Left alone deliberately:** `IsOpenEdge_InfoBar_Mounted` and `IsOpenEdge_TeachingTip_Mounted`. Each is the sole guard in its own fixture class, so there is nothing for its name to disambiguate. Renaming those would be churn.

Verified after the rename: `IsOpenEdge` selftests 25 ok / 0 not ok, and zero stale references to either old name anywhere in the tree.

PR #1076, opened 2026-08-03
Fix bootstrap CI failure on VS 18.8 runners (#1074)

### LINKED ISSUE #1074: [Bug] bootstrap CI fails on VS 18.8 runners: `devenv /updateconfiguration` hangs and `Process.Kill()` orphans it
### What happened?

The `bootstrap.ps1 (windows-latest)` workflow started failing intermittently on **2026-08-01** and is now failing consistently, with:

```
Reinstall-Vsix.ps1:69  Visual Studio is running (PIDs: 2132). Close it and re-run.
    [warn] VS extension install reported a non-zero exit code (1). The rest of the bootstrap completed; ...
Bootstrap complete.
Exception: bootstrap.ps1 re-run exited 1
```

**Root cause is an environment change, not a repo change.** The runner image label is identical in every run (`windows-2025-vs2026`), but the VS build inside it is rolling from **18.7.11925.98** (instance `ece786fa`) to **18.8.12023.21** (instance `a4357766`). On 18.8, `devenv /updateconfiguration` never completes and always hits our 2-minute timeout:

| Run | VS version | 2-min timeouts | guard hits | Result |
|---|---|---|---|---|
| [30570104936](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/30570104936) Jul-30 `main` | **18.7**.11925.98 | 0 | 0 | pass |
| [30712269799](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/30712269799) Aug-01 18:19 `main` | **18.7**.11925.98 | 0 | 0 | pass |
| [30692473409](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/30692473409) Aug-01 08:48 | **18.8**.12023.21 | 1 | 0 | pass |
| [30695217657](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/30695217657) Aug-01 10:11 | **18.8**.12023.21 | 1 | 4 | **fail** |
| [30849981152](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/30849981152) Aug-03 `main` (dispatch) | **18.8**.12023.21 | 1 | 0 | pass |
| [30842251931](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/30842251931) Aug-03 PR #1072 | **18.8**.12023.21 | 1 | 4 | **fail** |

VS 18.7 → completes in 27–99s, exit 0 (2/2). VS 18.8 → hangs, exit -1 (4/4). The fleet is heterogeneous mid-rollout — Aug 1 has runs landing on both versions hours apart.

Note that **passing on 18.8 is luck, not health**: `main`'s green run hung identically and merely happened to have the orphaned process die before the next guard check. Three distinct defects combine:

**1. `Process.Kill()` orphans the devenv tree — `src/vs-reactor/Reinstall-Vsix.ps1:180`**

```powershell
$proc = [System.Diagnostics.Process]::Start($psi)
if (-not $proc.WaitForExit(120000)) { $proc.Kill(); Write-Warning "/updateconfiguration timed out after 2 min." }
Write-Host ("  /updateconfiguration exit code: {0} ({1:0.0}s)" -f $proc.ExitCode, ($proc.ExitTime - $proc.StartTime).TotalSeconds)
```

`Kill()` with no argument terminates **only that process, not its children**, and returns **without waiting**. `devenv /updateconfiguration` spawns a second `devenv` that survives, then poisons every later `Reinstall-Vsix.ps1` on the machine via the guard at line 67. Wants `Kill($true)` (kill tree) followed by `WaitForExit()`.

**2. The duration is computed on a process that hasn't exited — same line**

The logged duration is `-13430263500.8s` (≈ -425 years) because `ExitTime` is read on a live process and returns an unset default. This absurd value has been printed on every timeout since the rollout began and is effectively a broken instrument sitting right next to the bug it was announcing. Compute it after a confirmed exit, or use a `Stopwatch`.

**3. The warn-and-continue path still fails the job — `bootstrap.ps1:557`**

```powershell
if ($LASTEXITCODE -ne 0) {
    # Don't fail the whole bootstrap — the rest of the install is usable without the VS extension.
    Write-Host "    [warn] VS extension install reported a non-zero exit code ($LASTEXITCODE). ..."
}
```

The comment states the intent, but `Write-Host` doesn't clear `$LASTEXITCODE`, so the `1` leaks past `Bootstrap complete.` to the workflow's `if ($LASTEXITCODE -ne 0) { throw }` (`bootstrap.yml:211`). Also worth noting `[ok] VS extension installed` is printed even when `/updateconfiguration` returned `-1` after a timeout — a success message on a failed operation.

**Ruled out:** `Reinstall-Vsix.ps1` unchanged since 2026-06-08 (#543); `bootstrap.ps1` since Jul 13; `bootstrap.yml` since Jul 28 (checkout bump); `src/vs-reactor` since Jul 13. No repo change in ~3 weeks; failures begin Aug 1.

**Also ruled out:** the `no Reactor.VsExtension folder appeared within 30s` warning that accompanies these runs is *not* related — it also occurs on VS 18.7 and self-heals (later invocations log `Installed v.. at`). Benign async first-install race, called out here only so it isn't mistaken for a second symptom.

**Why this surfaced on a dependency-bump PR:** `Directory.Packages.props` is in `bootstrap.yml`'s trigger `paths` (line 18), so any package bump runs this workflow. #1072 is exposure, not causation — its E2E failure was the separate known foreground flake and went green on re-run.

### Steps to reproduce

1. Land any change touching a `bootstrap.yml` trigger path (e.g. `Directory.Packages.props`) so the bootstrap workflow runs.
2. Draw a runner whose VS instance is `a4357766` / 18.8.12023.21 (check `Target VS:` in the log).
3. Run 1's `devenv /updateconfiguration` hits `WARNING: /updateconfiguration timed out after 2 min.` and orphans a `devenv`.
4. The next `Reinstall-Vsix.ps1` invocation ("Verify mur upgrade is idempotent" or "Re-run bootstrap.ps1 (idempotence check)") trips the guard and exits 1, failing the job.

Reproduced 3/3 on PR #1072 (runs 30842251931, plus two re-runs).

### Reactor version / commit

`a2183058bbf65d14cf30dfbdf3356b8f8c6ef6e4` (`main`)

### Platform

x64

### .NET SDK version

.NET 10 SDK (CI `windows-latest`)

### Windows version

GitHub Actions `windows-2025-vs2026` image; VS Enterprise 2026 18.8.12023.21 (instance `a4357766`)

### Windows App SDK version

N/A — VSIX / bootstrap tooling issue, not a runtime issue.

### Logs / stack trace

```
Run bootstrap.ps1 (non-interactive)
  Target VS: Visual Studio Enterprise 2026 (a4357766, 18.8.12023.21)
  Running 'devenv /updateconfiguration' to force the menu/pkgdef merge synchronously.
  WARNING: /updateconfiguration timed out after 2 min.
    /updateconfiguration exit code: -1 (-13430263500.8s)
      [ok] VS extension installed into Visual Studio Enterprise 2026 — ...
  Bootstrap complete.

Verify mur upgrade is idempotent
  Write-Error: src\vs-reactor\Reinstall-Vsix.ps1:69
    Visual Studio is running (PIDs: 2980). Close it and re-run.

Re-run bootstrap.ps1 (idempotence check)
  Write-Error: src\vs-reactor\Reinstall-Vsix.ps1:69
    Visual Studio is running (PIDs: 2980). Close it and re-run.
      [warn] VS extension install reported a non-zero exit code (1). The rest of the bootstrap completed; ...
  Bootstrap complete.
  Exception: bootstrap.ps1 re-run exited 1
  ##[error]Process completed with exit code 1.
```

### Confirmation

- [x] This bug reproduces against the current `main` branch (the underlying `/updateconfiguration` hang reproduces on `main` on any VS 18.8 runner; whether it fails the job depends on the orphaned-process race).

azchohfi:
Fixes #1074.

## The failure

`bootstrap.ps1 (windows-latest)` started failing when the runner image's Visual Studio rolled 18.7 → 18.8. On 18.8 `devenv /updateconfiguration` never returns. Three defects in our own scripts turned that upstream hang into a red job:

1. **`Process.Kill()` orphaned the devenv tree.** No argument means "kill this process only", and it returns without waiting. The second `devenv` that `/updateconfiguration` spawns survived and tripped the next run's "Visual Studio is running" guard.
2. **The duration was read from `ExitTime` on a live process**, printing `-13430263500.8s` (≈ −425 years) — a broken instrument sitting next to the bug it was announcing.
3. **The warn-and-continue path leaked `$LASTEXITCODE`** past `Bootstrap complete.` into the workflow's `if ($LASTEXITCODE -ne 0) { throw }`.

## One correction to the issue's diagnosis

The issue attributes defect 3 to `Write-Host` not clearing `$LASTEXITCODE`. That's true but insufficient — *how* `bootstrap.ps1` is invoked is load-bearing. `bootstrap.yml` runs it **in-process**, and `$LASTEXITCODE` is a global, which is why the child's code survives the script boundary.

This mattered concretely: my first regression test used `pwsh -File` as the positive control and **failed to reproduce the leak at all** (that form returns 0 on fall-through). Shipped as-is, it would have been green against both broken and fixed code.

| invocation | leaky | + `$global:LASTEXITCODE = 0` | + trailing `exit 0` |
|---|---|---|---|
| in-process (what CI does) | **7** | 0 | 0 |
| `pwsh -File` | 0 | 0 | 0 |

## The fix

New `src/vs-reactor/VsProcessLib.ps1` (following the repo's existing `CoverageLib.ps1` / `PerfLib.ps1` convention so it's testable headlessly): on timeout, kill the launched process **and its descendants**, wait for them, then poll until they're gone. Duration comes from a `Stopwatch`. `bootstrap.ps1` resets `$LASTEXITCODE` and ends with an explicit `exit 0`.

### Ownership attribution — the part worth reviewing closely

The riskiest thing this code does is kill a process named `devenv`, which is also the name of the developer's IDE.

My first implementation swept every same-named process outside a pre-launch snapshot. A review pass caught that this was unsafe: because `Reinstall-Vsix.ps1` refuses to run while any devenv is alive, **the snapshot is virtually always empty**, making the sweep an unconditional "kill every devenv" with `/F`. Open Visual Studio during the 120-second window and you'd lose unsaved work. The code comment claimed the snapshot protected exactly that case — it did not.

Ownership now comes from the `Win32_Process` parent chain, captured **while the tree is still alive** (Windows doesn't re-parent orphans, so once the root dies the ancestry is unrecoverable). Anything unattributable is reported in `ForeignPids` and left running. `Stop-ProcessIds` re-checks the process name to survive pid reuse between snapshot and kill.

### Exit-code contract

`.OUTPUTS` claimed `0` meant "`/updateconfiguration` completed", but duplicate installs and a missing `devenv.exe` both fell through to a bare `exit 0`. Now:

| Code | Meaning |
|---|---|
| `0` | VSIX installed and `/updateconfiguration` completed |
| `1` | Install failed |
| `3` | VSIX installed, but `/updateconfiguration` did not complete (timed out, or skipped) |

The mapping lives in one function because three scripts depend on it: `Reinstall-Vsix.ps1` emits it, `bootstrap.ps1` and `mur upgrade` branch on it. `[ok] VS extension installed` is no longer printed for a partly-failed operation.

## What the tests actually prove

Real process trees, not mocks — `Kill()` vs `Kill($true)` differ only in what happens to a real child. The fake "devenv" is a renamed copy of `cmd.exe`.

| mutation | result |
|---|---|
| `Kill($true)` → `Kill()` | 3 assertions redden |
| ancestry → name matching (the unsafe version) | 2 redden |
| `taskkill.exe` → bogus path | suite fails |
| `Drained` hard-coded `$true` | 1 reddens |
| exit mapping always `0` | 1 reddens |
| exit routed back to a literal | 1 reddens |
| drop one incomplete-flag write | 1 reddens |
| remove trailing `exit 0` / the `$LASTEXITCODE` reset | 2 / 1 redden |
| **`Stopwatch` → `ExitTime - StartTime`** | **all pass** |

**Two things I want to flag rather than bury.**

The end-to-end attribution test earned its place. My first version exercised the helpers directly, and reverting the product to name-matching left all 31 assertions green — a vacuous test for the exact safety property it was named after. Driving the real entry point with a foreign same-named process running is what makes it bite.

The duration assertion **does not discriminate**. The −425-year symptom is a race on reading `ExitTime` before a *heavy* process is reaped; a `cmd.exe` fake dies sub-millisecond, so the race never opens. The `Stopwatch` is kept because it cannot be wrong, but that assertion is a bounds check on the symptom, documented as such in the test header and README rather than left to look like a guarantee.

## Also

- New `.github/workflows/vs-reactor-lib-tests.yml` (`windows-latest` — every behaviour under test is Windows-specific).
- `tests/vs_reactor/ci/README.md`, matching the sibling suites.
- The "Visual Studio is running" guard now reports process start times, so a leftover is distinguishable from a real IDE in one line.

**Deliberately unchanged:** the 120 s timeout (healthy 18.7 runs take 27–99 s), skipping `/updateconfiguration` in CI (it would hide the path this workflow exists to exercise), and force-killing from the guard (it would kill a developer's IDE). The `no Reactor.VsExtension folder appeared within 30s` warning is confirmed unrelated per the issue.

## Verification

35 + 13 assertions pass. `dotnet build src/Reactor.Cli -c Release` clean (0 warnings). Verified under **Windows PowerShell 5.1**, where `Process.Kill(bool)` is *absent* — so the `taskkill /T /F` fallback is the only path there, not a rare edge case — draining to zero survivors with no `$LASTEXITCODE` leak.

<sub>One pre-existing, unrelated thing noticed and left alone: `bootstrap.ps1` declares `#requires -Version 5.1` but won't *parse* under 5.1 — it's BOM-less UTF-8 with em dashes, which 5.1 decodes as ANSI. Confirmed identical at `main` and confirmed decode-only (`ParseInput` on correctly-decoded text gives 0 errors). Harmless today because CI runs it under `shell: pwsh`.</sub>

### DIFF
--- a/.github/workflows/vs-reactor-lib-tests.yml
+++ b/.github/workflows/vs-reactor-lib-tests.yml
@@ -1,0 +1,49 @@
+# Fast, headless tests for the PowerShell that installs the Reactor VS extension:
+#   - src/vs-reactor/VsProcessLib.ps1  (devenv /updateconfiguration timeout + tree kill)
+#   - src/vs-reactor/Reinstall-Vsix.ps1 + bootstrap.ps1 (the exit-code handshake)
+# No build, no dotnet, no Visual Studio — the "devenv" under test is a renamed
+# copy of cmd.exe. Runs on windows-latest because the process semantics being
+# tested (tree kill, taskkill fallback, Get-Process by name) are Windows-specific.
+#
+# Guards issue #1074: on VS 18.8 `devenv /updateconfiguration` never returns, and
+# a bare Process.Kill() orphaned its child, poisoning every later VSIX install on
+# the machine. Each test script exits non-zero on any failed assertion.
+name: VS extension script tests
+
+on:
+  push:
+    paths:
+      - 'bootstrap.ps1'
+      - 'src/vs-reactor/VsProcessLib.ps1'
+      - 'src/vs-reactor/Reinstall-Vsix.ps1'
+      - 'tests/vs_reactor/ci/**'
+      - '.github/workflows/vs-reactor-lib-tests.yml'
+  pull_request:
+    paths:
+      - 'bootstrap.ps1'
+      - 'src/vs-reactor/VsProcessLib.ps1'
+      - 'src/vs-reactor/Reinstall-Vsix.ps1'
+      - 'tests/vs_reactor/ci/**'
+      - '.github/workflows/vs-reactor-lib-tests.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  vs-reactor-scripts:
+    name: VsProcessLib + bootstrap exit-code tests
+    runs-on: windows-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run VsProcessLib tests
+        shell: pwsh
+        run: ./tests/vs_reactor/ci/VsProcessLib.Tests.ps1
+
+      - name: Run bootstrap exit-code tests
+        shell: pwsh
+        run: ./tests/vs_reactor/ci/BootstrapExitCode.Tests.ps1
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -557,11 +557,12 @@ if ($SkipVsExtension) {
                     $vsixExit = $LASTEXITCODE
                     # Reinstall-Vsix.ps1's exit-code contract (see its .OUTPUTS):
                     # 3 == VSIX installed but `devenv /updateconfiguration`
-                    # timed out. Don't claim [ok] for that, and don't fail
-                    # either — the extension is installed and usable.
+                    # did not complete (timed out, or skipped). Don't claim [ok]
+                    # for that, and don't fail either — the extension is
+                    # installed and usable.
                     if ($vsixExit -eq 3) {
                         Write-Host ''
-                        Write-Host "    [warn] VS extension installed into $($target.displayName), but 'devenv /updateconfiguration' timed out and was terminated. Launch VS once to finish the menu merge; if View -> Other Windows -> Reactor Preview is missing, re-run src\vs-reactor\Reinstall-Vsix.ps1." -ForegroundColor Yellow
+                        Write-Host "    [warn] VS extension installed into $($target.displayName), but 'devenv /updateconfiguration' did not complete (see the warnings above). Launch VS once to finish the menu merge; if View -> Other Windows -> Reactor Preview is missing, re-run src\vs-reactor\Reinstall-Vsix.ps1." -ForegroundColor Yellow
                     } elseif ($vsixExit -ne 0) {
                         # Don't fail the whole bootstrap — the rest of the install is
                         # usable without the VS extension. Surface the failure clearly
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -554,15 +554,28 @@ if ($SkipVsExtension) {
                     # the next launch. -VsInstanceId pins to the same instance we probed.
                     $powerShellExe = (Get-Process -Id $PID).Path
                     & $powerShellExe -NoLogo -NoProfile -ExecutionPolicy Bypass -File $reinstall -Configuration $Configuration -VsInstanceId $target.instanceId
-                    if ($LASTEXITCODE -ne 0) {
+                    $vsixExit = $LASTEXITCODE
+                    # Reinstall-Vsix.ps1's exit-code contract (see its .OUTPUTS):
+                    # 3 == VSIX installed but `devenv /updateconfiguration`
+                    # timed out. Don't claim [ok] for that, and don't fail
+                    # either — the extension is installed and usable.
+                    if ($vsixExit -eq 3) {
+                        Write-Host ''
+                        Write-Host "    [warn] VS extension installed into $($target.displayName), but 'devenv /updateconfiguration' timed out and was terminated. Launch VS once to finish the menu merge; if View -> Other Windows -> Reactor Preview is missing, re-run src\vs-reactor\Reinstall-Vsix.ps1." -ForegroundColor Yellow
+                    } elseif ($vsixExit -ne 0) {
                         # Don't fail the whole bootstrap — the rest of the install is
                         # usable without the VS extension. Surface the failure clearly
                         # so users debugging install issues see it.
                         Write-Host ''
-                        Write-Host "    [warn] VS extension install reported a non-zero exit code ($LASTEXITCODE). The rest of the bootstrap completed; re-run src\vs-reactor\Reinstall-Vsix.ps1 directly to retry." -ForegroundColor Yellow
+                        Write-Host "    [warn] VS extension install reported a non-zero exit code ($vsixExit). The rest of the bootstrap completed; re-run src\vs-reactor\Reinstall-Vsix.ps1 directly to retry." -ForegroundColor Yellow
                     } else {
                         Write-Ok "VS extension installed into $($target.displayName) — launch VS, then View -> Other Windows -> Reactor Preview."
                     }
+                    # Write-Host does NOT clear $LASTEXITCODE, so without this
+                    # the child's non-zero code survives to the end of the
+                    # script and fails callers that check it (issue #1074:
+                    # bootstrap.yml's `if ($LASTEXITCODE -ne 0) { throw }`).
+                    $global:LASTEXITCODE = 0
                 }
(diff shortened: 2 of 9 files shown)


azchohfi:
Closing as a duplicate of #1075.

Both PRs fix #1074. This branch was opened from a forked session that inherited the parent session's history and independently redid the same review fixes — my fault for not spotting the duplication before opening a second PR.

#1075 is the one to take: it is further along (13 commits vs 2), has all 24 checks green including `bootstrap.ps1 (windows-latest)` on a runner that drew the exact failing VS build from the issue (`a4357766` / 18.8.12023.21), and ran its Copilot review loop to zero comments.

Any findings unique to this branch have been handed to #1075 directly.

PR #1089, opened 2026-08-13
Update js-yaml and shell-quote in committed RN harness lockfiles

pranav-gupta-msft:
The React Native harnesses under tests/ carry committed package-lock.json files, so an upstream advisory fix never reaches us until the lockfile is updated.

Note that js-yaml 3.15.0 and 4.3.0 are NOT sufficient. Those releases fix CVE-2026-59869 (merge-key chains forcing quadratic CPU), but the follow-up CVE-2026-59870 / GHSA-5p4m-2wfm-xmqj (quadratic CPU in !!omap resolution) was not included in them; both the 3.x and 4.x lines remained vulnerable. The first patched releases are 3.15.1 and 4.3.1.

  js-yaml      3.14.2, 3.15.0 -> 3.15.1
  js-yaml      4.1.1,  4.3.0  -> 4.3.1
  shell-quote  1.8.4          -> 1.10.0   GHSA-395f-4hp3-45gv

Affected: tests/startup_perf/BlankRNW, tests/stress_perf_rn/StocksGrid, tests/stress_perf_rn/VirtualList. VirtualList already carried shell-quote 1.10.0 and is left untouched.

None of these are direct dependencies, and every declaring range already permitted the patched version (js-yaml ^4.1.0 and ^3.13.1, shell-quote ^1.8.4, ^1.8.3 and ^1.6.1) - only the lockfile held the vulnerable versions, so package.json needs no change.

Edited surgically rather than regenerated. A previous attempt with `npm audit fix` rewrote ~70 unrelated entries and, because it resolved through a proxy registry that strips dist.integrity, silently downgraded those entries from sha512 to sha1. This change touches only the eight affected entries (version, resolved, integrity), leaving every other line byte-identical: 24 changed lines total.

Both the dependency and bin sets are unchanged across all three bumps (js-yaml 3.x: argparse ^1.0.7 + esprima ^4.0.0; js-yaml 4.x: argparse ^2.0.1; shell-quote: none), so no other lockfile entries are affected. All integrity values are sha512, computed by hashing the published tarballs; the method was validated against the shell-quote 1.10.0 integrity already present in this repo, which it reproduces exactly. All resolved URLs remain on registry.npmjs.org.

Verified: `npm ci` succeeds in all three projects and leaves the lockfiles unmodified, which also confirms every integrity value validates against the real tarball; the patched versions are the ones installed on disk; `npm audit` no longer reports js-yaml or shell-quote in any of the three (it still reports unrelated advisories, so the check is not vacuous); and all three packages load and parse correctly, including the !!omap path covered by the newer advisory.

### DIFF
--- a/tests/startup_perf/BlankRNW/package-lock.json
+++ b/tests/startup_perf/BlankRNW/package-lock.json
@@ -2242,9 +2242,9 @@
       }
     },
     "node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-      "version": "3.15.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.15.0.tgz",
-      "integrity": "sha512-ttBQIIQPDeLjpPOohtUdXuXUVoA2uIB6fEH9HyJ7234s5mBJ5wTx20njxplLZQgLaOfpmPQA7X2t5AX6tIPbog==",
+      "version": "3.15.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.15.1.tgz",
+      "integrity": "sha512-S99WuO3HlhO3XN41EtYUNl9zzXjoJx7QvmipxsJVxtCBT0YHEFy+iOJhjSvrmV12nYhWpZaM8lPHkJm0yUMbag==",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


codemonkeychris (approved):
looks good to me, i'll merge

PR #1106, opened 2026-08-19
Add min/max constraints to grid track sizes

### LINKED ISSUE #1105: [Feature] Grid column and row definitions are missing min and max sizes.
### Problem

I am converting an existing winui application to see how this can be implemented using the Reactor framework. In one of my codepaths I am using something like:
``` csharp
<Grid>
    <Grid.ColumnDefinitions>
        <ColumnDefinition Width="150" />
        <ColumnDefinition Width="150" />
        <ColumnDefinition Width="150" />
        <ColumnDefinition Width="*" MinWidth="150" MaxWidth="450"/>
        <ColumnDefinition Width="150" />
        <ColumnDefinition Width="150" />
    </Grid.ColumnDefinitions>
</Grid>
```
I have not found an equivalent for this in Reactor.

### Proposed solution

The file PanelAttachedHooks.cs already resolves the Gridsize to ```WinUI.ColumnDefinition``` and ```WinUI.RowDefintions```. Both expose the Min and Max properties. To keep the string based initialization intact my proposed solution is to extend the syntax for GridSize to allow for "*;min=100;max=500" when passing in a string and use Fluent syntax when passing in the structured object, for example ```GridSize.Star().MinSize(100).MaxSize(500)```.

### Alternatives considered

None. The needed property is available in the underlying api but never exposed in this library.

### Additional context

_No response_

### Confirmation

- [x] I have searched [existing issues and specs](https://github.com/microsoft/microsoft-ui-reactor/issues?q=is%3Aissue) for prior discussion of this idea.

rhartskeerl:
This PR adds optional minimum and maximum sizing support for grid columns and rows while preserving the existing  `GridSize[]`, `GridDefinition`, and `Factories.Grid(...)` APIs.

The code example in the issue can now be implemented as:
``` csharp
GridSize[] gridColumns = [
    GridSize.Px(150),
    GridSize.Px(150),
    GridSize.Px(150),
    GridSize.Star().MinSize(100).MaxSize(400),
    GridSize.Px(150),
    GridSize.Px(150)
];
```

## Summary

- Added nullable Min and Max metadata to GridSize.
- Added MinSize(double) and MaxSize(double) fluent extension methods. Extended GridSize.ToString() to include optional min and max clauses.
- Updated grid track parsing to apply:
  - MinWidth and MaxWidth to ColumnDefinition
  - MinHeight and MaxHeight to RowDefinition
- Preserved the existing string-based GridDefinition representation and factory signatures.
- Added tests covering:
  - Fluent star, auto, and pixel track syntax
  - Serialization and parser round-tripping
  - Existing unconstrained track behavior

## Linked issue / spec

This should fix #1105

## Test plan
- Added tests for new functionality.
- Verified existing tests that target this code path.
- Verified behavior in another project.


rhartskeerl:
@microsoft-github-policy-service agree

azchohfi (changes requested):
Main blocker: `MinSize`/`MaxSize` don't validate anything, and a bad value doesn't just drop the constraint — it eats the base track size with it.

`GridSize.Px(200).MinSize(-5)` stringifies to something `TryParse` rejects, and then:

```csharp
var parsed = GridTrackDefinition.TryParse(def, out var track)
    ? track
    : new GridTrackDefinition(GridSize.Star(), null, null);
```

...so that 200px column comes back as `1*`. Same deal with `MinSize(500).MaxSize(100)`. Should throw at the boundary. Guarding just the two extension methods won't cover it though — `Min`/`Max` are public `init`, so `x with { Min = -5 }` walks straight past them. Either validate on the properties or make them internal.

The fallback should also keep the base size and drop only the bad clauses instead of resetting the whole track.

Root cause is that `GridDefinition` holds `string[]` and the typed ctor stringifies immediately (`Element.cs:2425-2436`), so we serialize a perfectly good `GridSize[]` just to reparse it over in `PanelAttachedHooks`. If the typed path carried `GridTrackDefinition[]` through, none of this would be reachable.

Other stuff:

`ToString()` emits `;min=` but `Parse` doesn't accept it, so `GridSize.Parse(GridSize.Star().MinSize(100).ToString())` throws. The existing round-trip test only covers unconstrained values so it slips through. Either `Parse` learns the clauses or the wire format moves somewhere internal.

`"0*"` used to work and now falls back to `1*` — `GridSize.Parse` wants `stars > 0`, the old switch didn't care. Minor, but silent.

`GridTrackDefinition` carries `Min`/`Max` and also holds a `GridSize` that carries `Min`/`Max`. Pick one.

Tests: parser side is good, but nothing touches the WinUI side at all — whether `MinWidth`/`MaxWidth` actually land on the definitions, whether clamping happens. `LayoutFixtures.cs:78-109` already mounts a real Grid and pokes at `RowDefinitions`, so extending that should be cheap. Parser gaps: null, trailing `;`, `;=5`, `;min=`, duplicate `max`, `min == max`, Infinity. Also `parts.Length == 0` can't happen, `Split` always returns at least one.

Nits: the `try`/`catch (FormatException)` around `GridSize.Parse` is doing `TryParse`'s job — worth adding a real `GridSize.TryParse`. And `layout.md.dt` still only documents Auto/Star/Px, needs the new methods and the `;min=` grammar (guide is generated, so the template, not the `.md`).

rhartskeerl:
@azchohfi Thanks for the feedback. I wanted to check a few things before I make any adjustments.

I noticed that the `GridDefinition` ctor takes a `string[]` or `GridSize[]`. But when you pass in `GridSize[]` it changes it to `string`. And then, as you mentioned changes it back in `PanelAttachedHooks`. My initial try was to change the ctor in `GridDefintion` around taking `GridSize` as the base ctor and overloading `string[]`. Then do the string parsing there. In `PanelAttachedHooks` the conversion would be obsolete and you could directly reshape the `GridSize` into `ColumnDefinition` and `RowDefinition`. But when I did that a lot of tests in other parts broke, for example several `DataGrid` tests. So I tried to come up with a solution that would not break other code paths. Let me rethink my original approach as I believe that is much cleaner. My thought is to leave the original implementation as is but take `GridSize[]` in the ctor and only use fluent syntax for passing in the Min/Max values. The `PanelAttachedHooks` can be fixed when I overload the ctors for `ParseColumnDef` and `ParseRowDef`.

For the boundary checks I found that in a current WinUI solution the checks are also not in place. So this works perfectly well:
``` csharp
<Grid.ColumnDefinitions>
    <ColumnDefinition Width="*" MinWidth="600" MaxWidth="500"/>
    <ColumnDefinition Width="Auto" MinWidth="-1" MaxWidth="500"/>
    <ColumnDefinition Width="200" MinWidth="100" MaxWidth="100"/>
</Grid.ColumnDefinitions>
```
A `MinWidth` of -1 seems to become 0 and a `MinWidth` that is larger than `MaxWidth` overrides `MaxWidth`. As I want to mimic the original xaml behavior as much as possible I did not change this.

Would love to hear your feed back on this.

azchohfi:
Yeah, you're right on the XAML side, and I'll take your testing on the coercion behavior.

But that's the argument for changing this, not leaving it. I ran the current code to be sure:

- `GridSize.Px(200).MinSize(-5)` → `"200;min=-5"` → TryParse false
- `GridSize.Star(2).MinSize(500).MaxSize(100)` → `"2*;min=500;max=100"` → TryParse false

and ParseColumnDef turns a false into `new GridTrackDefinition(GridSize.Star(), null, null)`, so both end up as `1*`. XAML would give you a 200px column and a 600-wide column. So we're not matching XAML today, we're throwing the base size away, which is worse than either behavior.

If parity is the goal the fix is small: stop rejecting those two in TryParse, clamp negative to 0, let min > max through and let WinUI apply its own precedence. Then there's nothing to complain about on validation either.

What I do still care about is the malformed stuff. `"200;min=abc"` and `"200;minimum=5"` also come back false, so a typo costs you the whole track, not just the constraint. Those should keep the size and drop the bad clause.

On the ctor, the thing to watch isn't the ctor shape, it's that the apply path is gated on reference identity (`comparer: GridDefinitionReferenceComparer.Instance` in PanelAttachedHooks). DataGrid leans on that pretty hard — DataGridState hands back a cached `_cachedGridDef` when nothing's changed, and DataGridComponent holds static RootGridDefCache/RootRowDefsCache arrays. Whatever shape you land on, the same logical definition has to keep giving back the same instance or the grid rebuilds its defs every render.

I did wonder if the breakage was GridSize.Parse being stricter than the old switch, but I checked and the strings DataGrid actually produces ("24", "40", "Auto", "*", and computed widths like "120"/"150.5") all parse fine, so that's not it. What were the actual failures? Happy to dig in if you push the branch.

Middle ground works for me.

rhartskeerl:
These are the changes on my initial thought. The test errors for DataGrid that did not compile are all because they check against a string instead of GridSize. I did add an implicit converter for string and that fixes it for the most part. It does expose that this could cause a breaking change for any code that expects GridSize to be a string. 

I haven't done the Winui checks yet and did not change documentation to describe the behavior. I first wanted to make sure this would actually be a better implementation.

azchohfi:
Yeah, this is the right shape. `GridDefinition` holding `GridSize[]` and `PanelAttachedHooks` reading `.Min`/`.Max` straight off the struct beats the string round-trip, and there's no `;min=` grammar to document now. Plain doubles defaulting to 0/PositiveInfinity also settles the validation thing from last time, since those are WinUI's own defaults.

Drop the implicit string operator though. Not for compat — breaking is fine, I'd rather have the clean type.

I removed it and built to see what actually leans on it: Reactor, Reactor.Advanced and minesweeper all compile fine. Only Reactor.Tests breaks, 34 errors but 16 call sites, all `Assert.Equal("<track>", <GridSize>)`. So it's test compat, not shipping compat. And it's hurting those tests, since `ToString()` drops Min/Max — both of these pass green on values they're no longer checking:

```csharp
Assert.Equal("*", GridSize.Star().MinSize(999).MaxSize(1000));
Assert.Equal("24", GridSize.Px(24).MinSize(500));
```

Worth converting those 16 rather than deleting them — they cover InterspersedGrid's star/separator interleaving, UniformGrid, and DataGrid's 24/40/widths/Auto ordering. Should be mechanical, and they get stronger, since several are currently substring checks on the string form (`Assert.Contains("0.7", rows[0])` → `Assert.Equal(GridSize.Star(0.7), rows[0])`).

Two other things:

`default(GridSize).Max` is 0, not infinity — property initializers don't run for `default` or array slots. Factories are all fine, but `new GridSize[n]` with an unfilled slot gives you MaxWidth 0, a silently collapsed column. Nothing hits it today (Dsl.cs:750 and minesweeper fill every slot), just a landmine. Probably treat `<= 0` as unbounded in ParseColumnDef/ParseRowDef.

`GridDefinition(string[])` now throws via `GridSize.Parse` where the old `_ =>` arm gave you `1*`. Better behavior, just needs a CHANGELOG line.

rhartskeerl:
I have made the following changes:
- Changed the ctor to accept Min/Max as optional values
- Moved validation logic to the ctor
- Added safeguards for default(GridSize)
- Refactored all tests to accept GridSize as object instead of string

I am not clear on the behavior change in `GridDefintion(string[])`. I might be overlooking a code path because I don't see a change in functionality there.

azchohfi:
You're right, I overstated that one. There's no string `Grid(...)` factory anymore, so the only door in is `new GridDefinition(string[], ...)`, and everything in-repo feeds it valid strings. That `_ =>` arm was effectively dead already. Nothing to fix, just CHANGELOG fodder:

```csharp
new GridDefinition(["garbage"], ["*"])   // was silent 1*, now FormatException
new GridDefinition(["0*"], ["*"])        // was a 0-weight star, now FormatException
new GridDefinition([" Auto "], ["*"])    // was 1*, now actually Auto (this one's an upgrade)
```

Default handling and the operator removal both look right.

The nullable backing fields did smuggle in one thing though. Record structs do equality over fields, not properties:

```csharp
Assert.Equal(new GridSize(0, GridUnitType.Auto), default(GridSize));
// Expected: Auto
// Actual:   Auto
```

Same Value, Type, Min, Max and ToString, but `_min`/`_max` are null vs 0/Infinity, so they're unequal and the hashes differ too. Render path's fine since GridDefinition is reference-compared, but anyone comparing GridSize or keying on it gets that message, and I would not wish that debugging session on anyone.

Normalising in the ctor keeps your default-safety and fixes it:

```csharp
_min = min == 0 ? null : min;
_max = double.IsPositiveInfinity(max) ? null : max;
```

Also CHANGELOG: `Deconstruct` is gone and `Value`/`Type` lost `init`, so `with { Value = ... }` is now CS0200.

Minor, your call: `.MinSize(100).MaxSize(400)` works, `.MaxSize(100).MinSize(500)` throws on the intermediate state. Comes with strict validation, just noting it's the opposite of the XAML-parity you were arguing for earlier.

rhartskeerl:
I made the change to the ctor and added a test to validate equality still works between the default values and default operator. 

But thinking this through and also in regard to your last comment on the XAML-parity I wonder if it would not be better to:
- Reinstate the original ctor: `public readonly record struct GridSize(double Value, GridUnitType Type)`.
- Set Min and Max as nullable fields.
- This will bring back the `with { Value = ...} ` logic.
- Check for `null` on min and max fields in `PanelAttachedHooks.cs` and set them accordingly in `ParseColumnDef` and `ParseRowDef`. 

I have a feeling that will give cleaner code and you leave the logic up to WinUI as it is with Xaml today.

azchohfi:
Yes, I like that a lot better — go for it.

The bit that makes it worth doing beyond the cleaner code: it closes the equality problem by construction rather than by patching. Right now the extensions materialize the properties (`size.Max` is `_max ?? PositiveInfinity`), so a null turns into a concrete Infinity on the way through and `GridSize.Star().MinSize(100)` ends up unequal to `new GridSize(1, Star, 100)` even though every observable value matches. The `default` case you just fixed happens to be the one arrangement where both sides leave the fields null. With `with { Min = ... }` there's a single construction path, so it can't drift apart in the first place.

Couple of things you'll probably hit, just so they're not a surprise:

- `Min`/`Max` become `double?` publicly, so `GridSize_Default_Values_Are_Set_Correctly` wants `Assert.Null` instead of 0/Infinity, and the expected arrays in the fluent tests become `double?[]`.
- Worth keeping the guards in `Star()`/`Px()` since the primary ctor stops validating — that's where they lived before this PR anyway.

Unrelated to which direction you pick: `GridSize_Invalid_MinMax_Throws` isn't testing what it looks like. All three cases pass `value: 0` with `Star`, so they hit "Star weight must be > 0" and never reach the min/max checks — ParamName comes back as `"value"` on all three. A legal weight makes them exercise the real guards. If validation moves to WinUI, the test just goes away instead.

rhartskeerl:
This looks a lot cleaner. I fixed the tests and removed the `MinMax()` extension method. To me it felt like I was just throwing in options there.

azchohfi:
Nice, this is much tidier. Dropping MinMaxSize was the right call too — MinSize().MaxSize() reads fine.

Equality's consistent now across all the construction paths, `with { Value = ... }` is back, unset reads as null, and the guards are where they should be. Both api.txt copies are in sync.

One thing that fell out of removing the ctor validation. I ran the values that can now reach WinUI through a real grid:

| Min | | Max | |
|---|---|---|---|
| `null` | renders | `null` | renders |
| `100` | renders | `400` | renders |
| `-5` | renders | `NaN` | **no content** |
| `NaN` | **no content** | `Infinity` | renders |
| `Infinity` | **no content** | | |

So your XAML-parity argument holds for the negative and inverted cases — those render and WinUI does its own coercion. The non-finite ones don't: `new GridSize(1, Star, double.NaN)` constructs happily, then the grid just never renders its content. No exception at Mount, the content silently isn't there. Underneath, `MinWidth = NaN` and `MinWidth = Infinity` both throw ArgumentException; `MaxWidth = Infinity` is fine since that's WinUI's own default.

So it's asymmetric — Min has to be finite, Max just can't be NaN:

```csharp
if (def.Min is { } min && double.IsFinite(min))
    colDef.MinWidth = min;
if (def.Max is { } max && !double.IsNaN(max))
    colDef.MaxWidth = max;
```

NaN in Value isn't new, for what it's worth — `Px(double.NaN)` slips past the `< 0` guard on main too and `GridLength(NaN)` throws. Only Min/Max is new here.

Also, the CHANGELOG bullet you removed about the string ctor is still true: `new GridDefinition(["This is not a grid"], ["*"])` still throws FormatException where it used to give you `1*`. The removed-init bullet was right to drop since init's back, but that one's worth keeping.

rhartskeerl:
I think this should do it. I also brought backup the removed statement in the CHANGELOG file.

azchohfi:
Nice catch on the column one before I even got to it.

One case left: `!double.IsNaN(Max)` lets negatives through, and WinUI rejects a negative MaxWidth — `MaxWidth = -5` throws ArgumentException. (Negative MinWidth it accepts, which is the confusing part.) So `Max = -5` and `Max = -Infinity` still lose the content, both axes.

`Max >= 0` covers the lot — false for null, NaN and negatives, true for +Infinity and anything non-negative:

```csharp
if (def.Max is { } max && max >= 0)
    colDef.MaxWidth = max;
```

Min's fine as-is, `IsFinite` is right there since WinUI does accept a negative min.

Still think this wants a selftest. The rows/columns split is exactly the kind of thing that comes back, and nothing in the suite would have caught it — a fixture that mounts each axis with a bad constraint and asserts the content is still visible would cover the whole family.

rhartskeerl:
I brought back the validation inside the ctor. From your comments I did some testing as well and found that `GridSize.Star(0)` throws an exception but `new GridSize(0, GridUnitType.Star)` does not. Moving the validation inside the constructor should fix this and keep the same logic for both paths. 
I also did some checks on the Xaml `GridLength` to see what it is checking and found a potential issue there as well.
```
  // This is valid
  var gridSize = new GridSize(-1, GridUnitType.Star);
  // But this will fail
  var gridLength = (GridLength)gridSize;
```
This should be fixed by adding the logic into the constructor. I added a deconstruct to keep parity with the original constructor.

Personally I am not looking for Xaml parity. I just observed it and may have overstated the importance from my side. But I do think the outcome should be clear to the user. If Xaml parity really was the case then passing in `GridSize` as column and rowdefintion would be incorrect as well. Xaml has `GridDefinition` and `RowDefinition` which take a `GridLength`. At the end of the day, I am looking for something that behaves like an HTML table not a WPF/WinUI clone.

I think that all edge cases are covered in the constructor. So a selftest would validate the behavior. But since this is passing values as is to WinUI I feel this already covered or handled by WinUI.

azchohfi:
Good catch on `Star(0)` vs `new GridSize(0, Star)` — that asymmetry was real and the ctor is the right home for it.

Fair enough on XAML parity too, and agreed the outcome just needs to be predictable. Which is the one thing I'd push back on: `with` doesn't run your constructor — it copies the value and applies the init setters, so the validation is skipped. Since the guards came out of ParseColumnDef/ParseRowDef there's now nothing between that and WinUI:

```csharp
GridSize.Star() with { Max = -5 }        // no throw
GridSize.Star() with { Min = double.NaN }
GridSize.Star() with { Value = -5, Type = GridUnitType.Pixel }
```

All of those mount and then render nothing — the content silently isn't there, both axes. Since you wanted `with { Value = ... }` back, that path is a supported way to build a GridSize, so it needs to land somewhere sane too. Either keep the projection guards as a backstop, or drop `init` on Min/Max/Value so the ctor is the only door.

One other thing the ctor now blocks that WinUI allows: `new GridSize(1, Star, null, double.PositiveInfinity)` throws, but `MaxWidth = +Infinity` is WinUI's own default for unbounded. Not a big deal since null does the same job, just noting IsFiniteAndNonNegative is stricter than WinUI there.

On the selftest — "WinUI handles it" is the bit I'd disagree with. WinUI throws, and in the render path that surfaces as silently missing content rather than an error anyone sees. That's the whole reason the rows/columns split slipped through: the code looked right, the unit tests passed, and the only symptom was an invisible column. The unit tests can't catch it either, since they can't construct a live ColumnDefinition. One fixture mounting a bad constraint on each axis and asserting the content is visible would have caught that in one run.

rhartskeerl:
So it's either move the boundary validation in `ParseColumnDef` and `ParseRowDef` or force everyone through the constructor and validate there. I choose the latter because that gives a clear indication what is being checked where before it is pushed down for rendering. I changed the `MaxSize` validation to align with WinUI.

I have extended the existing SelfTest with `MinSize` and `MaxSize` specified and validation for this configuration.

azchohfi:
Removing `init` does it — the `with` hole is closed.

Heads up though, CI's red: `GridSizeTests.cs:327`. Changing the Max rule left four assertions behind:

| line | |
|---|---|
| 327 `new GridSize(1, Star, 100, NaN)` | no longer throws |
| 329 `new GridSize(1, Star, 100, +Inf)` | no longer throws |
| 334 `Star(1).MaxSize(NaN)` | no longer throws |
| 336 `Star(1).MaxSize(+Inf)` | no longer throws |

329 and 336 are the change you meant. 327 and 334 aren't: `!double.IsNaN(max.Value) && max.Value < 0` short-circuits on NaN, so NaN skips the check entirely. Since the projection guards are gone, `MaxWidth = NaN` goes straight to WinUI and throws — content disappears on both axes, same through `MaxSize(NaN)`.

`max.Value < 0` on its own gives you what you want: false for NaN and +Inf, true for negatives. Then 327/334 keep passing and only 329/336 need updating.

The selftest is a good one. I checked it actually bites — stub out the min/max writes in ParseColumnDef/ParseRowDef and both new checks go red, so it's not just riding along on the default layout.

rhartskeerl:
Sorry. That makes me feel a bit silly. Not to defend myself, but both the unittests and selftest always come up with errors on my machine. All because of string comparisons and locale differences. But the number of errors did change so I should have caught something was off. 

To make sure my fixes are what I think I ran a filtered test to validate the result. It looks a bit better now.

azchohfi:
Don't feel silly at all — you've actually found something.

I forced the suite to nl-NL to see what you're seeing, and there really are 12 failures out of 13,370. But one of them isn't test noise:

```
Expected: "rgba(128, 64, 32, 0.5)"
Actual:   "rgba(128, 64, 32, 0,5)"
```

`D3Color.ToRgb()` interpolates the opacity with the current culture, so on any comma-decimal locale it emits invalid CSS. That's a real bug on `main` that our CI can't see because it only ever runs en-US. Same story for `mur localize status` printing `100,0%`. The rest are tests asserting en-US strings against paths that use CurrentCulture deliberately.

So thanks for mentioning it — I've filed it separately, and it explains why the red run didn't stand out to you.

In the meantime, exactly those 12 are your baseline locally; anything past that is real. Your area is clean either way — all 98 across GridSizeTests, DataGridPerfCacheTests and DeclarativeModifierTests pass under nl-NL, since GridSize formats and parses invariantly throughout.

The guard reads right now: NaN and negatives rejected, +Infinity allowed. CI's still churning but it's green locally.

azchohfi:
CI is green now (small flakiness), so merging it.
Again, thank you so much for your contribution! This was very productive!

azchohfi (approved):

PR #1114, opened 2026-08-24
Scale demo example fixed.

afshin29:
## Summary

The `ScaleDemo` example in the docs was missing one line.

https://microsoft.github.io/microsoft-ui-reactor/animation/#scale-transition

### DIFF
--- a/docs/guide/animation.md
+++ b/docs/guide/animation.md
@@ -96,6 +96,7 @@ class ScaleDemo : Component
             ).Padding(12)
              .CornerRadius(8)
              .Background("#e8e8e8")
+             .Scale(enlarged ? 1.5f : 1.0f)
              .ScaleTransition()
         ).Padding(24);
     }


azchohfi:
Thanks, but I'm going to close this — `docs/guide/animation.md` is generated output, not a source file. The snippet comes from `docs/_pipeline/apps/animation/App.cs` (`// <snippet:scale-transition>`), pulled in by `docs/_pipeline/templates/animation.md.dt` via `snippet="animation/scale-transition"`, so this edit gets reverted the next time anyone runs `mur docs compile` and the actual demo app keeps the bug. The catch is real though — `ScaleDemo` never calls `.Scale()`, so `enlarged` is dead state. If you'd like to redo it, add `.Scale(enlarged ? 1.5f : 1.0f)` in `App.cs` and commit the regenerated page.

PR #1146, opened 2026-08-27
Rename two files whose names still said Duct

azchohfi:
## Summary

Two files kept `Duct` in their **filenames** after the Duct → Reactor rename, even though their contents were renamed:

| Path | Declares |
|---|---|
| `samples/WinFormsInterop/SampleDuctComponent.cs` | `class SampleReactorComponent` |
| `tests/Reactor.WinFormsTests.Host/TestDuctComponent.cs` | `class TestReactorComponent` |

Neither file contains the string `Duct` anywhere in its contents — the only occurrence was in the path. So each filename contradicted the type it declared. This is the same defect class #1129 fixed when it renamed `AzureOpenAiProvider.cs`, a file that declared `CopilotTranslationProvider`: a name that lies about what it refers to.

**Why #1129 missed these two, which is the more useful half of this PR:** that sweep used `git grep`, which searches file *contents*. A path-only occurrence is invisible to *any* content regex, however sophisticated — the corpus is wrong, not the pattern. A rename audit needs a `git ls-files` path scan alongside the content scan. These two files were the entire residue.

Pure `git mv`, both renames `R100` with zero content change (`2 files changed, 0 insertions(+), 0 deletions(-)`). The classes were already correctly named, so nothing inside the files needed to move.

Two things deliberately **not** changed:

- **The historical rename record.** `docs/specs/018-namespace-rename.md` (lines 132-133, 280-281, 285-286) and `docs/specs/tasks/namespace-rename-implementation.md` (line 106) still name the old files, because the Duct → Reactor rename is their *subject matter*. Rewriting them would destroy the record; #1129 preserved them for the same reason. After this PR those 7 lines are the only remaining `*DuctComponent` references in the repo, which is correct.
- **`samples/WinFormsInterop/README.md`.** Its directory listing at line 136 already read `SampleReactorComponent.cs` — it documented the renamed filename before the file had it. The rename makes the README true, so it needed no edit. Independent evidence that the filename was the thing that was wrong, not the class.

## Linked issue / spec

Follow-up to #1129 (merged, `158eab4d`), which fixed this defect class but could only see content occurrences. Related historical context: `docs/specs/018-namespace-rename.md`.

## Test plan

A build is the real oracle here, and it is non-vacuous: each renamed file has a forcing reference at a `typeof` site — `WinFormsOutsideForm.Designer.cs:101` does `typeof(SampleReactorComponent)` and `TestForm.cs:102` does `typeof(TestReactorComponent)`. Both projects rely on the SDK default `*.cs` glob with no explicit `<Compile>` items, so if a rename had escaped the glob the type would be missing and compilation would fail with CS0246 rather than silently passing.

- [x] `dotnet build samples/WinFormsInterop -c Debug -p:Platform=x64` — **0 errors**
- [x] `dotnet build tests/Reactor.WinFormsTests.Host -c Debug -p:Platform=x64` — **0 errors**
- [x] Confirmed the diff is `R100` / zero content change on both files
- [x] `git grep -InE "SampleDuctComponent|TestDuctComponent"` before and after — after, the only hits are the 7 historical spec lines above (run with a positive control on `*ReactorComponent` to confirm the grep could match at all)
- [x] Confirmed neither `.csproj` has explicit `<Compile>` items, and no props/targets/workflow/script names either file
- [x] Confirmed both projects are in the root `Reactor.slnx`, so CI builds them — this is live code, not archived samples

**On the Release gate — not claimed, because it does not genuinely run here.** `dotnet build Reactor.slnx -c Release` fails in my environment at *restore*: 258 errors, all of them `NU1900` (234) and `NU1301` (24), from `api.nuget.org` being unreachable over TLS (`HandshakeFailure`). Zero `CS####` compile errors, because nothing ever got as far as compiling. That failure is environmental and independent of this change, but it also means I cannot assert a green Release build, so I am not going to imply one. **Please let CI's `Build solution` job be the gate on that.** (The two Debug builds above did succeed, restoring from the local package cache.)

## Risk / breaking changes

None. No public API change, no behavior change, no perf-sensitive path. Both types are `internal` sample/test-host components, and every reference to them is by type name — which is unchanged. The only thing that moved is two paths on disk.

Worth a reviewer's eye on one point: the renames are recorded as pure renames, so if your tooling shows a content diff on either file, something went wrong and I'd want to know.


azchohfi:
Status note for the human reviewer, because "no review comments" here does **not** mean "reviewed and found clean".

The Copilot reviewer returned *"Copilot wasn't able to review any files in this pull request"* — not its usual "no new comments". I checked via GraphQL rather than trusting the summary line: `reviewThreads.totalCount=0`, `comments.totalCount=0`, nothing minimized or suppressed. So there is genuinely nothing outstanding, but the reason is that the diff has **zero content lines**, and a content-oriented reviewer has nothing to read. Review impossible, not review clean — worth not conflating the two, since the automated gate contributes no signal on this PR and your manual pass carries the full weight.

Which is a fairly on-the-nose demonstration of the thing this PR is about: the reason #1129 missed these two files is that `git grep` searches contents, and a path-only change is invisible to content-oriented tooling. The reviewer just hit the same blind spot from the other side.

CI is green on all 22 checks, including the two slow load-bearing ones — `Build solution` (8m49s) and `Analyze (csharp)` (7m44s). That `Build solution` pass is doing real work here: it is the Release gate I could not run locally (restore fails against nuget.org over TLS, detailed in the description), so CI is the only place the Release `TreatWarningsAsErrors` path actually got exercised. It passed.

Not merging — over to you.

Issue #1052, opened 2026-08-01
CI compiles every guide page and discards the result — compiled-doc drift is ungated (architecture-overview.md is stale on main)


azchohfi:
## Summary

`docs/guide/**` is generated from `docs/_pipeline/templates/*.md.dt`, and the `Docs build` CI job runs a real `mur docs compile --no-screenshots --ci` on every PR. **The compiled output is then thrown away.** Only two of the 74 compiled topics are checked for freshness:

```yaml
- name: Verify docs version substitution is committed
  run: git diff --exit-code -- docs/guide/getting-started.md docs/guide/packaging.md
```

For the other 72, CI generates the evidence that a page is stale and discards it. Nothing fails, so the drift is invisible.

## Live instance

`docs/guide/architecture-overview.md` is currently stale on `main`. Its snippet source is `src/Reactor/Core/ElementFactory.cs`, which was restructured in `b65bb026` ("ElementFactory: stop stranding recycled rows on root-type flips"):

| | committed doc | compiled output |
|---|---|---|
| contains `_recyclePool.Pop()` | **yes** | no |
| contains `TryTakeCompatibleFromPool` | no | **yes** |

The published page therefore shows a `GetElement` body that no longer exists, missing the pass-2 reuse path, `CanSafelyAdopt`, and `RetireAlreadyUnmounted`.

Timeline:

```
doc  last touched: d01ea5b7  2026-07-13
src  last touched: 7a0f4d27  2026-07-28
```

Reproduce on a clean checkout of `main`:

```
dotnet run --project src/Reactor.Cli -- docs compile --no-screenshots --ci
git status --porcelain -- docs/guide
```

I measured this while merging `origin/main` into #1011. A fresh compile drifts **exactly one** file, so the rest of the corpus is currently fresh — this is a single stale page, not a broken pipeline. That is also why it is worth gating now, while the backlog is one file.

## Why this is worth a gate rather than a one-off fix

`AGENTS.md` already states the rule — *generated files must be regenerated, never side-picked* — but nothing enforces it for compiled guide pages. The failure is silent in both directions:

- a source change that alters a snippet leaves the page stale, and CI is green;
- the next unrelated PR that happens to run a compile picks up someone else's regeneration as drive-by churn in its diff, which is how I hit it.

The second is the more annoying one in a merge queue: every PR that regenerates docs inherits an unrelated 40-line hunk and has to decide whether to adopt or revert it.

[...]

PR #1157, opened 2026-08-31
Fix REACTOR_DSL_001 for ForEach projections

### LINKED ISSUE #1156: REACTOR_DSL_001 (missing .WithKey) doesn't fire for ForEach, only Select
## Summary

`REACTOR_DSL_001` (*"Dynamic list item missing `.WithKey`"*) only fires for the `Select(...)` projection shape. The equivalent `ForEach(items, item => …)` form — which is Reactor's own documented idiom for this — is silently unprotected.

The result is that **the recommended way to build a list is the one the analyzer doesn't cover.**

## Repro

Both snippets build the same UI. Only the first warns.

```csharp
// Warns: REACTOR_DSL_001
FlexColumn(
    items.Select(item =>
        FlexRow(
            CheckBox(item.IsCompleted, done => Toggle(item.Id, done)),
            TextBlock(item.Text).Flex(grow: 1),
            Button("Delete", () => Delete(item.Id))
        )
    ).ToArray<Element?>()
);

// Silent — same missing key, no diagnostic
FlexColumn(
    ForEach(items, item =>
        FlexRow(
            CheckBox(item.IsCompleted, done => Toggle(item.Id, done)),
            TextBlock(item.Text).Flex(grow: 1),
            Button("Delete", () => Delete(item.Id))
        )
    )
);
```

Verified against `Microsoft.UI.Reactor` `0.1.0-preview.13` on .NET 10.0.400 / x64, building both variants in the same directory with the same setup. The `Select` variant emitted:

```
warning REACTOR_DSL_001: Element produced by Select(...) doesn't call .WithKey(...).
Without a key, the reconciler matches by position and re-mounts every row on
insert/reorder, losing focus, animation, and ElementRef state.
```

The `ForEach` variant built clean. The `Select` case acts as a positive control, so the silence is a real gap rather than the analyzer failing to load.

## Cause

`MissingWithKeyAnalyzer.Analyze` gates DSL_001 on the method name:

https://github.com/microsoft/microsoft-ui-reactor/blob/main/src/Reactor.Analyzers/MissingWithKeyAnalyzer.cs#L89-L93

```csharp
// REACTOR_DSL_001 — a missing key on a Select projection.
if (methodName == "Select")
{
    AnalyzeMissingKey(ctx, inv);
}
```

There is no `ForEach` branch.

There's a second obstacle worth noting before anyone attempts a one-line fix: `Analyze` returns early for anything that isn't a member access —

```csharp
if (inv.Expression is not MemberAccessExpressionSyntax member) return;
```

A bare `ForEach(items, …)` imported via `using static Microsoft.UI.Reactor.Factories` is an `IdentifierNameSyntax`, so it never reaches the name check at all. Handling it means letting that shape through before the `member` cast, or registering it separately.

`AnalyzeMissingKey` itself also assumes the `Select` shape — one argument, the lambda at index 0 — whereas `ForEach(items, lambda)` takes two with the lambda at index 1.

## Why this is cheap to fix

Most of the work already exists in the same file, added for `REACTOR_DSL_002`, which **does** handle both shapes:

- `EnclosingProjectionLambda` accepts `Select` *and* Reactor's `ForEach` (L177-L188)
- `IsReactorForEachReceiver` (L200-L205) already distinguishes Reactor's static `ForEach` from `List<T>.ForEach`, `Parallel.ForEach`, and arbitrary `X.ForEach(items, lambda)` — the exact false-positive risk a naive fix would introduce

So DSL_002 correctly inspects keys inside `ForEach` projections, while DSL_001 doesn't notice a missing one. The two rules disagree about what a projection is, within one file.

## Suggested fix

1. Let bare-identifier invocations through the early return in `Analyze`.
2. Add a `ForEach` branch to DSL_001, reusing `IsReactorForEachReceiver` for the receiver check.
3. Generalize `AnalyzeMissingKey` to locate the lambda by position rather than assuming argument 0.
4. Tests covering: keyed `ForEach` (no diagnostic), unkeyed `ForEach` (diagnostic), and the negatives `IsReactorForEachReceiver` guards against — `List<T>.ForEach`, `Parallel.ForEach`, and a user-defined `X.ForEach(items, lambda)`.

No `AnalyzerReleases` row is needed — the id already exists; its coverage widens.

## Impact

Missing keys are exactly the failure DSL_001 exists to prevent: the reconciler falls back to positional matching and re-mounts rows on insert or reorder, losing focus mid-keystroke, restarting animations, and invalidating `ElementRef`s. It reproduces only after a specific interaction and reads as a haunting rather than a crash.

`ForEach` is also the shape most likely to be reached for — it's what the XML docs recommend (*"Map a list to elements (like `.map()` in React JSX)"*), it returns a single `Element` so it drops straight into a layout factory without `.ToArray<Element?>()`, and it has an `IReadOnlyList` fast path that skips the `Select` iterator and closure allocations.

Found while writing list samples for the Reactor announcement blog series: the samples were migrated from `Select` to `ForEach` for exactly those reasons, and the analyzer coverage silently disappeared along with the ceremony.

jibin7jose:
## Summary

Teach `REACTOR_DSL_001` to recognize Reactor `ForEach` projections in addition to `Select`, so missing `.WithKey(...)` warnings now cover the documented list-building idiom. Also update the code fix and add analyzer coverage for bare and qualified `ForEach` calls.

## Linked issue / spec

Fixes #1156

## Test plan

- [x] Added analyzer tests for bare `ForEach(...)`, `Factories.ForEach(...)`, and keyed `ForEach(...)`
- [x] Added code-fix coverage for `ForEach(...)`
- [x] Built `src/Reactor.Analyzers/Reactor.Analyzers.csproj --no-restore -p:Platform=x64`
- [ ] Full `dotnet test tests/Reactor.Tests` could not be completed in this environment because unrelated repo build issues blocked the test project

### DIFF
--- a/docs/_pipeline/apps/collections/App.cs
+++ b/docs/_pipeline/apps/collections/App.cs
@@ -557,7 +557,8 @@ class LetterJump : Component<IReadOnlyList<Person>>
                     {
                         if (groupStarts.TryGetValue(letter, out var start))
                             listRef.Current?.ScrollToIndex(start);
-                    }).AutomationName($"Jump to {letter}")))
+                    }).AutomationName($"Jump to {letter}")
+                        .WithKey(letter.ToString())))
         );
     }
 }
--- a/docs/_pipeline/apps/commanding/App.cs
+++ b/docs/_pipeline/apps/commanding/App.cs
@@ -199,7 +199,8 @@ class ParameterizedCommandExample : Component
                     // by design, so call .Execute(arg) directly from the click handler.
                     Button(delete.Label, () => delete.Execute?.Invoke(item))
                         .AutomationName($"Delete {item.Title}")
-                        .IsEnabled(delete.IsEnabled)))
+                        .IsEnabled(delete.IsEnabled))
+                    .WithKey(item.Id.ToString()))
         ).Padding(24);
     }
 }
--- a/docs/_pipeline/apps/components/App.cs
+++ b/docs/_pipeline/apps/components/App.cs
@@ -221,7 +221,7 @@ class MutatingPropsDont : Component<ItemsProps>
         // Don't — this mutates the parent's collection, so the parent never
         // sees the change and doesn't re-render.
         Props.Items.Add("new item");
-        return VStack(4, ForEach(Props.Items, i => TextBlock(i)));
+        return VStack(4, ForEach(Props.Items, (item, i) => TextBlock(item).WithKey($"{i}-{item}")));
     }
 }
 // </snippet:mutating-props>
--- a/docs/_pipeline/apps/controls/App.cs
+++ b/docs/_pipeline/apps/controls/App.cs
@@ -55,7 +55,7 @@ class CollectionsGroup : Component
         var items = new[] { "Alpha", "Bravo", "Charlie", "Delta" };
         return VStack(4,
             TextBlock("Items").Bold(),
-            ForEach(items, item => TextBlock($"  • {item}"))
+            ForEach(items, item => TextBlock($"  • {item}").WithKey(item))
         ).Padding(16);
     }
 }
@@ -110,7 +110,7 @@ class DataSystemGroup : Component
             ForEach(rows, r => HStack(16,
                 TextBlock(r.Item1),
                 TextBlock(r.Item2.ToString())
-            ))
+            ).WithKey(r.Item1))
         ).Padding(16);
     }
 }
--- a/docs/_pipeline/apps/dev-tooling/App.cs
+++ b/docs/_pipeline/apps/dev-tooling/App.cs
@@ -81,7 +81,7 @@ class IterationDemo : Component
                     }
                 })
             ),
-            ForEach(items, item => TextBlock($"  - {item}"))
+            ForEach(items, (item, i) => TextBlock($"  - {item}").WithKey($"{i}-{item}"))
         ).Padding(24);
     }
 }
--- a/docs/_pipeline/apps/effects/App.cs
+++ b/docs/_pipeline/apps/effects/App.cs
@@ -164,7 +164,7 @@ class FetchCancellationExample : Component
 
         return VStack(8,
             TextBox(query, setQuery, header: "Search query").Width(250),
-            ForEach(items, i => TextBlock(i))
+            ForEach(items, item => TextBlock(item).WithKey(item))
         ).Padding(24);
     }
 }
--- a/docs/_pipeline/apps/forms/App.cs
+++ b/docs/_pipeline/apps/forms/App.cs
@@ -315,7 +315,7 @@ class AutoSuggestDemo : Component
             When(matches.Length > 0, () =>
                 VStack(2,
                     ForEach(matches, m =>
-                        TextBlock(m).Padding(8, 4))
+                        TextBlock(m).Padding(8, 4).WithKey(m))
                 ).Background(Theme.CardBackground).Width(280))
         ).Padding(24);
     }
--- a/docs/_pipeline/apps/hooks/App.cs
+++ b/docs/_pipeline/apps/hooks/App.cs
@@ -55,7 +55,7 @@ class ReducerDemo : Component
                 Button("Clear", () =>
                     updateItems(_ => new List<string>()))
             ),
-            ForEach(items, item => TextBlock($"  - {item}"))
+            ForEach(items, (item, i) => TextBlock($"  - {item}").WithKey($"{i}-{item}"))
         );
     }
 }
--- a/docs/_pipeline/apps/layout/App.cs
+++ b/docs/_pipeline/apps/layout/App.cs
@@ -97,7 +97,7 @@ class ScrollBorderDemo : Component
                     VStack(4,
                         ForEach(
                             Enumerable.Range(1, 20),
-                            i => TextBlock($"Scrollable item {i}"))
+                            i => TextBlock($"Scrollable item {i}").WithKey(i.ToString()))
                     ).Padding(8)
                 ).Height(120)
             ).CornerRadius(4).Background(Theme.CardBackground)
--- a/src/Reactor.Analyzers/MissingWithKeyAnalyzer.cs
+++ b/src/Reactor.Analyzers/MissingWithKeyAnalyzer.cs
@@ -10,14 +10,16 @@ using Microsoft.CodeAnalysis.Diagnostics;
 namespace Microsoft.UI.Reactor.Analyzers;
 
 /// <summary>
-/// <c>REACTOR_DSL_001</c> — when a LINQ <c>Select</c> projects to Reactor
-/// elements and the result is materialized into a layout container's children
+/// <c>REACTOR_DSL_001</c> — when a LINQ <c>Select</c> or Reactor
+/// <c>ForEach</c> projects to Reactor elements and the result is materialized
+/// into a layout container's children
 /// (<c>VStack</c>, <c>HStack</c>, <c>FlexRow</c>, <c>FlexColumn</c>, <c>Grid</c>, ...),
 /// every projected element should call <c>.WithKey(...)</c>. Without keys, the
 /// reconciler matches positionally and re-mounts every row on insert / reorder
 /// — losing focus, animation state, and ElementRef identity.
 ///
-/// Heuristic: a <c>Select(x =&gt; expr)</c> invocation whose lambda body
+/// Heuristic: a <c>Select(x =&gt; expr)</c> or Reactor <c>ForEach(items, x =&gt; expr)</c>
+/// invocation whose lambda body
 /// (a) returns a Reactor element-shaped expression, and
 /// (b) contains no <c>.WithKey(</c> token anywhere in the lambda body.
 ///
@@ -44,7 +46,7 @@ public sealed class MissingWithKeyAnalyzer : DiagnosticAnalyzer
     private static readonly DiagnosticDescriptor Rule = new(
         Id,
         "Dynamic list item missing .WithKey",
-        "Element produced by Select(...) doesn't call .WithKey(...). Without a key, the reconciler matches by position and re-mounts every row on insert/reorder, losing focus, animation, and ElementRef state.",
+        "Element produced by Select(...) or ForEach(...) doesn't call .WithKey(...). Without a key, the reconciler matches by position and re-mounts every row on insert/reorder, losing focus, animation, and ElementRef state.",
         "Reactor.Dsl",
         DiagnosticSeverity.Warning,
         isEnabledByDefault: true,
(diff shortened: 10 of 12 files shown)


azchohfi:
Thanks for this — and for the issue write-up that preceded it. I'd just started looking at #1156 when your PR came in, and you'd already found all three obstacles: the member-access early return, the missing `ForEach` branch, and the hard-coded lambda index.

Nice catch relaxing `ExtractParameterName` to `Count >= 1` too — that makes the code fix work on the `(item, i)` overload, which I'd have missed.

Three things before CI goes green. The second one is on us, not you.

### 1. The code fix doesn't fire — your own test catches it

`CodeFix_Offers_WithKey_Item_On_ForEach` fails on this branch: the fix produces no change at all. In `FlexColumn(ForEach(...))` the `ArgumentSyntax` and the `InvocationExpressionSyntax` have identical spans, and `FindNode` walks up through a span tie by default — so the `as InvocationExpressionSyntax` cast yields `null`. `Select` never hit this because its diagnostic sits inside a larger `.ToArray()` argument.

```csharp
// MissingWithKeyCodeFix.RegisterCodeFixesAsync
var reported = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
if (reported is not InvocationExpressionSyntax selectInv) continue;
```

### 2. The guide's own samples are the unprotected shape

Not discoverable from the issue — sorry. `docs/_pipeline/apps/**` is the literal source of the `snippet=` blocks in `docs/guide/`, so it gets the analyzer wired in via its own `Directory.Build.props`, and CI builds it with `TreatWarningsAsErrors` and no suppressions permitted. Widening the rule turns **10 sites across 9 doc apps** into errors — which is the rule working: we were shipping unkeyed lists as the recommended idiom.

<details>
<summary>Patch for the 10 sites — should <code>git apply</code> cleanly</summary>

```diff
diff --git a/docs/_pipeline/apps/collections/App.cs b/docs/_pipeline/apps/collections/App.cs
index 529b9501..1ceca35d 100644
--- a/docs/_pipeline/apps/collections/App.cs
+++ b/docs/_pipeline/apps/collections/App.cs
@@ -557,7 +557,8 @@ class LetterJump : Component<IReadOnlyList<Person>>
                     {
                         if (groupStarts.TryGetValue(letter, out var start))
                             listRef.Current?.ScrollToIndex(start);
-                    }).AutomationName($"Jump to {letter}")))
+                    }).AutomationName($"Jump to {letter}")
+                        .WithKey(letter.ToString())))
         );
     }
 }
diff --git a/docs/_pipeline/apps/commanding/App.cs b/docs/_pipeline/apps/commanding/App.cs
index 39c16a24..58eff324 100644
--- a/docs/_pipeline/apps/commanding/App.cs
+++ b/docs/_pipeline/apps/commanding/App.cs
@@ -199,7 +199,8 @@ class ParameterizedCommandExample : Component
                     // by design, so call .Execute(arg) directly from the click handler.
                     Button(delete.Label, () => delete.Execute?.Invoke(item))
                         .AutomationName($"Delete {item.Title}")
-                        .IsEnabled(delete.IsEnabled)))
+                        .IsEnabled(delete.IsEnabled))
+                    .WithKey(item.Id.ToString()))
         ).Padding(24);
     }
 }
diff --git a/docs/_pipeline/apps/components/App.cs b/docs/_pipeline/apps/components/App.cs
index e5803276..1d33cf59 100644
--- a/docs/_pipeline/apps/components/App.cs
+++ b/docs/_pipeline/apps/components/App.cs
@@ -221,7 +221,7 @@ class MutatingPropsDont : Component<ItemsProps>
         // Don't — this mutates the parent's collection, so the parent never
         // sees the change and doesn't re-render.
         Props.Items.Add("new item");
-        return VStack(4, ForEach(Props.Items, i => TextBlock(i)));
+        return VStack(4, ForEach(Props.Items, (item, i) => TextBlock(item).WithKey($"{i}-{item}")));
     }
 }
 // </snippet:mutating-props>
diff --git a/docs/_pipeline/apps/controls/App.cs b/docs/_pipeline/apps/controls/App.cs
index 72fd261a..f143c811 100644
--- a/docs/_pipeline/apps/controls/App.cs
+++ b/docs/_pipeline/apps/controls/App.cs
@@ -55,7 +55,7 @@ class CollectionsGroup : Component
         var items = new[] { "Alpha", "Bravo", "Charlie", "Delta" };
         return VStack(4,
             TextBlock("Items").Bold(),
-            ForEach(items, item => TextBlock($"  • {item}"))
+            ForEach(items, item => TextBlock($"  • {item}").WithKey(item))
         ).Padding(16);
     }
 }
@@ -110,7 +110,7 @@ class DataSystemGroup : Component
             ForEach(rows, r => HStack(16,
                 TextBlock(r.Item1),
                 TextBlock(r.Item2.ToString())
-            ))
+            ).WithKey(r.Item1))
         ).Padding(16);
     }
 }
diff --git a/docs/_pipeline/apps/dev-tooling/App.cs b/docs/_pipeline/apps/dev-tooling/App.cs
index ad228623..05512c8b 100644
--- a/docs/_pipeline/apps/dev-tooling/App.cs
+++ b/docs/_pipeline/apps/dev-tooling/App.cs
@@ -81,7 +81,7 @@ class IterationDemo : Component
                     }
                 })
             ),
-            ForEach(items, item => TextBlock($"  - {item}"))
+            ForEach(items, (item, i) => TextBlock($"  - {item}").WithKey($"{i}-{item}"))
         ).Padding(24);
     }
 }
diff --git a/docs/_pipeline/apps/effects/App.cs b/docs/_pipeline/apps/effects/App.cs
index e2d3f6ba..5871c632 100644
--- a/docs/_pipeline/apps/effects/App.cs
+++ b/docs/_pipeline/apps/effects/App.cs
@@ -164,7 +164,7 @@ class FetchCancellationExample : Component
 
         return VStack(8,
             TextBox(query, setQuery, header: "Search query").Width(250),
-            ForEach(items, i => TextBlock(i))
+            ForEach(items, item => TextBlock(item).WithKey(item))
         ).Padding(24);
     }
 }
diff --git a/docs/_pipeline/apps/forms/App.cs b/docs/_pipeline/apps/forms/App.cs
index 3691d4ae..c3df20ee 100644
--- a/docs/_pipeline/apps/forms/App.cs
+++ b/docs/_pipeline/apps/forms/App.cs
@@ -315,7 +315,7 @@ class AutoSuggestDemo : Component
             When(matches.Length > 0, () =>
                 VStack(2,
                     ForEach(matches, m =>
-                        TextBlock(m).Padding(8, 4))
+                        TextBlock(m).Padding(8, 4).WithKey(m))
                 ).Background(Theme.CardBackground).Width(280))
         ).Padding(24);
     }
diff --git a/docs/_pipeline/apps/hooks/App.cs b/docs/_pipeline/apps/hooks/App.cs
index ed80b9d1..db5704c8 100644
--- a/docs/_pipeline/apps/hooks/App.cs
+++ b/docs/_pipeline/apps/hooks/App.cs
@@ -55,7 +55,7 @@ class ReducerDemo : Component
                 Button("Clear", () =>
                     updateItems(_ => new List<string>()))
             ),
-            ForEach(items, item => TextBlock($"  - {item}"))
+            ForEach(items, (item, i) => TextBlock($"  - {item}").WithKey($"{i}-{item}"))
         );
     }
 }
diff --git a/docs/_pipeline/apps/layout/App.cs b/docs/_pipeline/apps/layout/App.cs
index 06b573d3..753a69d1 100644
--- a/docs/_pipeline/apps/layout/App.cs
+++ b/docs/_pipeline/apps/layout/App.cs
@@ -97,7 +97,7 @@ class ScrollBorderDemo : Component
                     VStack(4,
                         ForEach(
                             Enumerable.Range(1, 20),
-                            i => TextBlock($"Scrollable item {i}"))
+                            i => TextBlock($"Scrollable item {i}").WithKey(i.ToString()))
                     ).Padding(8)
                 ).Height(120)
             ).CornerRadius(4).Background(Theme.CardBackground)
```

</details>

Two bits in there worth reading rather than pasting:

- `components` / `dev-tooling` / `hooks` get a composite `$"{i}-{item}"`. Those lists append user-typed strings, so items can repeat — and `KeyedListDiff` treats duplicate keys as a bailout to positional diffing, so a bare string key would look correct and silently do nothing.
- `commanding` needs `item.Id.ToString()`. `Id` is an `int` there, and the bare `.WithKey(item)` overload is constrained to `IReactorKeyed` (`CS0315`).

### 3. Regenerate the guide

Your change sits inside the `// <snippet:with-key-rule>` region, which `docs/guide/analyzer-architecture.md` embeds — so that page is stale and the freshness gate fails.

```bash
dotnet run --project src/Reactor.Cli -- docs compile --no-screenshots
```

---

**Leave the rest to us.** I'll follow up with a symbol gate on the `ForEach` arm — a bare `ForEach` from someone else's `using static` still passes the receiver check, and `DSL_001` is a warning-as-error — plus the samples, a CHANGELOG entry, and the prose on that guide page.

To check locally:

```bash
dotnet test tests/Reactor.Tests --filter-class "*MissingWithKeyAnalyzerTests*" -p:Platform=x64 -p:SkipSignaturesGen=true
dotnet build docs/_pipeline/apps/DocApps.proj -c Debug -p:Platform=x64 -p:TreatWarningsAsErrors=true -p:WarningsNotAsErrors=NU1900
```

Happy to push any of this to your branch if you'd rather — just say the word. Glad you picked this one up.

jibin7jose:
@azchohfi Alexandre, I applied the requested fixes. Please take a look at the pending changes when you have time. Local verification is blocked by environment build errors, and the generated guide still needs regeneration in CI.

azchohfi:
Correction on my part, and good news: **your PR is green — nothing is blocking merge.**

I verified your two fixes on `900d172f`:

- `MissingWithKeyAnalyzerTests` — **31/31 passing**
- The doc-app gate (`DocApps.proj` with `TreatWarningsAsErrors`) — **0 findings, build succeeded**

### My point 3 was wrong

I told you the guide being stale would fail a freshness gate. It doesn't. That check is scoped to exactly two version-bearing pages:

```yaml
- name: Verify docs version substitution is committed
  run: git diff --exit-code -- docs/guide/getting-started.md docs/guide/packaging.md
```

It guards `{{reactorVersion}}` substitution, not guide content. Its error message says *"docs/guide is out of date with the doc templates"*, which reads far broader than what it actually checks — I took the message at face value instead of reading the command. Sorry for sending you after a non-problem.

For the same reason, `Docs build` failing on your first commit wasn't a freshness check either: `docs compile` builds the doc apps in Phase 2, so it was tripping over the same 10 `DSL_001` errors. Once you keyed the samples it went green.

### The drift is real, but it's ours to clean up

Keying the samples changed the snippets embedded in the guide, so 10 pages now publish the old unkeyed code — including the `ForEach` in `docs/guide/hooks.md` that this PR's own analyzer would flag. Nothing in CI catches that today; I've added the details to #1052, which proposes widening the check to the whole generated tree.

**Nothing further needed from you.** I'll regenerate those pages in a follow-up PR, together with the other bits I'd already earmarked as ours — the symbol gate on the `ForEach` arm, a CHANGELOG entry, and two unkeyed sites in `samples/`.

Merging this now. Thanks for picking it up, and for turning the fixes around so quickly.

PR #1161, opened 2026-08-31
Attribute thin helpers to their callers and stamp converted arguments (spec 010)

azchohfi:
Layer 2 of spec 010. Builds on the per-element source mapping merged in #1147, and closes the two attribution gaps that drop documented.

Base is `main` - layer 1 was squash-merged as `33a35999`, so this branch was rebased onto `main` and the diff is only this work.

## The gaps

Measured on `main` with interception on:

| Expression | Outer element | Child element |
|---|---|---|
| `VStack("hi")` | correct | **`null`** |
| `VStack(TextBlock("hi"))` | correct | correct |
| helper `MyHeader() => TextBlock("header")` | n/a | **helper''s body line**, not the caller''s |

Neither is fixable the obvious way. Interceptors "can only occur for calls to ordinary member methods - not constructors, delegates, properties, local functions, **operators**", and the `implicit operator Element(string)` body is already compiled into `Reactor.dll`, so the bare-string child is structurally unreachable. And a helper''s body genuinely *is* the call site of `TextBlock`, so there is nothing incorrect to fix - only an intent the compiler cannot infer.

## Mechanism 1 - `[ReactorSourceTransparent]`

A public marker attribute meaning "attribute my results to my caller". The generator emits no interceptor for DSL calls lexically inside an annotated method (rule 1), and intercepts calls *to* it instead (rule 2). Recursion falls out: annotated calling annotated keeps deferring outward until a caller that is not annotated (rule 3).

```csharp
[ReactorSourceTransparent]
internal static Element Field(string label, string value)
    => HStack(TextBlock(label), TextBlock(value));

var form = VStack(
    Field("Name", name),    // reports THIS line
    Field("Email", email)); // and this one
```

**Deliberately opt-in.** For a `Component.Render()` body the body line is the correct answer, so this is never a blanket rule for helpers. `UnannotatedHelper_StillReportsItsOwnBodyLine` and `TwoCallSitesOfTheUnannotatedHelper_DoCollapse` pin that default in the same file, one attribute apart from the positive tests.

**A bad annotation is never worse than none.** Rule 1 is conditioned on the helper actually being *forwardable* (static, `Element`-returning, nameable from a generated file - so `public`/`internal`, not a local function, not in a `file`-local or generic type), not merely on it being annotated. Suppressing the body stamp with no rule-2 interceptor to replace it would trade "the helper''s line" for no line at all. Unusable annotations report **`REACTOR_SOURCEMAP_001`** rather than doing nothing silently, and are `#pragma`-suppressible.

Instance-method helpers are an explicit, documented gap: intercepting one requires an extension-method interceptor, a different shape.

## Mechanism 2 - argument-position stamping

The enclosing factory call *is* intercepted, so its interceptor stamps arguments that arrived via an implicit user-defined conversion to `Element`, using **each argument''s own line**:

```csharp
VStack(
    "a",   // reports this line
    "b");  // and this one
```

Works for any user-defined conversion, not just `string`. Respects first-stamp-wins, hands `EmptyElement.Instance` through without cloning it, and only ever writes into a `params` array the **compiler** synthesized for that call site - `VStack(myArray)` is left completely alone.

## Bonus: a latent layer-1 bug, fixed centrally

`ReturnsElement` compared `ToDisplayString()` against the metadata name. I measured with a standalone Roslyn 5.9 probe that **the default display format renders nullability**: a method returning exactly `Element?` renders as `...Element?`, misses the comparison, and - since `Element`''s base is `object` - the base-type walk goes straight past the answer. Such a method was silently skipped entirely.

Latent today (no factory returns bare `Element?`; a nullable *derived* return is unaffected, because `TextBlockElement?.BaseType` is unannotated). It stops being latent here, where a conditional `Element?` helper is a natural shape. Fixed at the root by comparing symbols - verified `SymbolEqualityComparer.Default.Equals(Element?, Element)` is `true` while `IncludeNullability` is `false`, which is the pair that proves Default ignores annotations rather than the two simply being the same symbol. `Signature.CanReturnEmpty` had the identical defect and is fixed with it, which matters now that `Element?`-returning methods are reachable at all.

**No behaviour change:** the generated interceptor file for `Reactor.SourceMap.Tests` is **byte-identical** before and after this fix (79,946 bytes, same 90 attribute applications) - a direct measurement rather than an argument from counting declarations.

## `PendingFactory.Pending` is now annotated

It is the textbook thin forwarder, and it is the **only** end-to-end proof that rule 2 reads the attribute from *metadata* rather than same-compilation source - the shipping shape. `TransparentAnnotationInAReferencedAssembly_IsHonoured` asserts the stamped **line**, not merely non-null: a null check alone would also pass if the inner `Component<,>` call had stamped it, pointing into Reactor''s own source. It also depends on the attribute being `public`; an internal one would be invisible across the boundary while every in-compilation test stayed green.

`NonFactoriesEntryPointTests` (which explicitly anticipated this change) now pins the gap against `PropertyGridDefaults.PropertyLabelTemplate` instead - still a real, unannotated, public entry point whose element is built inside Reactor. Docs updated in the same commit.

## Mutation evidence

Every mechanism was mutated in the generator and the suites re-run. No mutation was accepted until it reddened something.

| Mutation | Tests reddened |
|---|---|
| Rule 1 disabled | **7** |
| Rule 2 disabled (`isTransparentTarget` dropped) | **9** |
| Rule 1 applied unconditionally (ignore forwardability) | **1** - `PrivateAnnotatedHelper_BehavesExactlyLikeAnUnannotatedOne` |
| Argument stamping not emitted | **11** |
| Argument first-stamp-wins guard removed | **1** - `OperatorThatReturnsAStampedElement_KeepsTheOperatorsLine` |
| Argument `EmptyElement` guard removed | **1** - `ConversionThatYieldsTheEmptySingleton_IsNotCloned` |
| Argument line -> container''s line | **3** |
| `ParamArray` widened to `Explicit` | **3** |
| Normal-form guards dropped | **1** - `ExplicitlyWrittenArrayArgument_IsAlsoLeftAlone` |
| `ReturnsElement` reverted to name comparison | **2** (one live, one driver) |

Two things this process caught in my own work, both worth stating because they are the failure mode this repo warns about:

- My **first attempt at the argument-line mutation survived**, and the mutation was wrong, not the tests: `FirstAncestorOrSelf<InvocationExpressionSyntax>()` returns the argument''s *own* invocation for `At("x", out var l)`. Re-run against the enclosing `ArgumentListSyntax` it reddens 3 tests.
- `ExplicitlyWrittenArrayArgument_IsAlsoLeftAlone` was **added because the original normal-form test proved nothing under mutation** - its array is built in a separate statement, so no plausible mutation reaches it. The inline-array form is what actually distinguishes "the compiler built this array" from "the argument is an array creation".

The `REACTOR_SOURCEMAP_001` diagnostic got the same treatment: removing the `#pragma` produces the warning at the attribute (`TransparentHelperTests.cs(83,6)`) with the full message, so the green build is suppression working rather than the diagnostic never firing.

## Vacuity controls

- Every line assertion is checked against a `[CallerLineNumber]` oracle captured at the relevant physical line. `At(value, out int line)` puts that oracle *inside* an argument list, so multi-argument calls are checked argument-by-argument with no hard-coded offsets.
- `EachConvertedArgument_TakesItsOwnLine` asserts `firstLine != secondLine` before using them, so a degenerate oracle cannot make the comparison a tautology.
- `OperatorThatReturnsAStampedElement_KeepsTheOperatorsLine` is a differential oracle - the same conversion from two different lines must agree with each other - so no literal line number is written down.
- The driver suite is hermetic, so `AttributeMetadataName_MatchesTheRealAttribute` ties it to the shipping type; a rename would otherwise leave it green while the feature broke for every consumer.
- The driver harness asserts its own probe snippet compiles. That caught the `Microsoft.UI.System` shadowing trap immediately instead of as a mysterious "expected diagnostic missing".
- My `InterceptorCount` helper was itself wrong at first - it counted the emitted `InterceptsLocationAttribute` polyfill''s own declaration. Exact-count assertions caught it.

## Numbers

| | Baseline (`main`) | This branch |
|---|---|---|
| `Reactor.Tests` | 13754 total, 0 failed, 64 skipped | **13768 total, 0 failed, 64 skipped** (+14) |
| `Reactor.SourceMap.Tests` | 42 / 42 | **71 / 71** (+29) |
| `dotnet build Reactor.slnx -c Release` | 0 warnings, 3 `NU1301` | **0 warnings, 3 `NU1301`** (unchanged - nuget.org unreachable here, pre-existing in the `vs-reactor` projects) |

`Reactor.SourceMap.Tests` also builds clean in Release specifically (`0 warnings, 0 errors`), where warnings are errors - and its Release-generated file contains 106 interceptors and 16 argument stamps, so that clean build is the generator running and the pragma suppressing, not the generator being off.

## Also included

- Docs: `docs/_pipeline/templates/source-mapping.md.dt` (the source) recompiled via `mur docs compile --topic source-mapping`. The stale "bare-string children are not stamped" and "helper methods attribute to themselves" limitations are replaced with the new sections; `Pending.md` regenerates from the new XML doc.
- Public API index regenerated into both byte-identical copies.
- `CHANGELOG.md` entry under `Unreleased`/`Added`, cross-referencing spec 010 and ordered by spec number.
- A symbol-only pre-check gates the `GetOperation` call so only factories with an `Element`-typed parameter build an operation tree, keeping per-call-site generation cost where layer 1 measured it.

## Not addressed

`sig.ReturnsNullable`''s emitted `if (__e is null)` arm could previously only be reached through a nullable *derived* return type, for the same root cause fixed here. No machinery added for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,31 @@ Conventions for contributors:
   `ReactorSourceMap.Enabled` switch, which the devtools verb sets automatically. When
   the runtime switch is off, no locations are populated; in Release, the generator is
   not loaded unless the MSBuild property is explicitly enabled.
+- **`[ReactorSourceTransparent]` — attribute a thin helper to its caller, and per-argument
+  source mapping for implicitly converted children (spec 010).** Two follow-ups that close
+  the attribution gaps the initial source-mapping drop documented.
+
+  Marking a static, `Element`-returning helper
+  `[Microsoft.UI.Reactor.Diagnostics.ReactorSourceTransparent]` stops the generator
+  stamping DSL calls inside it and instead stamps calls *to* it, so a forwarder like
+  `Field(label, value)` reports each caller's own line rather than collapsing every call
+  site onto one line in its body. Annotated helpers compose: the deferral walks outward
+  until it reaches a caller that is not annotated. It is deliberately opt-in — for a
+  `Component.Render()` body the body line is the correct answer. An annotation the
+  generator cannot honour (a `private` helper, an instance method, a local function)
+  reports **`REACTOR_SOURCEMAP_001`** and leaves attribution exactly as it would be
+  without the attribute, so a bad annotation is never worse than none. The attribute is
+  read from metadata, so libraries can annotate their own forwarders;
+  `PendingFactory.Pending` now does, and its callers get a location where they
+  previously got `null`.
+
+  Separately, arguments that reach an `Element` parameter through an implicit
+  user-defined conversion — `VStack("a", "b")`, via `implicit operator Element(string)` —
+  are now stamped at **the argument's own line**. Those elements are built by an operator
+  body inside Reactor, which interceptors structurally cannot reach (they intercept
+  ordinary method calls, never operators), so they previously reported no location at
+  all. Applies to any user-defined conversion to `Element`, respects first-stamp-wins,
+  and never writes into a `params` array the caller owns.
 - **`NavigationTransition.Entrance()` — the WinUI page-refresh motion as a first-class
   transition (spec 011 §6).** The incoming page slides up a short distance and fades in,
   mirroring WinUI's `EntranceNavigationTransitionInfo`. `EntranceTransition` is public and
--- a/docs/_pipeline/templates/source-mapping.md.dt
+++ b/docs/_pipeline/templates/source-mapping.md.dt
@@ -173,20 +173,11 @@ is exactly where the author wrote the UI, and deferring it to the caller would b
 a regression.
 
 For a *thin forwarder*, whose own line carries no information anyone wants, mark
-it `[ReactorSourceTransparent]`:
+it `[ReactorSourceTransparent]`. This is the helper the source-map explorer
+sample uses for its two order rows — click them in the running app and they
+report their own distinct call sites, not this helper's body:
 
-```csharp
-using Microsoft.UI.Reactor.Diagnostics;
-
-[ReactorSourceTransparent]
-internal static Element Field(string label, string value)
-    => HStack(TextBlock(label), TextBlock(value));
-
-// Each row now reports ITS OWN line, instead of both reporting
-// the HStack( line inside Field.
-var form = VStack(
-    Field("Name", name),
-    Field("Email", email));
+```csharp snippet="source:samples/apps/source-map-explorer/App.cs#transparent-helper"
 ```
 
 Two rules do the work, and they compose. The generator emits no interceptor for
@@ -196,6 +187,11 @@ therefore keeps deferring outward until it reaches a caller that is not
 annotated. First-stamp-wins still applies, so a helper that merely passes an
 element through does not relabel it.
 
+Because rule 1 suppresses stamping for *everything* inside an annotated method,
+that includes argument-position stamps (below): a bare-string child written
+inside a transparent helper is not stamped either, since the element belongs to
+whoever called the helper.
+
 The annotated method has to be one the generator can emit a forwarding call to:
 
 | Requirement | Why |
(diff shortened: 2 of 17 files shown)


azchohfi:
Requesting a GitHub Copilot review.

PR #1181, opened 2026-09-05
fix(tray): avoid duplicate mouse notifications

### LINKED ISSUE #1180: [Bug] ReactorTrayIcon forwards duplicate mouse notifications
#### Expected behavior

Each physical tray interaction should invoke exactly one corresponding event: Click, RightClick, or DoubleClick.

#### Actual behavior

With one ReactorTrayIcon and one handler attached to each event, a single left click, right click, or double click invokes the corresponding handler twice. No exception is thrown.

The suspected failure point is TrayHiddenWindow.DispatchCallback: it forwards every callback notification without deduplicating equivalent notifications. The implementation maps WM_LBUTTONUP/NIN_SELECT/NIN_KEYSELECT to Click, WM_LBUTTONDBLCLK to DoubleClick, and WM_RBUTTONUP/WM_CONTEXTMENU to RightClick.

### Steps to reproduce

1. Create an unpackaged .NET 10 WinUI 3 application referencing Microsoft.UI.Reactor 0.1.0-preview.13.
2. Place a valid test.ico next to the executable and run:

```csharp
using Microsoft.UI.Reactor;

var clicks = 0;
var rightClicks = 0;
var doubleClicks = 0;

ReactorApp.ShutdownPolicy = ShutdownPolicy.Explicit;
ReactorApp.Run(_ =>
{
    var tray = ReactorApp.OpenTrayIcon(new TrayIconSpec(
        WindowIcon.FromPath(Path.Combine(AppContext.BaseDirectory, "test.ico")),
        "Test",
        new WindowKey("tray")));

    tray.Click += (_, _) => clicks++;
    tray.RightClick += (_, _) => rightClicks++;
    tray.DoubleClick += (_, _) => doubleClicks++;
});
```

3. Perform one left click, one right click, and one double click on the tray icon.
4. Observe that the corresponding handler/counter is invoked twice for a single gesture instead of once.

### Reactor version / commit

Microsoft.UI.Reactor 0.1.0-preview.13 (NuGet; repository commit c9191b97c40a2e4d6bcbc72df7714184862b4d36)

### Platform

ARM64

### .NET SDK version

10.0.400

### Windows version

Windows 11, build 26100 (ARM64)

### Windows App SDK version

2.3.6

### Logs / stack trace

```text
No exception is thrown; the failure is duplicate callback delivery.
```

### Confirmation

- [ ] This bug reproduces against the current main branch (or I've noted otherwise above).

kavyansh18:
## Summary

Fixes duplicate tray icon mouse event notifications when using `NOTIFYICON_VERSION_4`.

`TrayHiddenWindow.DispatchCallback` currently handles both legacy mouse messages and the corresponding semantic notifications, causing a single tray interaction to trigger `Click` or `RightClick` more than once. This change removes the duplicate legacy message handling while preserving the existing semantic notifications and double-click behavior.

## Linked issue / spec

Fixes #1180

## Test plan

- Reviewed the tray callback dispatch path and the `NOTIFYICON_VERSION_4` notification handling.
- Removed `WM_LBUTTONUP` from `Click` dispatch and `WM_RBUTTONUP` from `RightClick` dispatch.
- Verified that `NIN_SELECT` and `NIN_KEYSELECT` remain mapped to `Click`.
- Verified that `WM_CONTEXTMENU` remains mapped to `RightClick`.
- Verified that `WM_LBUTTONDBLCLK` remains mapped to `DoubleClick`.
- Confirmed the final change is limited to `src/Reactor/Hosting/Shell/TrayHiddenWindow.cs`.
- Local Windows/.NET tests could not be run because development is being performed on macOS without the .NET SDK and Windows App SDK environment.

## Risk / breaking changes

No public API changes.

This is a localized change to tray icon callback dispatch. Legacy `WM_LBUTTONUP` and `WM_RBUTTONUP` messages are no longer treated as separate event triggers when the corresponding `NOTIFYICON_VERSION_4` semantic notifications are received.

Existing keyboard selection, right-click context-menu, and double-click handling are preserved.

### DIFF
--- a/src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
+++ b/src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
@@ -190,9 +190,13 @@ internal sealed class TrayHiddenWindow : IDisposable
         {
             try
             {
+                // NOTIFYICON_VERSION_4 wire protocol:
+                // Explorer forwards both the legacy mouse message (WM_LBUTTONUP / WM_RBUTTONUP)
+                // and the version-4 semantic notification (NIN_SELECT / WM_CONTEXTMENU) for each
+                // physical user interaction. Handling both duplicates Click and RightClick (issue #1180).
+                // We handle only the v4 semantic notifications + WM_LBUTTONDBLCLK.
                 switch (mouseMessage)
                 {
-                    case TrayIconComInterop.WM_LBUTTONUP:
                     case TrayIconComInterop.NIN_SELECT:
                     case TrayIconComInterop.NIN_KEYSELECT:
                         hit.OnClick?.Invoke();
@@ -200,7 +204,6 @@ internal sealed class TrayHiddenWindow : IDisposable
                     case TrayIconComInterop.WM_LBUTTONDBLCLK:
                         hit.OnDoubleClick?.Invoke();
                         break;
-                    case TrayIconComInterop.WM_RBUTTONUP:
                     case TrayIconComInterop.WM_CONTEXTMENU:
                         hit.OnRightClick?.Invoke();
                         break;


kavyansh18:
hey @codemonkeychris can you please review the PR

src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
```diff
@@ -190,17 +190,20 @@ private void DispatchCallback(uint id, uint mouseMessage)
         {
             try
             {
+                // NOTIFYICON_VERSION_4 wire protocol:
+                // Explorer forwards both the legacy mouse message (WM_LBUTTONUP / WM_RBUTTONUP)
+                // and the version-4 semantic notification (NIN_SELECT / WM_CONTEXTMENU) for each
+                // physical user interaction. Handling both duplicates Click and RightClick (issue #1180).
+                // We handle only the v4 semantic notifications + WM_LBUTTONDBLCLK.
```
azchohfi, line 197: Nice catch on the root cause — this is the right fix.

One thing that would make it bulletproof: there's currently no test holding this in place, so if someone re-adds the two `case` arms later the whole suite still passes. The only tray selftest, `WindowModel_TrayIconRoundTrip` in `tests/Reactor.AppTests.Host/SelfTest/Fixtures/WindowModelFixtures.cs`, covers registration, tooltip, visibility and close — it never feeds a callback notification or counts handler invocations.

If you're up for it, a fixture that registers a `TrayCallbackEntry`, feeds `WM_LBUTTONUP` + `NIN_SELECT` for one interaction and `WM_RBUTTONUP` + `WM_CONTEXTMENU` for another, then asserts each counter is exactly 1 would do it. The counter is the important bit — a "didn't throw" assertion wouldn't catch the regression.

Fair warning that it's a bit fiddly: `DispatchCallback` is private, instance-bound, marshals through the `DispatcherQueue`, and constructing the window creates a real HWND. Extracting the switch into an internal pure classifier like `ClassifyNotification(uint) -> TrayCallbackKind` would make it testable from the headless `Reactor.Tests` tier, but that's a bigger change than your fix deserves.

**So genuinely no pressure here** — if you'd rather keep this PR tight, we're happy to merge as-is and add the coverage ourselves in a follow-up. Totally your call.

src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
```diff
@@ -190,17 +190,20 @@ private void DispatchCallback(uint id, uint mouseMessage)
...
                 switch (mouseMessage)
                 {
-                    case TrayIconComInterop.WM_LBUTTONUP:
                     case TrayIconComInterop.NIN_SELECT:
                     case TrayIconComInterop.NIN_KEYSELECT:
                         hit.OnClick?.Invoke();
                         break;
                     case TrayIconComInterop.WM_LBUTTONDBLCLK:
```
azchohfi, line 204: Small thing, and not a change request — just so we don't close #1180 prematurely.

The issue lists three duplicated events (Click, RightClick **and** DoubleClick) and this handles the first two. I went looking for a DoubleClick bug and couldn't find one: this is the only `OnDoubleClick` sink, `Register` dedups by icon id, and the `TaskbarCreated` path only calls `OnReapply` without re-registering. So I'm fairly confident your code is correct as-is and nothing more is needed here.

My hunch is the reporter saw `Click` fire once (from `NIN_SELECT`) alongside `DoubleClick` on a double click, which is expected behavior rather than a duplicate. If you happen to have the repro handy it'd be great to confirm, but no need to go set one up — we can verify on our side and note it on the issue before closing.

azchohfi (approved):

src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
kavyansh18: Thanks for the suggestion. Since the callback is tightly coupled to the native window/dispatcher flow, I’ll leave the test coverage for a follow-up and keep this PR focused on the fix.

src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
kavyansh18: Thanks for checking this. Since WM_LBUTTONDBLCLK is already the dedicated DoubleClick path, I’ll leave it unchanged and keep this PR focused on the Click/RightClick duplication.

src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
azchohfi: Totally fair — that's the right call for this PR. We'll pick up the regression test and the CHANGELOG entry in a follow-up on our side. Thanks for the fix!

src/Reactor/Hosting/Shell/TrayHiddenWindow.cs
azchohfi: Agreed, leaving it unchanged is correct. We'll verify the double-click sequence against this branch and note the finding on #1180 so it doesn't get closed on the wrong conclusion. Thanks again!

PR #1182, opened 2026-09-06
Fix localization extraction of icon glyphs

### LINKED ISSUE #1132: mur loc extract treats Segoe Fluent icon glyphs as translatable strings
## Summary

`mur loc extract` treats Segoe Fluent icon glyphs as user-visible text and extracts them into `.resw`. With `--rewrite` they are routed through `t.Message(...)`, making a decorative icon depend on a resource lookup — and making it eligible for AI translation.

## Repro

`samples/TodoApp/Program.cs` uses Private Use Area glyphs from the Segoe Fluent icon font as button/label content:

```csharp
Button("\uE74D", () => structural(new DeleteItem(item.Id)))   // trash can
    .FontFamily((FontFamily)Application.Current.Resources["SymbolThemeFontFamily"])

TextBlock("\uE73A")                                           // decorative checkbox
```

```
mur loc extract --source samples/TodoApp --output <out> --rewrite
```

Produces:

```xml
<data name="Text"  xml:space="preserve"><value>&#xE74D;</value></data>
<data name="Text2" xml:space="preserve"><value>&#xE73A;</value></data>
```

Verified code points (they are not empty strings — they render as blanks in most terminals):

```
Text   len=1  U+E74D
Text2  len=1  U+E73A
```

And rewrites the source to:

```csharp
Button(t.Message(Loc.TodoApp.Text), ...)
TextBlock(t.Message(Loc.TodoApp.Text2))
```

## Why this is wrong

1. **A PUA glyph is not linguistic content.** There is nothing to translate. `mur loc translate` would hand `U+E74D` to an LLM and ask for a French version.
2. **It's a translation-corruption risk.** A machine or human translation pass over a `.resw` containing a lone PUA character can easily mangle or drop it, silently breaking the icon.
3. **It converts a constant into a lookup.** After `--rewrite` the glyph resolves through `IStringResourceProvider`; a missing key in some locale degrades or breaks the icon rather than falling back to a glyph that never should have varied.
4. **Spec 005 §10.3 says it shouldn't happen.** The extractor is documented to ignore "strings that are clearly not user-visible."
5. **The generated keys are meaningless.** `Text` / `Text2` carry no information about what the glyph is.

## Suggested fix

Skip string literals that consist wholly of Private Use Area code points (U+E000–U+F8FF, plus the supplementary PUA planes) — optionally only when the call site also sets a symbol font, though the PUA test alone should be safe and is much simpler.

A stronger version of the same idea: skip any literal containing no letter or digit in any script, which also covers punctuation-only and separator strings.

## Notes

Not a systematic false-positive: the `samples/ReactorGallery` run produced 561 keys with **0** empty or glyph-only values. This is specific to the icon-glyph idiom, which the Gallery happens not to use in extractable positions — but the idiom is common in real WinUI apps and is used in the repo's own sample.

Reproduced with `mur` built from `88ef6db4`.

jibin7jose:
## Summary
Fixes #1132

Fixes localization extraction of Segoe Fluent icon glyphs.

## Changes

- Skip string literals containing only Private Use Area code points.
- Support BMP and supplementary Private Use Area ranges.
- Preserve mixed icon and text strings for translation.
- Add regression coverage for icon-only literals.

This prevents `mur loc extract --rewrite` from routing decorative icon glyphs through `t.Message(...)`.

### DIFF
--- a/src/Reactor.Cli/Loc/LocalizableStringScanner.cs
+++ b/src/Reactor.Cli/Loc/LocalizableStringScanner.cs
@@ -220,7 +220,7 @@ internal static class LocalizableStringScanner
         private void ProcessStringLiteral(LiteralExpressionSyntax literal, string className, string context)
         {
             var value = literal.Token.ValueText;
-            if (string.IsNullOrWhiteSpace(value)) return;
+            if (string.IsNullOrWhiteSpace(value) || IsPrivateUseOnly(value)) return;
 
             _results.Add(new LocalizableString
             {
@@ -231,6 +231,28 @@ internal static class LocalizableStringScanner
                 SpanStart = literal.SpanStart,
                 SpanLength = literal.Span.Length,
             });
+        }
+
+        private static bool IsPrivateUseOnly(string value)
+        {
+            if (value.Length == 0) return false;
+
+            for (var i = 0; i < value.Length; i++)
+            {
+                var codePoint = (int)value[i];
+                if (char.IsHighSurrogate(value[i]) && i + 1 < value.Length && char.IsLowSurrogate(value[i + 1]))
+                {
+                    codePoint = char.ConvertToUtf32(value[i], value[++i]);
+                }
+
+                var isPrivateUse = (codePoint >= 0xE000 && codePoint <= 0xF8FF)
+                    || (codePoint >= 0xF0000 && codePoint <= 0xFFFFD)
+                    || (codePoint >= 0x100000 && codePoint <= 0x10FFFD);
+                if (!isPrivateUse)
+                    return false;
+            }
+
+            return true;
         }
 
         private void ProcessInterpolatedString(InterpolatedStringExpressionSyntax interpolated, string className, string context)
--- a/tests/Reactor.Tests/Localization/LocalizableStringScannerTests.cs
+++ b/tests/Reactor.Tests/Localization/LocalizableStringScannerTests.cs
@@ -293,6 +293,27 @@ public class LocalizableStringScannerTests
     }
 
     [Fact]
+    public void PrivateUseOnlyString_Skipped()
+    {
+        var source = """
+            class App : Component {
+                public override Element Render() {
+                    return VStack(
+                        TextBlock("\uE74D"),
+                        TextBlock("\uE73A"),
+                        TextBlock("\U000F0000"),
+                        TextBlock("\uE74D Delete"));
+                }
+            }
+            """;
+
+        var results = Scan(source);
+
+        Assert.Single(results);
+        Assert.Equal("\uE74D Delete", results[0].Value);
+    }
+
+    [Fact]
     public void MultipleElementsInMethod_AllDetected()
     {
         var source = """


src/Reactor.Cli/Loc/LocalizableStringScanner.cs
```diff
@@ -220,7 +220,7 @@ private void ProcessExpression(ExpressionSyntax expr, string className, string c
         private void ProcessStringLiteral(LiteralExpressionSyntax literal, string className, string context)
         {
             var value = literal.Token.ValueText;
-            if (string.IsNullOrWhiteSpace(value)) return;
+            if (string.IsNullOrWhiteSpace(value) || IsPrivateUseOnly(value)) return;
```
azchohfi, line 223: Thanks for picking this up — glyphs leaking into the resource files is a real papercut.

One gap: this only filters the direct-literal path. `ProcessTernaryBranch` has its own copy of the literal extraction (lines 293-306) that still checks only `IsNullOrWhiteSpace`, so `TextBlock(cond ? "\uE70D" : "\uE70E")` still extracts both glyphs — the exact case this PR is fixing.

Ternaries are a live path, not hypothetical: `Ternary_BothBranchesExtracted` covers it, and there's a real instance at `samples/ReactorCharting.Gallery/Program.cs:127`.

Interpolated strings are unfiltered too — `InterpolationConverter.Convert` returns any non-empty result, so `$"\uE74D"` survives. Much rarer, so I'm fine either way, but worth deciding on explicitly rather than leaving it implicit.

Probably cleanest to pull the literal extraction into one helper that both call. One gotcha: `ProcessTernaryBranch` sets `TernaryBranch = branchIndex`, and `KeyNamer` uses that to build branch-specific keys — so the shared helper needs to take it as an optional parameter. Routing ternary literals straight through `ProcessStringLiteral` would quietly regress key naming.

src/Reactor.Cli/Loc/LocalizableStringScanner.cs
```diff
@@ -233,6 +233,28 @@ private void ProcessStringLiteral(LiteralExpressionSyntax literal, string classN
...
+            {
+                var codePoint = (int)value[i];
+                if (char.IsHighSurrogate(value[i]) && i + 1 < value.Length && char.IsLowSurrogate(value[i + 1]))
+                {
+                    codePoint = char.ConvertToUtf32(value[i], value[++i]);
+                }
+
+                var isPrivateUse = (codePoint >= 0xE000 && codePoint <= 0xF8FF)
+                    || (codePoint >= 0xF0000 && codePoint <= 0xFFFFD)
+                    || (codePoint >= 0x100000 && codePoint <= 0x10FFFD);
```
azchohfi, line 250: These ranges check out — I verified them against the Unicode database and `E000–F8FF` / `F0000–FFFFD` / `100000–10FFFD` match category `Co` exactly, upper bounds included.

Since we're on net10.0 though, `Rune` will do the whole thing for us:

```csharp
value.EnumerateRunes().All(r => Rune.GetUnicodeCategory(r) == UnicodeCategory.PrivateUse)
```

That hands surrogate decoding, malformed-UTF-16 handling, and the category data to the BCL instead of us carrying three hardcoded ranges plus a manual surrogate walk. Not a bug — just less for us to maintain as Unicode moves.

tests/Reactor.Tests/Localization/LocalizableStringScannerTests.cs
```diff
@@ -292,6 +292,27 @@ public override Element Render() {
...
+    public void PrivateUseOnlyString_Skipped()
+    {
+        var source = """
+            class App : Component {
+                public override Element Render() {
+                    return VStack(
+                        TextBlock("\uE74D"),
+                        TextBlock("\uE73A"),
+                        TextBlock("\U000F0000"),
+                        TextBlock("\uE74D Delete"));
```
azchohfi, line 305: Thanks for including a test. It pins both directions — drop the filter and you get 4 results, force it to always return true and you get 0 — so it's non-vacuous either way.

Two things would make it airtight. `\U000F0000` is the low edge of plane 15, and there's nothing for plane 16 (`100000–10FFFD`) at all, so that third range is currently untested — and no boundary cases either. A `[Theory]` over `U+E000`, `U+F8FF`, `U+F900`, `U+FFFFD`, `U+FFFFE`, `U+10FFFD`, `U+10FFFE` would lock the edges down and localize a failure better than one bundled `[Fact]`.

A lone-surrogate case would be worth adding too — `"\uD800"` on its own is a legal C# literal, and it currently falls through as non-PUA and gets extracted. I think that's the right call, but let's pin it so it's intentional rather than incidental.

src/Reactor.Cli/Loc/LocalizableStringScanner.cs
azchohfi: Adopted in `ea779d19` — now `value.EnumerateRunes().All(rune => Rune.GetUnicodeCategory(rune) == UnicodeCategory.PrivateUse)`, with a `Length > 0` guard so an empty string doesn't pass vacuously through `All`.

I checked the swap is genuinely behavior-preserving rather than assuming it: ran the classification against the real .NET runtime and every boundary matches the old hand-rolled ranges. The lone-surrogate case is a real improvement — `EnumerateRunes()` maps ill-formed UTF-16 to U+FFFD (`OtherSymbol`), so it degrades gracefully instead of depending on the manual surrogate walk.

src/Reactor.Cli/Loc/LocalizableStringScanner.cs
azchohfi: Fixed in `ea779d19`. `ProcessTernaryBranch` now delegates to `ProcessStringLiteral` / `ProcessInterpolatedString` with an optional `int? ternaryBranch`, and both share the new `IsNonLocalizable` gate — so the ternary and interpolated paths are covered along with plain literals.

The `TernaryBranch` hazard I flagged is handled correctly: the field is `int?`, so the `= null` default preserves the old behavior for non-ternary callers, and `PrivateUseOnlyStrings_SkippedAcrossExpressionForms` asserts `Value == "Visible" && TernaryBranch == 1` — a direct regression test for exactly that. `Ternary_BothBranchesExtracted` still passes too.

Verified locally: 30/30 scanner tests and 293/293 localization tests green, including `ExtractIntegrationTests` (my regression risk, since interpolated strings now get whitespace filtering they didn't have before). Nice cleanup as a bonus — the duplicated extraction block and the unused `parent` parameter are both gone.

tests/Reactor.Tests/Localization/LocalizableStringScannerTests.cs
azchohfi: Covered in `ea779d19` — the new `[Theory]` hits all seven boundaries plus the lone surrogate.

I didn't take the expectations on faith; I ran the same classification against the live .NET runtime and all 8 rows match, with `U+F900` / `U+FFFFE` / `U+10FFFE` coming back extractable as a positive control that the check can produce both outcomes. `\uD800` becomes U+FFFD and is extracted, exactly as the row asserts.

Splitting into a `[Theory]` also means a future Unicode/BCL change fails on one specific row instead of one bundled `[Fact]`.

azchohfi (approved):
Thanks for the follow-up commit — all three review threads are addressed and resolved.

Verified locally at `ea779d19`:

- Build clean (0 warnings, 0 errors)
- **30/30** `LocalizableStringScannerTests`
- **293/293** localization tests, including `ExtractIntegrationTests` — the regression I was watching for, since interpolated strings now get whitespace filtering they previously didn't

I also ran the PUA classification against the live .NET runtime rather than trusting the table: all 8 `[Theory]` rows match, with `U+F900` / `U+FFFFE` / `U+10FFFE` extractable as a positive control that the check can produce both outcomes.

The scanner ends up simpler than before the PR — shared `IsNonLocalizable` gate, duplicated ternary extraction block gone, unused `parent` parameter gone.

Two non-blocking doc items remain, which I will send as a follow-up PR rather than hold this one:

- `CHANGELOG.md` `[Unreleased]` -> `Fixed` has no entry for this user-visible `mur loc extract` change
- `docs/specs/005-localization-design.md` and `docs/reference/localization-ci.md` still describe extraction as capturing "bare string literals in DSL calls", with no mention of the private-use exclusion

PR #1184, opened 2026-09-07
Fix localization extraction for plural ternaries

### LINKED ISSUE #1131: mur loc extract turns C# ternary plurals into unlocalizable ICU
## Summary

`mur loc extract` converts a C# ternary used for pluralization into an ICU `select` keyed on an English boolean. The result renders correctly in English but **cannot be translated into any language whose plural rules differ from English** — which defeats the purpose of the tool.

## Repro

`samples/TodoApp/Program.cs:254`:

```csharp
Caption($"{remaining} item{(remaining == 1 ? "" : "s")} left")
```

```
mur loc extract --source samples/TodoApp --output <out>
```

Extracted:

```xml
<data name="ItemFalseLeft" xml:space="preserve">
  <value>{remaining} item{arg1, select, true {} false {s}} left</value>
  <comment>auto-extracted from interpolation</comment>
</data>
```

## Why this is wrong

The message is *valid* — I verified it against the real runtime engine (`Jeffijoe.MessageFormat` 8.0.0, the ICU engine per spec 005 §11.1) and it formats fine in English:

| | `n=1` | `n=3` | `n=5` |
|---|---|---|---|
| what `mur` generated | `1 item left` | `3 items left` | `5 items left` |

The defect is structural. The `one` vs `not-one` decision was **baked into C# at extract time**, and the only branches exposed to a translator are `true` and `false`. Polish needs three plural forms; there is no branch to write the third in:

| | `n=1` | `n=3` | `n=5` |
|---|---|---|---|
| correct ICU `plural` (pl-PL) | `1 element` | `3 elementy` | `5 elementów` |
| the shape `mur` produced | *(no way to express this)* | | |

The correct extraction is a `plural`, not a `select`:

```
{remaining, plural, one {# item} other {# items}} left
```

`plural` delegates the category choice to CLDR at format time, so a translator can add `few`/`many`. `select` on a boolean freezes English grammar into the resource.

## Related problems in the same output

- **Key name.** `ItemFalseLeft` is derived from message content including the literal `false` from the ternary.
- **Placeholder name.** `arg1` is opaque to a translator. 27 of 561 values extracted from `samples/ReactorGallery` carry `arg{n}`-style placeholders.
- **No warning fired.** Spec 005 §10.4 says the CLI "adds a comment hint (`consider adding plural support`) when it detects a variable named `count`, `total`, `num*`, or similar quantity-suggesting names." `remaining` isn't in that list, so nothing warned — the broken message was emitted silently.
- **`mur loc validate` passes it.** Its stated job is "Check ICU syntax and parameter consistency"; the message is syntactically valid, so it reports `Validation passed: no errors or warnings.` Nothing in the pipeline catches this.

## Suggested fix

Detect the `<expr> ? "" : "s"` / `? "s" : ""` ternary shape (and near-variants) during interpolation conversion and emit a `plural` keyed on the numeric variable, rather than a `select` on the boolean. Where the shape can't be recognised confidently, prefer the existing "skip with a warning" path over emitting a message that is silently unlocalizable — spec 005 §10.4 already establishes that convention for complex expressions.

Worth widening the quantity-name heuristic too (`remaining`, `left`, `items`, …), though a shape-based rule is more reliable than a name-based one.

## Environment

Reproduced with `mur` built from `88ef6db4`.

jibin7jose:
Fixes #1131

## Summary

Convert C# numeric singular/plural ternaries into ICU `plural` messages instead of boolean `select` messages.

## Changes

- Detect `count == 1 ? "" : "s"` and reversed `!= 1` patterns.
- Generate ICU plural syntax such as:
  `{remaining, plural, one {# item} other {# items}}`
- Support full-word branches such as `"item"` / `"items"`.
- Preserve regular boolean ternaries as ICU `select`.
- Add regression tests for the supported patterns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ Conventions for contributors:
 - Localization extraction now converts recognized count-based singular/plural ternaries
   into ICU plural messages (spec 005 Â§10.4, #1131).
 - Localization extraction normalizes boolean select arguments to the string keys expected
-  by ICU MessageFormat when rewriting source (spec 005 Â§10.4, #1131).
+  by ICU MessageFormat when rewriting source (spec 005 §10.4, #1131).
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ Conventions for contributors:
 
 ### Changed
 
+- Localization extraction now converts recognized count-based singular/plural ternaries
+  into ICU plural messages (spec 005 Â§10.4, #1131).
+- Localization extraction normalizes boolean select arguments to the string keys expected
+  by ICU MessageFormat when rewriting source (spec 005 Â§10.4, #1131).
+
 ### Deprecated
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ Conventions for contributors:
 ### Changed
 
 - Localization extraction now converts recognized count-based singular/plural ternaries
-  into ICU plural messages (spec 005 Â§10.4, #1131).
+  into ICU plural messages (spec 005 §10.4, #1131).
 - Localization extraction normalizes boolean select arguments to the string keys expected
   by ICU MessageFormat when rewriting source (spec 005 §10.4, #1131).
 
--- a/src/Reactor.Cli/Loc/InterpolationConverter.cs
+++ b/src/Reactor.Cli/Loc/InterpolationConverter.cs
@@ -37,6 +37,7 @@ internal static class InterpolationConverter
     private static readonly HashSet<string> QuantityHintNames = new(StringComparer.OrdinalIgnoreCase)
     {
         "count", "total", "length", "size", "amount", "quantity",
+        "remaining",
     };
 
     public static (string? icuMessage, Dictionary<string, string>? argumentMap, List<string> warnings)
@@ -46,6 +47,7 @@ internal static class InterpolationConverter
         var argumentMap = new Dictionary<string, string>();
         var warnings = new List<string>();
         var usedNames = new HashSet<string>();
+        var emittedHoleNames = new Dictionary<int, string>();
 
         for (var contentIndex = 0; contentIndex < interpolated.Contents.Count; contentIndex++)
         {
@@ -65,13 +67,13 @@ internal static class InterpolationConverter
                         expr = parens.Expression;
 
                     // Ternary with string-literal branches → ICU select
-                    if (TryConvertPluralSuffix(interpolated, contentIndex, expr, icuParts, usedNames, argumentMap, out var suffixPluralIcu))
+                    if (TryConvertPluralSuffix(interpolated, contentIndex, expr, icuParts, emittedHoleNames, warnings, out var suffixPluralIcu, out var suppressStandalonePlural))
                     {
                         icuParts.Add(suffixPluralIcu!);
                         break;
                     }
 
-                    if (TryConvertTernaryPlural(expr, usedNames, argumentMap, out var pluralIcu))
+                    if (!suppressStandalonePlural && TryConvertTernaryPlural(expr, usedNames, argumentMap, out var pluralIcu))
                     {
                         icuParts.Add(pluralIcu!);
                         break;
@@ -93,19 +95,7 @@ internal static class InterpolationConverter
                     }
 
                     // Ensure unique param names
-                    var baseName = paramName ?? $"arg{usedNames.Count}";
-                    var uniqueName = baseName;
-                    var suffix = 2;
-                    while (!usedNames.Add(uniqueName))
-                    {
-                        uniqueName = $"{baseName}{suffix++}";
-                    }
-
-                    // Map back to original expression
-                    if (exprText != uniqueName)
-                    {
-                        argumentMap[uniqueName] = exprText;
-                    }
+                    var uniqueName = AddParameter(paramName, exprText, usedNames, argumentMap);
 
                     // Convert format specifier to ICU formatter
                     var formatClause = hole.FormatClause;
@@ -131,6 +121,7 @@ internal static class InterpolationConverter
                     {
                         icuParts.Add($"{{{uniqueName}}}");
                     }
+                    emittedHoleNames[contentIndex] = uniqueName;
                     break;
             }
         }
@@ -177,14 +168,10 @@ internal static class InterpolationConverter
 
         // Resolve a parameter name from the condition
         var (condName, condExpr, _) = AnalyzeExpression(ternary.Condition);
-        var baseName = condName ?? $"arg{usedNames.Count}";
-        var uniqueName = baseName;
-        var suffix = 2;
-        while (!usedNames.Add(uniqueName))
-            uniqueName = $"{baseName}{suffix++}";
-
-        if (condExpr != uniqueName)
-            argumentMap[uniqueName] = condExpr;
+        var uniqueName = AddParameter(condName, condExpr, usedNames, argumentMap);
+        // ICU select keys are strings. Normalize the C# boolean condition rather
+        // than passing a bool to the MessageFormat runtime.
+        argumentMap[uniqueName] = $"({condExpr}) ? \"true\" : \"false\"";
 
         var trueText = EscapeForIcu(whenTrueLit.Token.ValueText);
         var falseText = EscapeForIcu(whenFalseLit.Token.ValueText);
@@ -198,15 +185,14 @@ internal static class InterpolationConverter
         int contentIndex,
         ExpressionSyntax expr,
         List<string> icuParts,
-        HashSet<string> usedNames,
-        Dictionary<string, string> argumentMap,
-        out string? icuPlural)
+        IReadOnlyDictionary<int, string> emittedHoleNames,
+        List<string> warnings,
+        out string? icuPlural,
+        out bool suppressStandalonePlural)
     {
         icuPlural = null;
+        suppressStandalonePlural = false;
         if (!TryGetPluralTernary(expr, out var quantity, out var singularText, out var pluralText))
-            return false;
-
-        if (string.IsNullOrEmpty(singularText) == string.IsNullOrEmpty(pluralText))
             return false;
 
         var previousTextIndex = contentIndex - 1;
@@ -233,13 +219,18 @@ internal static class InterpolationConverter
         if (wordStart == trimmedLiteral.Length || icuParts.Count < 2)
             return false;
 
+        // This ternary belongs to the preceding quantity, even if folding cannot proceed.
+        // Do not subsequently emit a second standalone plural for the same quantity.
+        suppressStandalonePlural = true;
+        if (!emittedHoleNames.TryGetValue(previousQuantityIndex, out var uniqueName))
+            return false;
+
+        if (previousQuantityHole.FormatClause is not null)
+            warnings.Add($"Format specifier on quantity '{previousQuantityText}' is represented by the ICU plural number sign");
+
         icuParts.RemoveRange(icuParts.Count - 2, 2);
-        var uniqueName = usedNames.Contains(quantity.name)
-            ? quantity.name
-            : AddParameter(quantity, usedNames, argumentMap);
-        var singular = EscapeForIcu($"#{trimmedLiteral}");
-        var pluralSuffix = string.IsNullOrEmpty(pluralText) ? singularText : pluralText;
-        var plural = EscapeForIcu($"#{trimmedLiteral}{pluralSuffix}");
+        var singular = $"#{EscapeForIcuPluralBody(trimmedLiteral)}{EscapeForIcuPluralBody(singularText)}";
+        var plural = $"#{EscapeForIcuPluralBody(trimmedLiteral)}{EscapeForIcuPluralBody(pluralText)}";
         icuPlural = $"{{{uniqueName}, plural, one {{{singular}}} other {{{plural}}}}}";
         return true;
     }
@@ -255,17 +246,24 @@ internal static class InterpolationConverter
             || (string.IsNullOrEmpty(singularText) && string.IsNullOrEmpty(pluralText)))
             return false;
 
+        // A standalone full-word ternary is only demonstrably plural when the
+        // plural arm is the singular arm plus its English plural suffix. Other
+        // conditionals (for example Enabled/Disabled) remain selects.
+        if (!string.IsNullOrEmpty(singularText) && !string.IsNullOrEmpty(pluralText)
+            && !string.Equals(pluralText, singularText + "s", StringComparison.Ordinal))
+            return false;
+
         var uniqueName = AddParameter(quantity, usedNames, argumentMap);
         if (string.IsNullOrEmpty(singularText) || string.IsNullOrEmpty(pluralText))
         {
-            var singularSuffix = EscapeForIcu(singularText);
-            var pluralSuffix = EscapeForIcu(pluralText);
+            var singularSuffix = EscapeForIcuPluralBody(singularText);
+            var pluralSuffix = EscapeForIcuPluralBody(pluralText);
             icuPlural = $"{{{uniqueName}, plural, one {{{singularSuffix}}} other {{{pluralSuffix}}}}}";
             return true;
         }
 
-        var singular = EscapeForIcu($"# {singularText}");
-        var plural = EscapeForIcu($"# {pluralText}");
+        var singular = $"# {EscapeForIcuPluralBody(singularText)}";
+        var plural = $"# {EscapeForIcuPluralBody(pluralText)}";
         icuPlural = $"{{{uniqueName}, plural, one {{{singular}}} other {{{plural}}}}}";
         return true;
     }
@@ -326,7 +324,7 @@ internal static class InterpolationConverter
             return false;
 
         var (name, exprText, isComplex) = AnalyzeExpression(quantityExpression);
-        if (name is null || isComplex)
+        if (name is null || isComplex || !IsQuantityName(name))
             return false;
 
         quantity = (name, exprText);
@@ -343,16 +341,23 @@ internal static class InterpolationConverter
     private static string AddParameter(
         (string name, string exprText) parameter,
         HashSet<string> usedNames,
+        Dictionary<string, string> argumentMap) =>
+        AddParameter(parameter.name, parameter.exprText, usedNames, argumentMap);
+
+    private static string AddParameter(
+        string? parameterName,
+        string exprText,
+        HashSet<string> usedNames,
         Dictionary<string, string> argumentMap)
     {
-        var baseName = parameter.name;
+        var baseName = parameterName ?? $"arg{usedNames.Count}";
         var uniqueName = baseName;
         var suffix = 2;
         while (!usedNames.Add(uniqueName))
             uniqueName = $"{baseName}{suffix++}";
 
-        if (parameter.exprText != uniqueName)
-            argumentMap[uniqueName] = parameter.exprText;
+        if (exprText != uniqueName)
+            argumentMap[uniqueName] = exprText;
 
         return uniqueName;
     }
@@ -402,4 +407,7 @@ internal static class InterpolationConverter
         // and single quotes
         return text.Replace("'", "''").Replace("{", "'{'").Replace("}", "'}'");
     }
+
+    private static string EscapeForIcuPluralBody(string text) =>
+        EscapeForIcu(text).Replace("#", "'#'");
 }
--- a/src/Reactor.Cli/Loc/InterpolationConverter.cs
+++ b/src/Reactor.Cli/Loc/InterpolationConverter.cs
@@ -33,11 +33,10 @@ internal static class InterpolationConverter
         ["f"] = "date, full",
     };
 
-    // Variable names that suggest quantities (for plural hint comments)
+    // Variable names that identify standalone plural candidates.
     private static readonly HashSet<string> QuantityHintNames = new(StringComparer.OrdinalIgnoreCase)
     {
         "count", "total", "length", "size", "amount", "quantity",
-        "remaining",
     };
 
     public static (string? icuMessage, Dictionary<string, string>? argumentMap, List<string> warnings)
@@ -133,7 +132,7 @@ internal static class InterpolationConverter
     }
 
     /// <summary>
-    /// Checks if a parameter name suggests a quantity (for plural hint comments).
+    /// Checks if a parameter name suggests a quantity.
     /// </summary>
     public static bool IsQuantityName(string name)
     {
@@ -192,7 +191,7 @@ internal static class InterpolationConverter
     {
         icuPlural = null;
         suppressStandalonePlural = false;
-        if (!TryGetPluralTernary(expr, out var quantity, out var singularText, out var pluralText))
+        if (!TryGetPluralTernary(expr, requireQuantityName: false, out var quantity, out var singularText, out var pluralText))
             return false;
 
         var previousTextIndex = contentIndex - 1;
@@ -210,6 +209,10 @@ internal static class InterpolationConverter
         if (!string.Equals(previousQuantityText, quantity.exprText, StringComparison.Ordinal))
             return false;
 
+        // This ternary belongs to the preceding quantity, even if folding cannot proceed.
+        // Do not subsequently emit a second standalone plural for the same quantity.
+        suppressStandalonePlural = true;
+
         var literal = previousText.TextToken.ValueText;
         var trimmedLiteral = literal.TrimEnd();
         var wordStart = trimmedLiteral.Length;
@@ -219,9 +222,6 @@ internal static class InterpolationConverter
         if (wordStart == trimmedLiteral.Length || icuParts.Count < 2)
             return false;
 
-        // This ternary belongs to the preceding quantity, even if folding cannot proceed.
-        // Do not subsequently emit a second standalone plural for the same quantity.
-        suppressStandalonePlural = true;
         if (!emittedHoleNames.TryGetValue(previousQuantityIndex, out var uniqueName))
             return false;
 
@@ -242,7 +242,7 @@ internal static class InterpolationConverter
         out string? icuPlural)
     {
         icuPlural = null;
-        if (!TryGetPluralTernary(expr, out var quantity, out var singularText, out var pluralText)
+        if (!TryGetPluralTernary(expr, requireQuantityName: true, out var quantity, out var singularText, out var pluralText)
             || (string.IsNullOrEmpty(singularText) && string.IsNullOrEmpty(pluralText)))
             return false;
 
@@ -270,6 +270,7 @@ internal static class InterpolationConverter
 
     private static bool TryGetPluralTernary(
         ExpressionSyntax expr,
+        bool requireQuantityName,
         out (string name, string exprText) quantity,
         out string singularText,
         out string pluralText)
@@ -279,7 +280,7 @@ internal static class InterpolationConverter
         pluralText = string.Empty;
 
         if (expr is not ConditionalExpressionSyntax ternary
-            || !TryGetPluralCondition(ternary.Condition, out quantity, out var trueIsSingular))
+            || !TryGetPluralCondition(ternary.Condition, requireQuantityName, out quantity, out var trueIsSingular))
             return false;
 
         if (ternary.WhenTrue is not LiteralExpressionSyntax trueLiteral
@@ -304,6 +305,7 @@ internal static class InterpolationConverter
 
     private static bool TryGetPluralCondition(
         ExpressionSyntax condition,
+        bool requireQuantityName,
         out (string name, string exprText) quantity,
         out bool trueIsSingular)
     {
@@ -324,7 +326,7 @@ internal static class InterpolationConverter
             return false;
 
         var (name, exprText, isComplex) = AnalyzeExpression(quantityExpression);
-        if (name is null || isComplex || !IsQuantityName(name))
+        if (name is null || isComplex || (requireQuantityName && !IsQuantityName(name)))
             return false;
 
         quantity = (name, exprText);
--- a/src/Reactor.Cli/Loc/InterpolationConverter.cs
+++ b/src/Reactor.Cli/Loc/InterpolationConverter.cs
@@ -229,8 +229,8 @@ internal static class InterpolationConverter
             warnings.Add($"Format specifier on quantity '{previousQuantityText}' is represented by the ICU plural number sign");
 
         icuParts.RemoveRange(icuParts.Count - 2, 2);
-        var singular = $"#{EscapeForIcuPluralBody(trimmedLiteral)}{EscapeForIcuPluralBody(singularText)}";
-        var plural = $"#{EscapeForIcuPluralBody(trimmedLiteral)}{EscapeForIcuPluralBody(pluralText)}";
+        var singular = $"#{EscapeForIcuPluralBody(literal)}{EscapeForIcuPluralBody(singularText)}";
+        var plural = $"#{EscapeForIcuPluralBody(literal)}{EscapeForIcuPluralBody(pluralText)}";
         icuPlural = $"{{{uniqueName}, plural, one {{{singular}}} other {{{plural}}}}}";
         return true;
     }
--- a/src/Reactor.Cli/Loc/InterpolationConverter.cs
+++ b/src/Reactor.Cli/Loc/InterpolationConverter.cs
@@ -47,8 +47,9 @@ internal static class InterpolationConverter
         var warnings = new List<string>();
         var usedNames = new HashSet<string>();
 
-        foreach (var content in interpolated.Contents)
+        for (var contentIndex = 0; contentIndex < interpolated.Contents.Count; contentIndex++)
         {
+            var content = interpolated.Contents[contentIndex];
             switch (content)
             {
                 case InterpolatedStringTextSyntax text:
@@ -64,6 +65,18 @@ internal static class InterpolationConverter
                         expr = parens.Expression;
 
                     // Ternary with string-literal branches → ICU select
+                    if (TryConvertPluralSuffix(interpolated, contentIndex, expr, icuParts, usedNames, argumentMap, out var suffixPluralIcu))
+                    {
+                        icuParts.Add(suffixPluralIcu!);
+                        break;
+                    }
+
+                    if (TryConvertTernaryPlural(expr, usedNames, argumentMap, out var pluralIcu))
+                    {
+                        icuParts.Add(pluralIcu!);
+                        break;
+                    }
+
                     if (TryConvertTernarySelect(expr, usedNames, argumentMap, out var selectIcu))
                     {
                         icuParts.Add(selectIcu!);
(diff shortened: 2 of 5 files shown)


src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+        {
+            var singularSuffix = EscapeForIcu(singularText);
+            var pluralSuffix = EscapeForIcu(pluralText);
+            icuPlural = $"{{{uniqueName}, plural, one {{{singularSuffix}}} other {{{pluralSuffix}}}}}";
+            return true;
+        }
+
+        var singular = EscapeForIcu($"# {singularText}");
+        var plural = EscapeForIcu($"# {pluralText}");
+        icuPlural = $"{{{uniqueName}, plural, one {{{singular}}} other {{{plural}}}}}";
```
azchohfi: This is the one I'd most like to resolve before merge.

`TryGetPluralCondition` only checks `name is null || isComplex`, so *any* simple identifier or member access compared to the literal `1` qualifies — and this path now runs before `TryConvertTernarySelect`. That means a plain boolean-ish ternary gets reinterpreted as a plural, and `#` is injected into text the author never put a number in:

```csharp
$"Status: {(status == 1 ? "Enabled" : "Disabled")}"
// now: Status: {status, plural, one {# Enabled} other {# Disabled}}
// renders for status = 2:  "Status: 2 Disabled"
// before: Status: {arg0, select, true {Enabled} false {Disabled}}
```

The `#` injection also bites genuine plural intent when the branches already carry a determiner:

```csharp
$"You have {(n == 1 ? "one file" : "many files")}"
// -> "You have 3 many files"
```

Could this path be gated on something stronger — the `IsQuantityName` / `QuantityHintNames` heuristic already in this file, and/or requiring the two branches to share a common stem so they're demonstrably singular/plural forms of one noun? Anything that doesn't qualify could fall through to `TryConvertTernarySelect` as before.

One caveat I noticed while checking the fallback: `{flag, select, true {...} false {...}}` appears to throw in the Jeffijoe runtime when handed a `bool`, so the existing `select` path may have its own issue — worth a look before leaning on it as the escape hatch.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+        while (wordStart > 0 && char.IsLetter(trimmedLiteral[wordStart - 1]))
+            wordStart--;
+
+        if (wordStart == trimmedLiteral.Length || icuParts.Count < 2)
+            return false;
+
+        icuParts.RemoveRange(icuParts.Count - 2, 2);
+        var uniqueName = usedNames.Contains(quantity.name)
+            ? quantity.name
+            : AddParameter(quantity, usedNames, argumentMap);
```
azchohfi: This reuses the quantity's *base* name, but the hole that just got removed may have been emitted under a **de-duplicated** name, so the plural can end up bound to a different argument entirely.

```csharp
$"{model.Count} pending, {count} item{(count == 1 ? "" : "s")}"
```

- `{model.Count}` → `AnalyzeExpression` yields name `count` → emitted as `{count}`, `argMap["count"] = "model.Count"`
- `{count}` → base name taken → emitted as `{count2}`, `argMap["count2"] = "count"`
- the fold removes `{count2}` + `" item"`, sees `usedNames.Contains("count")` is true, and reuses `count`

Result:

```
{count} pending, {count, plural, one {# item} other {# items}}
argMap = { count = model.Count, count2 = count }
```

The plural now formats `model.Count`. `SourceRewriter` iterates the final ICU params, so `count2` is dropped from the rewritten call and the local `count` never reaches the message — with `model.Count = 5, count = 1` the app renders `5 pending, 5 items`. The orphaned `count2` map entry is the smaller half of it.

Would it work to record the unique name (and arg-map entry) actually emitted per `Contents` index, and reuse *that* on rollback — plus drop the now-dead map entry — instead of the `usedNames.Contains(...)` check?

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+        HashSet<string> usedNames,
+        Dictionary<string, string> argumentMap,
+        out string? icuPlural)
+    {
+        icuPlural = null;
+        if (!TryGetPluralTernary(expr, out var quantity, out var singularText, out var pluralText))
+            return false;
+
+        if (string.IsNullOrEmpty(singularText) == string.IsNullOrEmpty(pluralText))
+            return false;
```
azchohfi: When both branches are non-empty this correctly declines the fold — but control then falls through to `TryConvertTernaryPlural`, which happily emits a *standalone* plural after the quantity and noun have already been written out. The count ends up in the message twice:

```csharp
$"{n} file{(n == 1 ? " is" : "s are")} selected"
// -> {n} file{n2, plural, one {#  is} other {# s are}} selected
// renders for n = 2:  "2 file2 s are selected"
```

(`n2` because `AddParameter` de-dupes against the `n` already claimed by the first hole.)

This shape feels common enough to matter — `file is` / `files are`, `item was` / `items were`. Two options that both seem reasonable: extend the fold to handle both-non-empty (and both-empty) continuations so it produces a single plural phrase, or have the suffix case decline plural conversion outright when the ternary directly follows the same quantity hole, rather than letting a second `plural` block through.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+        var literal = previousText.TextToken.ValueText;
+        var trimmedLiteral = literal.TrimEnd();
+        var wordStart = trimmedLiteral.Length;
+        while (wordStart > 0 && char.IsLetter(trimmedLiteral[wordStart - 1]))
+            wordStart--;
+
+        if (wordStart == trimmedLiteral.Length || icuParts.Count < 2)
+            return false;
+
+        icuParts.RemoveRange(icuParts.Count - 2, 2);
```
azchohfi, line 231: Small one: when the fold swallows the preceding quantity hole, any format clause on that hole disappears silently.

```csharp
$"{count:N0} item{(count == 1 ? "" : "s")}"
```

The first pass emits `{count, number}` via the `FormatSpecifierMap` lookup, then `RemoveRange` deletes it and the plural is rebuilt without the formatter — so `:N0` is gone and a large count loses its grouping separators.

Since `Convert` already has a `warnings` channel that surfaces to the CLI, it'd be nice to either carry the formatter into the plural argument or warn when one gets dropped, so it isn't invisible.

azchohfi (commented):
Thanks for picking this one up! Plural extraction has been a rough edge for a while, and the retroactive fold in `TryConvertPluralSuffix` — letting the main loop emit the quantity and then rewinding two parts to wrap them in a `plural` — is a genuinely clever way to handle the `{count} item{...}` shape without a separate pre-pass. The overall structure looks right to me.

I spent a while tracing the emitted ICU forward through `SourceRewriter` and the `Jeffijoe.MessageFormat` runtime, and hit a few inputs where the result comes out wrong at runtime rather than just suboptimal. I've left inline notes on each. Three I'd like to sort out before this merges:

1. **Non-quantity ternaries now get pluralized.** `TryGetPluralCondition` accepts any simple identifier or member access compared to `1`, without consulting the `IsQuantityName` heuristic that already lives in this file. `$"Status: {(status == 1 ? "Enabled" : "Disabled")}"` becomes `Status: {status, plural, one {# Enabled} other {# Disabled}}`, which renders `Status: 2 Disabled`. Before this change it was a `select`.

2. **The fold can bind the plural to the wrong argument.** It reuses the quantity's *base* name rather than the unique name the swallowed hole was actually emitted under, so a prior hole that already claimed that name wins. Details and a repro in the inline note.

3. **A suffix ternary with two non-empty branches emits a second plural.** `$"{n} file{(n == 1 ? " is" : "s are")} selected"` ends up duplicating the count — `2 file2 s are selected`.

The rest of the inline comments are smaller: a literal `#` escaping gap, a silently-dropped format specifier, and a couple of coverage suggestions.

A few things outside the diff that will drift once this lands, if you want to fold them in:

- `docs/reference/localization-howto.md` (around line 241) still says *"The extractor produces simple `{variable}` placeholders"* and then walks through hand-editing `.resw` to add plurals. That page is hand-written — only `docs/guide/*` is compiled from `docs/_pipeline/templates/*.md.dt` — so it can be edited directly.
- `docs/specs/005-localization-design.md`: the interpolation conversion table (~778-785) has no ternary/plural row, and the workflow at ~866-879 still describes the developer upgrading `{count} ...` to a plural by hand.
- `CHANGELOG.md` `[Unreleased]` — this changes `mur loc extract` output for existing projects, so it probably wants a `Changed` line citing spec 005 / #1131.

None of this is structural, and the hard part is already done. Happy to help work through any of it if it gets fiddly.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+        var uniqueName = usedNames.Contains(quantity.name)
+            ? quantity.name
+            : AddParameter(quantity, usedNames, argumentMap);
+        var singular = EscapeForIcu($"#{trimmedLiteral}");
+        var pluralSuffix = string.IsNullOrEmpty(pluralText) ? singularText : pluralText;
+        var plural = EscapeForIcu($"#{trimmedLiteral}{pluralSuffix}");
```
azchohfi: `EscapeForIcu` handles `'`, `{` and `}`, but inside a `plural` body `#` is special too — and the literal being folded in here is arbitrary author text:

```csharp
$"{count} #item{(count == 1 ? "" : "s")}"
// -> {count, plural, one {# #item} other {# #items}}
// renders for count = 3:  "3 3items"
```

The second `#` came from the source literal and should stay literal. A plural-body variant of the escape that quotes `#` as `'#'` would cover it — applied to `trimmedLiteral` here and to `singularText` / `pluralText` in `TryConvertTernaryPlural`.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+        if (!TryGetPluralTernary(expr, out var quantity, out var singularText, out var pluralText)
+            || (string.IsNullOrEmpty(singularText) && string.IsNullOrEmpty(pluralText)))
+            return false;
+
+        var uniqueName = AddParameter(quantity, usedNames, argumentMap);
+        if (string.IsNullOrEmpty(singularText) || string.IsNullOrEmpty(pluralText))
+        {
+            var singularSuffix = EscapeForIcu(singularText);
+            var pluralSuffix = EscapeForIcu(pluralText);
+            icuPlural = $"{{{uniqueName}, plural, one {{{singularSuffix}}} other {{{pluralSuffix}}}}}";
```
azchohfi, line 261: More of a design question than a bug, so take it or leave it.

This branch emits a plural containing only the suffix, with the noun left outside the block and no `#` at all — `item{remaining, plural, one {} other {s}}`. That's fine for English, but it leaves a translator with nothing they can actually work with: they can't inflect `item`, can't move it relative to the number, and can't add the `few` / `many` categories that Slavic and Arabic locales need, because the noun isn't inside the plural argument.

Given the whole point of extraction is to hand translators something localizable, would it be better to decline conversion (or warn) when the extractor can't capture the full pluralizable phrase, rather than emitting a shape that's structurally hard to localize? Mainly flagging it because the test below pins this output, which makes it the intended contract rather than a fallback.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +193,170 @@ private static bool TryConvertTernarySelect(
...
+
+    private static string AddParameter(
+        (string name, string exprText) parameter,
+        HashSet<string> usedNames,
+        Dictionary<string, string> argumentMap)
+    {
+        var baseName = parameter.name;
+        var uniqueName = baseName;
+        var suffix = 2;
+        while (!usedNames.Add(uniqueName))
```
azchohfi, line 358: Nice to see this pulled out. Minor follow-on: the same unique-name + arg-map loop still exists in two other places — the main `Convert` loop (~line 99) and `TryConvertTernarySelect` (~line 183) — so there are three copies now. If `AddParameter` grew a small overload for the `arg{n}` fallback name, those two could route through it and all ICU argument naming would live in one spot.

tests/Reactor.Tests/Localization/InterpolationConverterTests.cs
```diff
@@ -140,6 +140,47 @@ public void Ternary_WithMemberAccessCondition_ConvertsToIcuSelect()
...
     }
 
+    [Fact]
+    public void QuantitySuffixTernary_ConvertsToIcuPlural()
+    {
+        var node = ParseInterpolation("$\"{remaining} item{(remaining == 1 ? \"\" : \"s\")} left\"");
+        var (icu, argMap, warnings) = InterpolationConverter.Convert(node);
+
+        Assert.Equal("{remaining, plural, one {# item} other {# items}} left", icu);
+        Assert.Null(argMap);
```
azchohfi, line 152: Worth noting that none of the four new tests exercises the argument-map path through the new code — this one asserts it's null and the other three don't look at it. That's exactly where the rollback in `TryConvertPluralSuffix` gets interesting, since it leaves `usedNames` and `argumentMap` entries behind for a placeholder it just deleted.

Two cases that would cover it:

```csharp
$"{model.Count} item{(model.Count == 1 ? "" : "s")}"   // expect argMap["count"] == "model.Count"
$"{model.Count} pending, {count} item{(count == 1 ? "" : "s")}"  // the collision case from my note above
```

tests/Reactor.Tests/Localization/InterpolationConverterTests.cs
```diff
@@ -140,6 +140,47 @@ public void Ternary_WithMemberAccessCondition_ConvertsToIcuSelect()
...
+    {
+        var node = ParseInterpolation("$\"{remaining == 1 ? \"item\" : \"items\"}\"");
+        var (icu, _, warnings) = InterpolationConverter.Convert(node);
+
+        Assert.Equal("{remaining, plural, one {# item} other {# items}}", icu);
+        Assert.Empty(warnings);
+    }
+
+    [Fact]
+    public void QuantitySuffixTernary_WithoutAdjacentQuantity_StillUsesIcuPlural()
```
azchohfi, line 177: Related coverage thought: the happy paths are well covered, but each early return in `TryConvertPluralSuffix` silently produces a *different* output shape, and none of them is pinned by a test right now. So a future change that loosens one of the guards wouldn't fail anything.

The ones I'd pin:

```csharp
$"{total} item{(count == 1 ? "" : "s")}"   // adjacent hole is a different variable
$"{n} - {(n == 1 ? "" : "s")}"             // preceding literal doesn't end in letters
$"{n} file{(n == 1 ? " is" : "s are")}"    // both branches non-empty (see my note on line 209)
$"{count == 1 ? "Yes" : "No"}"             // pins plural-vs-select for the ambiguous shape
```

A `[Theory]` with the expected ICU per case would probably read nicely here.

tests/Reactor.Tests/Localization/InterpolationConverterTests.cs
```diff
@@ -140,6 +140,47 @@ public void Ternary_WithMemberAccessCondition_ConvertsToIcuSelect()
...
+        var (icu, _, warnings) = InterpolationConverter.Convert(node);
+
+        Assert.Equal("{remaining, plural, one {# item} other {# items}} left", icu);
+        Assert.Empty(warnings);
+    }
+
+    [Fact]
+    public void QuantityTernary_WithFullWords_ConvertsToIcuPlural()
+    {
+        var node = ParseInterpolation("$\"{remaining == 1 ? \"item\" : \"items\"}\"");
```
azchohfi: This is one of the two tests failing in CI (`Expected: "{remaining, plural, one {# item} other {# items}}"`, `Actual: "{arg0}"`).

The ternary needs parens — an unparenthesized `:` inside an interpolation hole is parsed as the format-specifier separator, so this becomes expression `remaining == 1 ? "item"` plus format clause `"items"` rather than a conditional:

```csharp
var node = ParseInterpolation("$\"{(remaining == 1 ? \"item\" : \"items\")}\"");
```

Worth noting this means `TryConvertTernaryPlural`'s full-words branch isn't actually being exercised right now — so once the parens are in, I'd double-check the expected value rather than assume it goes green, since that's the branch that prepends `# ` to both arms.

tests/Reactor.Tests/Localization/InterpolationConverterTests.cs
```diff
@@ -140,6 +140,47 @@ public void Ternary_WithMemberAccessCondition_ConvertsToIcuSelect()
...
+        var (icu, _, warnings) = InterpolationConverter.Convert(node);
+
+        Assert.Equal("{remaining, plural, one {# item} other {# items}}", icu);
+        Assert.Empty(warnings);
+    }
+
+    [Fact]
+    public void QuantitySuffixTernary_WithoutAdjacentQuantity_StillUsesIcuPlural()
+    {
+        var node = ParseInterpolation("$\"item{remaining == 1 ? \"\" : \"s\"}\"");
```
azchohfi: Same root cause as the other failure (`Expected: "item{remaining, plural, one {} other {s}}"`, `Actual: "item{arg0}"`) — the conditional needs wrapping:

```csharp
var node = ParseInterpolation("$\"item{(remaining == 1 ? \"\" : \"s\")}\"");
```

The two new tests that pass are the two that already parenthesize, which lines up neatly.

azchohfi (commented):
Small follow-up — CI is red, and it's the two new tests rather than anything pre-existing, so I dug into why in case it saves you the round trip.

[Unit Tests](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/34105707150/job/102234375430) reports `failed: 2`:

```
QuantityTernary_WithFullWords_ConvertsToIcuPlural
  Expected: "{remaining, plural, one {# item} other {# items}}"
  Actual:   "{arg0}"

QuantitySuffixTernary_WithoutAdjacentQuantity_StillUsesIcuPlural
  Expected: "item{remaining, plural, one {} other {s}}"
  Actual:   "item{arg0}"
```

(`Merged coverage` is failing downstream of the same thing; every other job is green.)

I think this is a test-input issue rather than a bug in the converter. The two failures are exactly the two new tests whose ternary **isn't parenthesized**, and the two that pass are the two that are. Inside an interpolation hole C# treats a bare `:` as the format-specifier separator, so `{remaining == 1 ? "item" : "items"}` doesn't parse as a conditional at all — it parses as expression `remaining == 1 ? "item"` (an incomplete conditional) plus format clause `"items"`. That's the reason the language requires parens around a conditional there.

The `Actual` values are a good fingerprint for this: `AnalyzeExpression` hits its `ConditionalExpressionSyntax` arm, returns `isComplex: true`, so the name becomes `arg0`; then `"items"` misses `FormatSpecifierMap` and takes the unknown-specifier branch, which emits a bare `{arg0}` — precisely what CI printed. `Assert.Empty(warnings)` would have failed on the same line too, since that branch logs a warning.

Adding the parens should fix both:

```csharp
ParseInterpolation("$\"{(remaining == 1 ? \"item\" : \"items\")}\"");
ParseInterpolation("$\"item{(remaining == 1 ? \"\" : \"s\")}\"");
```

Every other ternary test in the file already does this (`$"Value: {(flag ? GetValue() : \"default\")}"`, and the `select` tests above), so it reads as a slip rather than anything deeper.

One knock-on worth flagging: because `QuantityTernary_WithFullWords_ConvertsToIcuPlural` never reaches the intended code path today, `TryConvertTernaryPlural`'s full-words branch currently has no passing coverage at all — which is the same branch carrying the `# Enabled` / `# Disabled` behaviour I raised in my earlier review. Once the parens go in, that test will start genuinely exercising it, so it's worth re-checking the expected value at that point rather than just getting it to green.

Also a thought for the helper itself (outside the diff, `ParseInterpolation` around line 10): it calls `CSharpSyntaxTree.ParseText` and takes the first `InterpolatedStringExpressionSyntax` without ever looking at diagnostics, so malformed C# silently yields a malformed tree and the failure surfaces as a confusing ICU mismatch instead of "your test input doesn't compile." A quick `Assert.Empty(tree.GetDiagnostics().Where(d => d.Severity == DiagnosticSeverity.Error))` in there would turn this class of mistake into an obvious message for the next person. Entirely optional, and arguably its own PR.

Thanks again for the work on this one!

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +180,188 @@ private static bool TryConvertTernarySelect(
...
+        var wordStart = trimmedLiteral.Length;
+        while (wordStart > 0 && char.IsLetter(trimmedLiteral[wordStart - 1]))
+            wordStart--;
+
+        if (wordStart == trimmedLiteral.Length || icuParts.Count < 2)
+            return false;
+
+        // This ternary belongs to the preceding quantity, even if folding cannot proceed.
+        // Do not subsequently emit a second standalone plural for the same quantity.
+        suppressStandalonePlural = true;
```
azchohfi: The comment here is exactly right, but the assignment is one guard too low.

Adjacency — "this ternary belongs to the preceding quantity hole" — is fully established by line 210's `previousQuantityText == quantity.exprText` check. But `suppressStandalonePlural` isn't set until line 224, so the `wordStart == trimmedLiteral.Length` bail at 219 returns with the flag still `false`, and `TryConvertTernaryPlural` then emits a standalone plural for a quantity that's already in the message.

That guard fires whenever the preceding literal doesn't end in a letter — including the case where it's just a space, since `TrimEnd()` reduces `" "` to `""` and `wordStart == 0 == Length`:

```csharp
$"{count} {(count == 1 ? "item" : "items")}"
// -> {count} {count2, plural, one {# item} other {# items}}
// renders for count = 3:  "3 3 items"
```

Same thing with a bracket or dash separator, e.g. `$"{count} ({(count == 1 ? "item" : "items")})"` → `3 (3 items)`.

Moving `suppressStandalonePlural = true;` to directly after line 211 should cover all of them — the flag then means "this ternary is spoken for" regardless of whether the fold itself can proceed, which is what the comment says it means. Worth a test for the bare-space separator specifically, since that's the spelling most people will reach for first.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +180,188 @@ private static bool TryConvertTernarySelect(
...
+        if (IsOneLiteral(binary.Right))
+            quantityExpression = binary.Left;
+        else if (IsOneLiteral(binary.Left))
+            quantityExpression = binary.Right;
+
+        if (quantityExpression is null)
+            return false;
+
+        var (name, exprText, isComplex) = AnalyzeExpression(quantityExpression);
+        if (name is null || isComplex || !IsQuantityName(name))
```
azchohfi: This kills the `Enabled`/`Disabled` false positive, which was the important part — thank you. One thought on where it sits, though.

Because `TryGetPluralCondition` is shared, the name gate now applies to *both* conversion paths, and they have very different amounts of evidence available:

- `TryConvertTernaryPlural` (standalone) has nothing but the variable name to go on, so the heuristic is genuinely load-bearing there.
- `TryConvertPluralSuffix` already knows the immediately-preceding hole is *the same expression*, that it's followed by a noun, and that the ternary's arms are a singular/plural pair of that noun. That's strong structural evidence on its own — the name adds little.

The cost is false negatives on names that are obviously quantities in context but aren't on the list:

```csharp
$"{items} item{(items == 1 ? "" : "s")}"      // "items" — no match
$"{n} file{(n == 1 ? "" : "s")}"              // "n" — no match
$"{selectedCount} row{(selectedCount == 1 ? "" : "s")}"  // doesn't start with "count"
```

Would it work to pass a flag through `TryGetPluralCondition` so the name check applies only on the standalone path, and let the suffix fold rely on its structure? That'd keep the false positive fixed without narrowing the common case — and would mean the `QuantityHintNames` list doesn't have to keep growing one entry at a time (`"remaining"` here, and presumably the next one after that).

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -164,14 +168,10 @@ private static bool TryConvertTernarySelect(
...
-        var suffix = 2;
-        while (!usedNames.Add(uniqueName))
-            uniqueName = $"{baseName}{suffix++}";
-
-        if (condExpr != uniqueName)
-            argumentMap[uniqueName] = condExpr;
+        var uniqueName = AddParameter(condName, condExpr, usedNames, argumentMap);
+        // ICU select keys are strings. Normalize the C# boolean condition rather
+        // than passing a bool to the MessageFormat runtime.
+        argumentMap[uniqueName] = $"({condExpr}) ? \"true\" : \"false\"";
```
azchohfi, line 173: Good catch on the bool-vs-string key mismatch — I'd flagged the runtime behaviour but hadn't expected it fixed here.

Two things worth noting since this line runs for *every* select, not just the ones this PR introduced:

1. It's what's failing CI right now — `Ternary_WithMemberAccessCondition_ConvertsToIcuSelect` still asserts the old `"settings.DarkMode"` value.
2. The overwrite is unconditional, so a select whose condition matches its parameter name now produces a non-null `argumentMap` where it previously produced none (`Convert` returns `null` only when the map is empty). If any caller keys off `argumentMap == null` as "nothing to rewrite", that path changes shape — probably fine, but worth a quick look at `SourceRewriter` to confirm.

Given this changes output for every existing select, not just plurals, it might deserve its own line in the CHANGELOG entry rather than being folded into the plural description.

azchohfi (commented):
This is a really solid turnaround — thanks for working through all of it. The `emittedHoleNames` map is exactly the right shape for the naming problem, dropping the both-non-empty bail in favour of `suppressStandalonePlural` is cleaner than what I'd suggested, and I like that each fix came with a test that actually pins the behaviour. `QuantitySuffixTernary_WithTwoNonEmptyBranches_FoldsIntoOnePlural` and `QuantitySuffixTernary_ReusesEmittedUniqueParameterName` are both good oracles.

Two things to sort out, then I think this is close.

**1. CI is red again — one leftover assertion.** [Unit Tests](https://github.com/microsoft/microsoft-ui-reactor/actions/runs/34309564417/job/102336277330) has `failed: 1`:

```
Ternary_WithMemberAccessCondition_ConvertsToIcuSelect
  Expected: "settings.DarkMode"
  Actual:   "(settings.DarkMode) ? "true" : "false""
```

The new unconditional `argumentMap[uniqueName] = $"({condExpr}) ? \"true\" : \"false\""` in `TryConvertTernarySelect` applies to every select, and this test still expects the old raw expression. You updated the two sibling select tests for the new shape; this one just got missed. Should be a one-line change to `"(settings.DarkMode) ? \"true\" : \"false\""`.

**2. The duplicate-count case is still reachable, via a different guard.** I've left an inline note on it, but short version: `suppressStandalonePlural` is set at line 224, *after* the `wordStart == trimmedLiteral.Length` check at 219 — while adjacency is actually established back at line 210. So when the separator between the quantity and the ternary is bare whitespace, the flag never gets set:

```csharp
$"{count} {(count == 1 ? "item" : "items")}"
// -> {count} {count2, plural, one {# item} other {# items}}
// renders for count = 3:  "3 3 items"
```

That's the same class of bug as before, and unfortunately through what I'd guess is the most natural way to write the pattern. Your own comment on line 222 ("This ternary belongs to the preceding quantity, even if folding cannot proceed") describes the fix precisely — it's just sitting below the guard it needs to be above.

---

A few smaller notes, none blocking:

- **The `select` normalization is a nice catch, but it's a behaviour change beyond this PR's title.** Every existing `select` extraction's argument map now emits a `? "true" : "false"` expression instead of the bare condition, which changes the rewritten source for projects that already ran extraction. Worth calling out explicitly in the CHANGELOG entry alongside the plural work, since someone re-running `mur loc extract` will see churn in files they didn't expect.
- **Gating on `IsQuantityName` trades false positives for false negatives.** It fixes `Enabled`/`Disabled` cleanly, but `$"{items} item{(items == 1 ? "" : "s")}"`, `$"{n} file{...}"` and `$"{selectedCount} row{...}"` all silently decline now — none of those match the list or its `num`/`total`/`count` prefixes. Having to add `"remaining"` to make the existing test pass is a hint the list will keep growing. See the inline note on line 327 for a suggestion that might get both.
- **No test for the new format-specifier warning** (`$"{count:N0} item{(count == 1 ? "" : "s")}"`). It's new user-visible output and cheap to pin.
- **The comments on lines 36 and 136 still say "for plural hint comments"**, but `QuantityHintNames` / `IsQuantityName` now gate actual extraction too. Worth a word since the next person will read that comment and assume it's advisory only.

Thanks again — the hard parts are all handled at this point.

azchohfi (commented):
Everything from the last round is handled, and the `requireQuantityName` flag is a nicer resolution than what I suggested — threading it through `TryGetPluralTernary` means the standalone path keeps its guard while the suffix fold gets to rely on structure, and dropping `"remaining"` back out of `QuantityHintNames` is the right consequence. `AdjacentQuantityTernary_WithoutANoun_DoesNotDuplicateTheCount` pins the exact case I was worried about, and moving `suppressStandalonePlural` above the `wordStart` guard fixes it properly rather than special-casing.

I built this commit and ran the converter tests locally, which turned up two things.

**1. `QuantitySuffixTernary_WithFormatSpecifier_WarnsAboutPluralFormatting` fails.** The product code is fine — the assertion is using the wrong overload. `Assert.Contains(string, IEnumerable<string>)` resolves to the *collection* overload, which tests for an exact element match, not a substring. The warning that's actually produced is longer than the expected string:

```
Assert.Contains() Failure: Item not found in collection
Not found:  "Format specifier on quantity 'count'"
```

I probed the converter directly to confirm the warning itself is correct:

```
ICU        {count, plural, one {# item} other {# items}}
WARN count 1
  [0] <<Format specifier on quantity 'count' is represented by the ICU plural number sign>>
```

So it's just the assertion — `Assert.Contains(warnings, w => w.Contains("Format specifier on quantity 'count'"))` would do it. Worth noting the two neighbouring tests sidestep this by indexing (`Assert.Contains("Complex expression", warnings[0])`), which picks the string overload instead.

**2. Removing the both-non-empty bail opened a word-joining case.** This one's on me — folding instead of emitting a second plural was my suggestion, and it exposed that the separator whitespace gets eaten. `trimmedLiteral` is `TrimEnd()`-ed and then concatenated directly onto the branch text, so when the space between the noun and the ternary lives in the *literal* rather than in the branch, the two words run together. Measured on this commit:

```
$"{count} unread {(count == 1 ? "message" : "messages")}"
  -> {count, plural, one {# unreadmessage} other {# unreadmessages}}

$"{count} new {(count == 1 ? "item" : "items")}"
  -> {count, plural, one {# newitem} other {# newitems}}

$"{count} selected {(count == 1 ? "row" : "rows")}"
  -> {count, plural, one {# selectedrow} other {# selectedrows}}
```

The two shapes already covered by tests both come out correct, which is what makes this easy to miss:

```
$"{count} file{(count == 1 ? " is" : "s are")} selected"
  -> {count, plural, one {# file is} other {# files are}} selected   ✓
$"{count} item{(count == 1 ? "" : "s")} left"
  -> {count, plural, one {# item} other {# items}} left              ✓
```

Details and a suggested fix inline on line 217.

---

Still outstanding from the first round, whenever you're ready for them — no rush, and happy to keep them as a follow-up PR if you'd rather land the converter work first:

- `docs/reference/localization-howto.md` (~line 241) still says the extractor emits only simple placeholders and walks through adding plurals by hand. It's hand-written, so it can be edited directly.
- `docs/specs/005-localization-design.md` — the interpolation table (~778-785) has no ternary/plural row, and the workflow at ~866-879 still describes the manual upgrade.
- `CHANGELOG.md` `[Unreleased]` — worth covering both the plural extraction and the `select` argument-map normalization, since the latter changes output for projects that never had a plural.

Nice work on this — it's come a long way.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +179,191 @@ private static bool TryConvertTernarySelect(
...
+
+        if (!emittedHoleNames.TryGetValue(previousQuantityIndex, out var uniqueName))
+            return false;
+
+        if (previousQuantityHole.FormatClause is not null)
+            warnings.Add($"Format specifier on quantity '{previousQuantityText}' is represented by the ICU plural number sign");
+
+        icuParts.RemoveRange(icuParts.Count - 2, 2);
+        var singular = $"#{EscapeForIcuPluralBody(trimmedLiteral)}{EscapeForIcuPluralBody(singularText)}";
+        var plural = $"#{EscapeForIcuPluralBody(trimmedLiteral)}{EscapeForIcuPluralBody(pluralText)}";
```
azchohfi: `trimmedLiteral` is `TrimEnd()`-ed at 217, then at 232-233 it's concatenated straight onto the branch text with no separator. That's correct when the ternary abuts the noun (`item{...}`, where there's no trailing space to lose), but when the separator space lives in the literal instead of in the branch, it gets deleted and the words join up.

Measured on this commit:

```
$"{count} unread {(count == 1 ? "message" : "messages")}"
  -> {count, plural, one {# unreadmessage} other {# unreadmessages}}
  renders for count = 3:  "3 unreadmessages"
```

Same for `$"{count} new {(...)}"` → `# newitem` and `$"{count} selected {(...)}"` → `# selectedrow`. The `{adjective} {(singular : plural)}` shape feels common enough to matter, and it's newly reachable — before the both-non-empty bail was removed it declined the fold here.

The `TrimEnd()` looks like it's really there so the `wordStart` scan at 218-220 can find the trailing noun, rather than because the output wants the literal trimmed. If that's right, keeping the two concerns separate would fix it — scan on the trimmed copy, but emit the original `literal`:

```csharp
var singular = $"#{EscapeForIcuPluralBody(literal)}{EscapeForIcuPluralBody(singularText)}";
var plural   = $"#{EscapeForIcuPluralBody(literal)}{EscapeForIcuPluralBody(pluralText)}";
```

That keeps `{count} item{...}` and `{count} file{(" is" : "s are")}` byte-identical to today (neither has trailing whitespace), while `{count} unread {...}` becomes `{count, plural, one {# unread message} other {# unread messages}}`. Worth a test with the separator in the literal, since both currently-tested shapes carry it in the branch and so can't catch this.

tests/Reactor.Tests/Localization/InterpolationConverterTests.cs
```diff
@@ -136,10 +138,128 @@ public void Ternary_WithMemberAccessCondition_ConvertsToIcuSelect()
...
     }
 
+    [Fact]
+    public void QuantitySuffixTernary_WithFormatSpecifier_WarnsAboutPluralFormatting()
+    {
+        var node = ParseInterpolation("$\"{count:N0} item{(count == 1 ? \"\" : \"s\")}\"");
+        var (icu, _, warnings) = InterpolationConverter.Convert(node);
+
+        Assert.Equal("{count, plural, one {# item} other {# items}}", icu);
+        Assert.Contains("Format specifier on quantity 'count'", warnings);
```
azchohfi: This one fails — `Assert.Contains() Failure: Item not found in collection`.

With a `string` first argument and a `List<string>` second, overload resolution picks `Assert.Contains<T>(T expected, IEnumerable<T> collection)`, which looks for an element *equal* to the expected value rather than one containing it. The real warning is longer:

```
Format specifier on quantity 'count' is represented by the ICU plural number sign
```

so the exact-match lookup misses. The warning itself is emitted correctly — I checked the converter output directly, `warnings.Count == 1` with that text — so this is purely the assertion.

Either of these works:

```csharp
Assert.Contains(warnings, w => w.Contains("Format specifier on quantity 'count'"));
// or, matching the style used at lines 108 and 271:
Assert.Contains("Format specifier on quantity 'count'", warnings[0]);
```

The second is how `Ternary_WithNonLiteralBranch_StillWarnsComplex` does it — indexing forces the `(string expected, string actual)` substring overload, which is probably why that one passes and this doesn't.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +179,191 @@ private static bool TryConvertTernarySelect(
...
+        while (previousExpression is ParenthesizedExpressionSyntax parens)
+            previousExpression = parens.Expression;
+
+        var previousQuantityText = AnalyzeExpression(previousExpression).exprText;
+        if (!string.Equals(previousQuantityText, quantity.exprText, StringComparison.Ordinal))
+            return false;
+
+        // This ternary belongs to the preceding quantity, even if folding cannot proceed.
+        // Do not subsequently emit a second standalone plural for the same quantity.
+        suppressStandalonePlural = true;
```
azchohfi, line 214: This is the fix for the duplicate-count case and it reads correctly now — the flag is set as soon as adjacency is established at 210, so every later bail still suppresses the standalone plural. Verified the case I raised:

```
$"{count} {(count == 1 ? "item" : "items")}"
  -> {count} {arg1, select, true {item} false {items}}
```

No duplicated count, and the select's argument map gets the normalized `(count == 1) ? "true" : "false"` so it works against the runtime. Good.

One thought for later, definitely not for this PR: the fallback is a `select`, so the noun ends up outside a plural argument and a translator can't add `few`/`many` for locales that need them. Given the fold already knows the quantity and both arms here, this shape could in principle produce a real plural. Fine as-is — the correctness bug is gone, which was the point.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@ Conventions for contributors:
 
 ### Changed
 
+- Localization extraction now converts recognized count-based singular/plural ternaries
+  into ICU plural messages (spec 005 Â§10.4, #1131).
+- Localization extraction normalizes boolean select arguments to the string keys expected
+  by ICU MessageFormat when rewriting source (spec 005 Â§10.4, #1131).
```
azchohfi: The section signs here are double-encoded — both lines render as `spec 005 Â§10.4` instead of `spec 005 §10.4`.

Reading the raw bytes of the committed file:

```
... 73 70 65 63 20 30 30 35 20 | C3 82 C2 A7 | 31 30 2E 34 ...
           "spec 005 "          Â    §        "10.4"
```

`C2 A7` is the correct UTF-8 for `§`; those two bytes have themselves been re-encoded as UTF-8, giving the four-byte `C3 82 C2 A7`. Looks like the text passed through a Latin-1 round-trip somewhere.

It's isolated to these two lines — the file has 42 `§` characters total and the other 40 are fine, including the convention block's own `"(spec 033 §1)"` example. So just retyping the character on both lines should do it.

Content-wise the entries themselves are good, and I'm glad you split the `select` normalization into its own line — that one changes output for projects with no plurals at all, so it deserved separate mention.

src/Reactor.Cli/Loc/InterpolationConverter.cs
```diff
@@ -180,6 +179,191 @@ private static bool TryConvertTernarySelect(
...
+
+        if (!emittedHoleNames.TryGetValue(previousQuantityIndex, out var uniqueName))
+            return false;
+
+        if (previousQuantityHole.FormatClause is not null)
+            warnings.Add($"Format specifier on quantity '{previousQuantityText}' is represented by the ICU plural number sign");
+
+        icuParts.RemoveRange(icuParts.Count - 2, 2);
+        var singular = $"#{EscapeForIcuPluralBody(literal)}{EscapeForIcuPluralBody(singularText)}";
+        var plural = $"#{EscapeForIcuPluralBody(literal)}{EscapeForIcuPluralBody(pluralText)}";
```
azchohfi, line 233: Confirmed working. Probed on this commit:

```
$"{count} unread {(count == 1 ? "message" : "messages")}"
  -> {count, plural, one {# unread message} other {# unread messages}}   (was: # unreadmessage)
$"{count} unread  {(...)}"        -> one {# unread  message}    (both spaces preserved)
$"{count} item{(...)} left"       -> one {# item} other {# items}} left   (unchanged)
$"{count} file{(" is" : "s are")}" -> one {# file is} other {# files are}  (unchanged)
$"{count} #item{(...)}"           -> one {# '#'item}            (escaping intact)
```

Keeping the `TrimEnd()` copy for the `wordStart` scan at 218-220 while emitting the original `literal` here is the right separation — the trim was only ever needed to locate the trailing noun, not to shape the output.

One case worth being aware of rather than changing: `$"{count} item {(count == 1 ? "" : "s")}"` now emits `one {# item } other {# item s}`, i.e. `"3 item s"`. That reads like a bug but it's what the original C# produces for the same input, so faithfully reproducing it is correct — and it's arguably an improvement, since the old `TrimEnd()` path would have quietly emitted `"3 items"` and changed the string's meaning during extraction. Not asking for anything here; just flagging it in case it turns up later and looks alarming out of context.

azchohfi (commented):
This all looks right to me. I built this commit and ran the converter suite locally — **31/31 passing**, including the two that were red last round.

I verified each fix rather than just reading the diff:

- **The word-joining case is fixed.** `$"{count} unread {(count == 1 ? "message" : "messages")}"` now gives `{count, plural, one {# unread message} other {# unread messages}}`. I also probed a couple of shapes that weren't in the tests, to check that emitting the raw `literal` didn't overcorrect: `$"{count} item {(count == 1 ? "" : "s")}"` → `one {# item } other {# item s}`, which looks wrong at a glance but is exactly what the C# source itself renders (`"3 item s"`), so preserving it is the faithful result. The old `TrimEnd()` version silently *changed* that string's meaning. Multi-space separators come through intact too, and the `#`-escaping and `" is"`/`"s are"` cases are byte-identical to before. Good fix — scanning on the trimmed copy while emitting the original is exactly the split I was hoping for.
- **The assertion fix is right,** and the predicate form is the better of the two options I offered since it doesn't depend on the warning landing at index 0.
- **Docs land in the right places.** `localization-howto.md` is genuinely hand-written — there's no `.dt` template for it and the pipeline doesn't emit into `docs/reference/` (75 templates present, none matching), so editing it directly is correct. I also checked the localization *guide* and its template for the same drift and neither mentions ICU at all, so there was nothing to update there.
- **The spec 005 citation checks out** — §10.4 is `String Interpolation → ICU Conversion (Key Design)`, which is precisely the section the new table row belongs in. Rewriting the `consider adding plural support` note to say when the hint still applies is a nicer outcome than just deleting it.
- Both CHANGELOG entries are there, including the `select` normalization one — thanks for splitting that out.

One small thing, and it's the only thing I'd still fix: **the `§` in both new CHANGELOG lines is double-encoded UTF-8.** The bytes are `C3 82 C2 A7` where they should be `C2 A7`, so it renders as `spec 005 Â§10.4` rather than `spec 005 §10.4`. Inline note with the details. Everything else in the file is fine — 42 correct `§` characters, and only these 2 are affected, so it looks like a one-off from however those lines got inserted rather than anything systemic.

Worth knowing: **CI hasn't actually run on this commit.** All six workflow runs for `6aa7910a` are sitting at `action_required` (awaiting maintainer approval), so the green checkmark on the PR is only `license/cla`. My local run covers the converter tests, but the full suite still needs a real CI pass before this merges.

Once the encoding nit is sorted this looks good to me. Thanks for staying with it through four rounds — the plural handling ended up in a much better place than where it started, and the doc updates are a genuinely nice addition on top of the original scope.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@ Conventions for contributors:
 
 ### Changed
 
+- Localization extraction now converts recognized count-based singular/plural ternaries
+  into ICU plural messages (spec 005 Â§10.4, #1131).
```
azchohfi: This one still has the double-encoded section sign — `C3 82 C2 A7` where it should be `C2 A7`, so it renders as `spec 005 Â§10.4`.

Line 36 below it is now correct, so the two render differently in the file even though they look the same in most diff views:

```
34:   into ICU plural messages (spec 005 Â§10.4, #1131).
36:   by ICU MessageFormat when rewriting source (spec 005 §10.4, #1131).
```

Copying the `§` from line 36 onto this line should be all it needs.

azchohfi (commented):
Almost — the second entry is fixed, but the first one still has the double-encoded `§`. Easy one to miss given the two lines look identical on screen.

Checking the raw bytes of the committed file at each `spec 005` occurrence:

```
line 34:  C3 82 C2 A7   <- still Â§
line 36:  C2 A7         <- fixed ✓
```

So the count of `Â§` sequences in the file went 2 → 1 rather than 2 → 0. Just line 34 left.

One bit of good news on the CI question from my last comment: the workflows on `6aa7910a` were approved and **everything passed** — CI, Coverage, Package, Bootstrap, Node projects, Build metrics all green. And since this commit only touches `CHANGELOG.md` (the converter and test blobs are byte-identical, `c7baf874` and `9c7923e8` on both commits), that green run covers the code as it stands. The current HEAD only shows `license/cla` because no workflows were triggered for it, but there's nothing unvalidated in it.

Fix that one character and I think you're done here.

azchohfi (commented):
That's got it — both section signs are clean now.

Verified against the raw bytes rather than the rendering, since that was the thing that made this easy to miss in the first place:

```
line 34:  C2 A7   ✓
line 36:  C2 A7   ✓
```

I also swept the whole file for any other double-encoding damage (`Â`/`Ã` lead-byte sequences) and it comes back zero — with the same probe returning 1 hit on the previous commit, so that zero is a real result and not a broken check.

Everything else is byte-identical to `6aa7910a`, which already passed the full suite — converter, tests, and both doc files all unchanged since then. CI is running on this commit now too, so it'll get its own clean validation rather than inheriting one.

From my side all the review feedback is resolved:

- plural detection no longer fires on non-quantity ternaries
- the fold binds to the right argument, with the collision case pinned by a test
- no duplicated counts, including the bare-separator shape
- literal `#` escaped, dropped format specifiers warned, word spacing preserved
- docs, spec, and changelog all updated

Thanks for working through six rounds of this — the plural handling is in a genuinely good state now, and the docs and spec updates went beyond what the original issue asked for. Nice work.

azchohfi (approved):
Approving — this is in good shape.

Full CI is green on this exact commit (25 passed, 3 skipped), and every substantive point from the review rounds is resolved and verified against the current tree rather than taken on trust:

- plural detection no longer fires on non-quantity ternaries, and the `requireQuantityName` split keeps the suffix fold working on names outside the hint list
- the fold binds to the placeholder name actually emitted, with the collision case pinned by a test
- no duplicated counts, including the bare-separator and both-non-empty shapes
- literal `#` escaped inside plural bodies, word spacing preserved, dropped format specifiers now warned
- `AddParameter` is down to a single implementation from three
- the how-to, spec 005 §10.4, and the changelog all describe the new behaviour

The test file went from 4 cases to 26, and the ones that matter are real oracles — they'd fail if the behaviour regressed, which is the part that's easy to skip.

I've resolved the review threads that were confirmations or since-fixed. One left open on purpose: the note about suffix-only ternaries leaving the noun outside the plural block, which affects locales needing `few`/`many`. That's a design trade-off rather than a defect in this PR — worth a follow-up issue if you agree, or just close it out if you'd rather keep the current shape.

One thing to be aware of on merge: the branch is 25 commits behind `main`, so CI validated against an older base. Nothing here looks likely to interact with what's landed since, and the repo doesn't require up-to-date branches, but you may want to update the branch first if you'd like the run to reflect current `main`.

Thanks for the patience across the review rounds — the extraction logic ended up noticeably more robust than the original fix, and the docs and spec updates were a nice bonus beyond what #1131 asked for.

PR #1214, opened 2026-09-14
fix: work around WinUI ItemsView recycled-anchor crashes

### LINKED ISSUE #1213: [Bug] Keyed ItemsView history replacement exposes WinUI invalid-anchor crash
*This content was largely generated by AI. AI makes mistakes.*

# Reactor ItemsView history replacement exposes WinUI's invalid bring-into-view anchor crash

## Summary

Replacing a keyed Reactor `ItemsView` history while an
`ItemsView.StartBringItemIntoView` request is pending can terminate the process
with `E_INVALIDARG` (`0x80070057`, surfaced as `0xC000027B`).

This is a Reactor-based reproduction of the native WinUI bug tracked in
[microsoft/microsoft-ui-xaml#11865](https://github.com/microsoft/microsoft-ui-xaml/issues/11865).
Unlike the native minimal reproduction, **the application never collapses a
container**. A diagnostic run showed Reactor doing that during recycling:
`ElementFactory<T>.RecycleElement -> TryParkForPool -> ParkOrphan`.

The purpose of this report is to document the confirmed Reactor/WinUI interaction
and request investigation of a safe integration behavior/regression test. It does
not establish that Reactor, rather than WinUI, must own the eventual fix.

Related application investigation:
[openclaw/openclaw-windows-node#1167](https://github.com/openclaw/openclaw-windows-node/issues/1167).
OpenClaw's application-side workaround is
[openclaw/openclaw-windows-node#1407 (fix(chat): work around WinUI session-switch crash)](https://github.com/openclaw/openclaw-windows-node/pull/1407).
That workaround avoids index-based bring navigation. It is not included in this
sample.

## Reproduction

The complete sample is included below: one `Program.cs` and a small project
file. It starts with `ReactorApp.Run<ChatRepro>()` and is
**unpackaged**, with no application XAML, custom titlebar, window sizing, icons,
assets, or packaging setup.

1. Save the two files below in an empty directory. On Windows ARM64, build and
   launch the project:

   ```powershell
   winapp run .\ReactorItemsViewRepro.csproj --arch arm64 -p Platform=ARM64 --debug-output
   ```

2. Wait for the synthetic session history and **Switch session** button to appear.
3. Click **Switch session** once.

**Actual:** the process crashes during subsequent native layout with a stowed
`E_INVALIDARG`. In the confirmed runs, the native bring call returned and the
button handler requested the new Reactor state before the crash.

**Expected:** the old pending navigation should be canceled, superseded, or
otherwise handled safely when its row is recycled. Replacing the history should
not terminate the process.

The first button activation reproduced the crash in **two out of two fresh,
uninstrumented launches of the minimal unpackaged sample**, using UI Automation
Invoke. The earlier packaged prototype also crashed on its first click in three
out of three launches, including a physical-pointer click. These are observed
success rates, not guarantees across machines.

## What the sample models

- One Reactor root and one persistent keyed `ItemsView`.
- Two immutable synthetic histories with 77 variable-height text rows each.
- Stable, unique keys within each history; different keys between sessions.
- `ItemContainer` row roots with `.WithKey(row.Key)`.
- `StackLayout`, no selection, and no item invocation.
- Non-animated, bottom-aligned `ItemsView.StartBringItemIntoView`.
- A `UseState` session change, causing Reactor to reconcile the history.

These are the relevant list shapes used by OpenClaw before its workaround.
OpenClaw embeds its root in `ReactorHostControl`; the simplified sample uses
`ReactorApp.Run` instead. The embedded host and packaging are not necessary to
trigger the bug.

**Timing reduction:** the button deliberately starts a bring request for the
current history's tail, then immediately calls the state setter for the other
history. This makes the pending-navigation/history-replacement overlap reliable.
OpenClaw reaches overlapping operations through asynchronous history updates and
queued tail requests; this sample does not reproduce its entire scheduling
pipeline.

There is no explicit `UpdateLayout`, application-written visibility change,
custom element factory, manual `ItemsSource` mutation, reflection, artificial
exception, timer, network, Markdown renderer, or OpenClaw dependency. The requested
index is 76 and the current history contains 77 items when the native call starts.

### Program.cs

```csharp
using Microsoft.UI.Reactor;
using Microsoft.UI.Reactor.Core;
using Microsoft.UI.Xaml;
using Microsoft.UI.Xaml.Controls;
using static Microsoft.UI.Reactor.Factories;
using NativeItemsView = Microsoft.UI.Xaml.Controls.ItemsView;

ReactorApp.Run<ChatRepro>();

public sealed class ChatRepro : Component
{
    private sealed record Message(string Key, string Text);

    private static readonly Message[][] Histories = Enumerable.Range(0, 2)
        .Select(session => Enumerable.Range(0, 77).Select(index => new Message(
            $"session:{session}|message:{index}",
            $"Session {session + 1}, message {index + 1}\n"
                + string.Concat(Enumerable.Repeat(
                    "Synthetic chat text. This row has a different measured height.\n",
                    index % 7 + 1)))).ToArray()).ToArray();

    private NativeItemsView? _itemsView;

    public override Element Render()
    {
        var (session, setSession) = UseState(0);
        var rows = Histories[session];
        var items = ItemsView(
            rows,
            static row => row.Key,
            (row, _) => ItemContainer(
                TextBlock(row.Text).TextWrapping(TextWrapping.Wrap))
                .WithKey(row.Key)) with
        {
            LayoutKind = ItemsViewLayoutKind.StackLayout,
            SelectionMode = ItemsViewSelectionMode.None,
            IsItemInvokedEnabled = false,
        };

        return Grid(
            [GridSize.Star()],
            [GridSize.Auto, GridSize.Star()],
            Button("Switch session", () =>
            {
                if (_itemsView is not { IsLoaded: true } view)
                    throw new InvalidOperationException("The timeline is not loaded.");

                // Put the old history's tail request in flight, then replace the
                // keyed history through Reactor. No visibility or layout mutation.
                view.StartBringItemIntoView(rows.Length - 1, new BringIntoViewOptions
                {
                    AnimationDesired = false,
                    VerticalAlignmentRatio = 1.0,
                });
                setSession(1 - session);
            }).Grid(row: 0),
            items.Set(view => _itemsView = view)
                .Grid(row: 1));
    }
}
```

### ReactorItemsViewRepro.csproj

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>WinExe</OutputType>
    <TargetFramework>net10.0-windows10.0.26100.0</TargetFramework>
    <Platforms>x64;ARM64</Platforms>
    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">win-$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant())</RuntimeIdentifier>
    <UseWinUI>true</UseWinUI>
    <WindowsPackageType>None</WindowsPackageType>
    <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.UI.Reactor" Version="0.1.0-preview.15" />
    <PackageReference Include="Microsoft.WindowsAppSDK" Version="2.4.0" />
  </ItemGroup>
</Project>
```

## Observed Reactor transition

In a diagnostic build of the earlier packaged prototype, a read-only
visibility-property observer was registered on the current tail container after
`StartBringItemIntoView` returned. It reported:

```text
REPRO: bring session=1, index=76
REPRO: target before reconcile: Visible
REPRO: requested session=2
REPRO: target visibility changed to Collapsed
```

The captured managed stack identified this path, shown in caller-to-callee order:

```text
ReactorHostControl.Render
  Reconciler.BindItemsViewErasedKeyedItems
    Reconciler.ApplyItemsViewErasedKeyedDiffOrFallback
      KeyedListDiff.Apply / ApplyCore / ApplyGeneral
        KeyedListDiff.Bailout
          ReactorListState.Reset
            ObservableCollection.OnCollectionChanged
              [WinUI recycling]
                ElementFactory<T>.RecycleElement
                  TryParkForPool
                    ParkOrphan
                      UIElement.Visibility = Collapsed
```

The disjoint session keys take the keyed-diff reset path in this reproduction.
The observer neither changed visibility nor caught the failure. It was removed
before the confirmation runs; it is not required or present in the delivered
unpackaged sample.

Relevant Reactor source, pinned to the tested package's repository commit:

- [ItemsView binding and keyed diff](https://github.com/microsoft/microsoft-ui-reactor/blob/f17ce2dd12083e7e8709883c7c4ccd04c556810d/src/Reactor/Core/Reconciler.KeyedItemsBinding.cs#L251-L318)
- [ParkOrphan](https://github.com/microsoft/microsoft-ui-reactor/blob/f17ce2dd12083e7e8709883c7c4ccd04c556810d/src/Reactor/Core/ElementFactory.cs#L756)
- [TryParkForPool](https://github.com/microsoft/microsoft-ui-reactor/blob/f17ce2dd12083e7e8709883c7c4ccd04c556810d/src/Reactor/Core/ElementFactory.cs#L856-L867)
- [RecycleElement](https://github.com/microsoft/microsoft-ui-reactor/blob/f17ce2dd12083e7e8709883c7c4ccd04c556810d/src/Reactor/Core/ElementFactory.cs#L907-L939)

The factory comments explain that parking collapsed containers prevents ghost
rows while WinUI keeps recycled elements parented. Simply deleting that collapse
is therefore not proposed as a safe fix.

## Native failure and comparison with OpenClaw

The minimal unpackaged sample dump contains `0x80070057` in its stowed
exceptions. Matching Microsoft symbol-server PDBs resolve the original throw to:

```text
ScrollingAnchorRequestedEventArgs::AnchorElement
  ScrollingAnchorRequestedEventArgs.cpp:43
  Microsoft.UI.Xaml.Controls.dll + 0x325E48

ItemsView::OnScrollViewAnchorRequested
  ItemsView.cpp:972
  Microsoft.UI.Xaml.Controls.dll + 0x360B04

ScrollPresenter::EnsureAnchorElementSelection
  ScrollPresenterAnchoring.cpp:367
  Microsoft.UI.Xaml.Controls.dll + 0x3287E0
```

The minimal unpackaged run also includes:

```text
FlowLayoutAlgorithm::Measure
  -> ViewportManagerWithPlatformFeatures::SuggestedAnchor
    -> ScrollPresenter::AnchorElement
      -> EnsureAnchorElementSelection
        -> ItemsView::OnScrollViewAnchorRequested
          -> ScrollingAnchorRequestedEventArgs::AnchorElement
```

The first **24 Controls-module frames** in the minimal unpackaged dump, compared by module-relative address,
match the original OpenClaw crash's original-throw context. The module PDB
identities also match:

| PDB | GUID plus age |
| --- | --- |
| `Microsoft.ui.xaml.pdb` | `1AA39E9A93038419472E12ABB014892F1` |
| `Microsoft.UI.Xaml.Controls.pdb` | `EF5A6F388B15033C45BBC2D6F489BE2B1` |

An earlier prototype run entered anchor selection through
`ScrollPresenter::ArrangeOverride` instead. Both reach the same anchor setter;
the minimal unpackaged run matches OpenClaw's measure-path entry. This is not
just a match on the generic top-level `0xC000027B` crash code.

## Control experiment and validation

| Experiment | Result |
| --- | --- |
| Clean/build minimal unpackaged ARM64 Debug sample | Passed, 0 warnings, 0 errors |
| Minimal unpackaged sample, two fresh launches | First button activation crashed in both; native dump confirmed the anchor setter and OpenClaw measure path |
| Earlier packaged prototype, three fresh launches | First click crashed in all three |
| Temporary visibility observer in the earlier prototype | Reactor collapsed the pending tail target, then the process crashed |
| Earlier prototype with only the native bring request omitted from the button's behavior | 20 session switches completed; all 20 selected-session transitions verified through UI Automation; process remained responsive |

The earlier prototype's no-bring experiment is evidence that keyed replacement alone was insufficient
to trigger this failure in that run. It is not a general stress-safety claim for
Reactor. There is no no-bring mode or workaround UI in the delivered sample.

The project, screenshots, diagnostic logs, and dumps were generated locally with
synthetic data. No private chat data was used. Full dumps are not needed to run
the reproduction.

## Environment

Tested September 14, 2026:

| Component | Version |
| --- | --- |
| OS | Windows 11, build `10.0.26200`, ARM64 |
| Configuration | Unpackaged Reactor application, ARM64 Debug |
| .NET SDK | `10.0.400` |
| Target framework | `net10.0-windows10.0.26100.0` |
| `Microsoft.UI.Reactor` | `0.1.0-preview.15` |
| Reactor package source commit | `f17ce2dd12083e7e8709883c7c4ccd04c556810d` |
| `Microsoft.WindowsAppSDK` | `2.4.0` |
| Native Xaml and Controls file version | `3.2.3.2608` |
| Native source revision | `aa188bfdf74608810e8e57228db6dd137f4e74e4` |
| WinApp CLI used for launch/input/crash capture | `0.6.3-prerelease.36` |

The NuGet versions are intentionally pinned to match the investigated OpenClaw
build, not selected as recommendations for new applications. The inspected
Reactor checkout was at `d9ee0d9b`; its relevant factory and ItemsView binding
files were unchanged from the package commit above.

Final `Program.cs` SHA-256:
`D8ED327C3704B56153D9791C932DD5F60D4BF3290BB8EB538D20AD95F88F31F7`.

## Scope and requested investigation

This confirms that normal Reactor recycling can expose the already-reported
WinUI invalid-anchor failure without an application explicitly hiding a row.
It does not prove an independent Reactor defect, test a newer source-built
Reactor version, establish behavior on x64/Release, or resolve the separate
sustained-hang report in OpenClaw.

Please investigate how pending index-based bring requests should interact with
Reactor's keyed history replacement and collapsed-container recycling, and add a
real-WinUI regression test for this combination. Coordinate the ownership of the
fix with microsoft/microsoft-ui-xaml#11865 rather than assuming that suppressing a
synchronous managed exception or leaving parked containers visible is sufficient.

dotMorten:
## Summary

Park recycled `ItemsView` rows by collapsing their template content, not the outer `ItemContainer`. Keep recycled rows out of native scroll anchoring.

The two parts address distinct failures:

- `ElementFactory` keeps the outer container visibility-valid and collapses its template content (or its child before templating). It captures the actual target and restores that target's original `Visibility` value source on reuse. Other root types retain their existing collapse behavior.
- `ItemsViewAnchoring` registers once per `ScrollView`, after the `ItemsView` handler, and rejects a selected `ItemContainer` when its `ItemsRepeater` index is negative. An explicit pending bring can otherwise retain an offscreen recycled row and drive a layout cycle even though the outer container remains visible.

This is the normal parking design, not the earlier opacity/disabled-state implementation. Parking does **not** mutate outer `Opacity`, opacity transitions, `IsEnabled`, `IsHitTestVisible`, or `IsTabStop`. A binding on the actual visibility target causes retirement rather than unsafe pooling. Other outer bindings remain untouched.

`ElementFactory` continues to own parking, restoration, retirement, and pool limits. The 41-line anchor helper owns only recycled-anchor rejection and weak-key registration bookkeeping. Existing `ParkedVisibility` naming and the detailed ownership/value-source documentation are retained.

## Linked issues

Fixes microsoft/microsoft-ui-reactor#1213.

Native WinUI issue: microsoft/microsoft-ui-xaml#11865.

Downstream investigation: openclaw/openclaw-windows-node#1167.

## Performance: current head versus latest main

**The large throughput/allocation penalty is not present in these measurements.** Throughput and allocations show no regression distinguishable from noise in StocksGrid or the scrolling control. Targeted parking throughput is also within noise; its working set increased slightly, approximately 1%.

These are **new local measurements of `cc97e6049433e6082036790f87999cd64352fcd4` versus `72aa9fc63ad1f35385cd384e319cae593a250c93`**, not results from an earlier implementation or a GitHub Actions run.

Windows 11 ARM64, Snapdragon X1E80100 (12 logical cores), 63 GiB RAM, .NET SDK 10.0.400. Each comparison used fresh processes, interleaved main/PR runs on the same machine, and identical native/runtime dependencies within each workload. Builds, OpenClaw, its synthetic gateway, and ETW capture were finished/stopped before the timed comparisons.

The main/PR columns below are **medians**. Delta is the **mean of paired percentage changes**, with the repository's paired 95% confidence interval; it is not the percentage difference between those medians. Intervals including zero mean no change was resolved at this sample size. Throughput is higher-is-better; all other metrics are lower-is-better.

### Repository StocksGrid benchmark

Unmodified `StressPerf.ReactorOptimized`, 50% mutation, 10 seconds/run, 12 measured pairs, two discarded warmup runs per side. Native WinUI `3.2.1.2605`. Used the repository's `Run-PerfBenchmark.ps1` and `PerfLib.ps1`, not a reimplementation.

| Metric | Main median | PR median | Paired delta [95% CI] |
| --- | ---: | ---: | ---: |
| Reconciliations/sec | 4.04 | 4.02 | -0.33% [-1.85%, +1.19%] |
| Average reconcile, ms | 36.17 | 36.68 | +1.39% [-0.01%, +2.79%] |
| Average diff, ms | 30.81 | 31.32 | +1.62% [-0.10%, +3.34%] |
| Average working set, MiB | 512.97 | 512.85 | +0.41% [-0.88%, +1.70%] |
| Allocated bytes/reconciliation | 5,136,894 | 5,136,847 | +0.32% [-0.26%, +0.91%] |
| Gen0 collections/1,000 reconciliations | 285.71 | 285.71 | +2.04% [-1.92%, +6.00%] |

### Targeted ItemsView parking: exact local-package DLLs

The same independent harness used to investigate the earlier parking regression: two disjoint-key, 500-row variable-height histories in a 900x700 window. Each row contains an `ItemContainer`, two text blocks (one wrapped), and an action button. A dispatcher timer requests the next history only after the prior reconciliation completes. The unchanged workload uses `StressPerf.Shared.PerfTracker`.

Only `Reactor.dll` was swapped between runs, using the DLLs extracted from the current-head and latest-main local NuGet packages. The executable and native WinUI `3.2.3.2608` stayed identical. Each process had a two-second warmup; one leading run per side was also discarded. Measurement windows were 10 seconds, with 12 measured pairs.

| Metric | Main median | PR median | Paired delta [95% CI] |
| --- | ---: | ---: | ---: |
| Completed reconciliations/sec | 22.95 | 22.77 | +0.57% [-3.38%, +4.53%] |
| Average working set, MiB | 179.41 | 180.83 | +1.06% [+0.11%, +2.00%] |
| Allocated bytes/reconciliation | 364,638 | 361,184 | -0.65% [-1.26%, -0.04%] |
| Gen0 collections/1,000 reconciliations | 57.45 | 58.04 | +0.76% [-1.06%, +2.58%] |

The small working-set increase is **not** classified as within noise: its interval excludes zero. This run does not establish the cause of that increase. Allocations decreased slightly; throughput did not measurably change.

### Scrolling control

Same app, packages, native runtime, and warmup policy; six measured pairs. Alternates nonanimated `ScrollTo` between top and bottom without replacing history. This is a **scroll-command rate proxy**, not scrolling FPS or completed-frame throughput.

| Metric | Main median | PR median | Paired delta [95% CI] |
| --- | ---: | ---: | ---: |
| Scroll commands/sec | 30.01 | 30.00 | -0.03% [-0.36%, +0.30%] |
| Average working set, MiB | 150.42 | 152.44 | +0.98% [-1.72%, +3.68%] |
| Allocated bytes/command | 6,034 | 6,014 | -0.01% [-0.88%, +0.86%] |
| Gen0 collections/1,000 commands | 0 | 0 | Not estimable from zero baselines |

<details>
<summary>Benchmark commands and scope</summary>

Build the repository harnesses for both checkouts first:

```powershell
dotnet build .\tests\stress_perf\StressPerf.ReactorOptimized\StressPerf.ReactorOptimized.csproj -c Release -p:Platform=ARM64 -p:PerfCiSelfContained=true -p:SkipSignaturesGen=true -m:1 -nodereuse:false
dotnet build .\tests\stress_perf\StressPerf.Direct\StressPerf.Direct.csproj -c Release -p:Platform=ARM64 -p:PerfCiSelfContained=true -m:1 -nodereuse:false
```

Then, from the PR checkout, with `$MainRoot` pointing to the clean baseline:

```powershell
.\tests\stress_perf\ci\Run-PerfBenchmark.ps1 `
  -BaselineRoot $MainRoot -SkipBuild -Platform ARM64 `
  -Reps 12 -Warmup 2 -Duration 10 -RefReps 1 -RefWarmup 1 `
  -IncludeMicro:$false -IncludeSkipFloor:$false -IncludeKeyedList:$false `
  -IncludeFlex:$false -IncludeDataGrid:$false -IncludeRowMemo:$false
```

The targeted harness is a separate local experiment, not a committed production/test change. Its existing `Compare.ps1` was run with `-Reps 12 -Seconds 10 -Modes reset` and `-Reps 6 -Seconds 10 -Modes scroll`, supplying a variant manifest pointing to the two extracted package DLLs. Every recorded pair was checked for completeness, positive work, and the expected package DLL hashes. `Summarize.ps1` uses the repository's `Get-PerfMedian` and `Get-PerfPairedDeltaStats`.

The runtime versions differ **between** workloads but never **within** an A/B comparison. Absolute rates are not compared across workloads. No presented-frame/FPS or OpenClaw interaction-latency claim is made. The original screenshot's benchmark source was not available, so this is not an exact reproduction of that screenshot's workload.

</details>

## Downstream OpenClaw NuGet validation

Built and consumed **`Microsoft.UI.Reactor 0.1.0-preview.15-local.parking.cc97e604`**, with repository commit metadata identifying the current PR head.

OpenClaw was tested in an isolated checkout at `bd9ce43b4d9c63fa90196a6a33b1c1be409850b2`, preserving its original `StartBringItemIntoView` call. No application-side scroll workaround was applied. The only final tracked change in that checkout is the core NuGet package reference. `Microsoft.UI.Reactor.Advanced` remains official `0.1.0-preview.15`; native WinUI remains `3.2.3.2608`.

1. **Before applying the new package**, rebuilt with official core `0.1.0-preview.15` and a fresh isolated profile. Stock OpenClaw again crashed during initial history loading, with process exception `0xC000027B`.
2. Changed the core package reference, restored from a scoped local feed, and rebuilt Release ARM64. Verified the **loaded** `Reactor.dll` SHA-256 equals the DLL extracted from the new package. Both native WinUI DLL hashes were unchanged from stock.
3. The same synthetic loopback gateway and equivalent fresh profile loaded successfully. Completed **20 real Sessions-page "Open in chat" switches** between two 77-message mixed-height histories. All 20 showed the replacement history's final message; all 20 had zero stale visible rows. The app remained responsive and was explicitly stopped after proof.

The gateway serves synthetic prose, tables, and code blocks only; this is real OpenClaw UI against a fixture, **not a live OpenClaw Gateway**.

```text
Current-head package SHA-256:
44F4785A71AD8D5C2B362488023930D92265173DDC6145AD414480BB03C2E74E

Packaged and loaded Reactor.dll SHA-256:
75D1D3EEE34702110B3D9D3C75FBB9FFE76E9B2610F475AD640BE3D18DB0A74F

Requested/completed switches: 20/20
Final-message-visible checks: 20/20
Stale visible old-history rows: 0
Responsive at finish: true
```

## Validation

| Current-source validation | Result |
| --- | --- |
| Reactor full Release solution build | Passed; two existing `WIN2D0001` warnings |
| Reactor complete unit suite, Release ARM64 | 13,932 passed, 64 skipped, 0 failed |
| Reactor live `ItemsView_` fixtures | 12 passed |
| Reactor live `EFR_` fixtures | 22 passed |
| Newly merged `ShutdownPolicy_` live fixture | 1 passed |
| OpenClaw Release ARM64 build with local package | Passed |
| OpenClaw required full `.\build.ps1` | Passed |
| OpenClaw Shared tests | 3,998 passed, 34 skipped, 0 failed |
| OpenClaw Tray tests | 2,976 passed, 0 failed |
| OpenClaw current-package UI proof | 20/20 switches, final messages visible, no stale rows |

The four added parking fixtures cover actual-target local/default/style visibility restoration, unchanged outer properties and bindings, target-bound retirement before/after templating, eviction ownership, and fixed/variable-height pending-bring history resets. The variable-height fixture exercises 24 cycles and positively verifies that recycled bring targets reached the native anchor callback.

```powershell
dotnet build .\Reactor.slnx --no-restore -c Release -p:SkipSignaturesGen=true -m:1 -nodereuse:false
dotnet test .\tests\Reactor.Tests --no-restore -c Release -p:Platform=ARM64 -p:SkipSignaturesGen=true --report-trx
dotnet run --project .\tests\Reactor.AppTests.Host --no-build -c Release -p:Platform=ARM64 -- --self-test --filter ItemsView_
dotnet run --project .\tests\Reactor.AppTests.Host --no-build -c Release -p:Platform=ARM64 -- --self-test --filter EFR_
dotnet run --project .\tests\Reactor.AppTests.Host --no-build -c Release -p:Platform=ARM64 -- --self-test --filter ShutdownPolicy_

# In the isolated OpenClaw checkout, with the scoped local NuGet feed:
$env:OPENCLAW_REPO_ROOT = (Get-Location).Path
.\build.ps1
dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore
dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore
```

The Reactor solution's npm TLS download failure was resolved using the exact required Copilot SDK archive with verified package integrity and the SDK's normal caches. No dependency versions or build gates were changed to bypass the failure.

A separate read-only rubber-duck review checked the draft against package/module provenance, raw paired measurements, benchmark code, metric semantics, and UI-proof output. It found no required corrections or measurement-validity blockers. This is not a successful structured autoreview run.

## Real behavior proof

The current-head live fixtures and packaged OpenClaw integration above are the closeout proof. The prior unchanged minimal reproducer completed 40 UIA switches on the pre-merge implementation; that result is historical supporting evidence, not a new run of the merged head.

Controlled earlier 2x2 experiments on both native runtime versions established that focus is **not required** for the layout cycle and `IsTabStop` suppression is **not reliable protection**. Both tab-stop states failed without anchor filtering; both passed with filtering. The weak table is subscription/lifetime bookkeeping, not focus handling.

Also used the locally built WinApp CLI `0.6.3-dotmorten-perf.37` to record and analyze ETW during the OpenClaw switching proof:

```powershell
winapp perf start -a $AppPid --output $TraceDirectory --duration-sec 180 --max-size-mib 128 --json
# Exercise the verified session-switch sequence.
winapp perf stop $CaptureId --json
winapp perf analyze $TraceDirectory --view summary --limit 3 --max-bytes 8192 --json
```

**ETW is partial, not performance-comparison proof.** The recording reached its 128 MiB size limit at approximately 29 seconds; the recorder reported terminal state `failed` with `capture_incomplete`. Analysis returned usable data with one decode error, 51 incomplete UI scopes, and nonzero/unknown loss coverage. The new stock dump's automatic stowed-exception triage was skipped by WinApp; no new symbolized native stack is claimed here. The previously established native stack remains documented in the linked issues.

## Required proof pools

None: this is a Reactor library PR, not an OpenClaw feature PR. OpenClaw serves as the downstream integration consumer; its required local build/tests and visible behavior proof are reported above.

## Risk / not verified

- No public API or dependency changes in the PR diff. Other recycled root types retain their previous behavior.
- The additional anchor handler leaves realized anchors untouched. No private native reflection, forced layout, timer-based cancellation, tab-stop manipulation, or permanent disabling of anchoring.
- This handles Reactor's own recycling path, not arbitrary application removal/collapse of pending bring targets.
- Small working-set growth in the targeted parking benchmark is reported rather than hidden. These measurements do not establish universal performance behavior.
- Variable-height tail-positioning inaccuracies seen with stock Reactor are not claimed fixed. Neither are all symptoms reported in the original OpenClaw issue.
- Full live-selftest corpus, packaged/MSIX tier, NativeAOT execution, x64 runtime, live gateway end-to-end behavior, and complete ETW coverage were not verified locally.
- Structured autoreview (`python ...\autoreview --mode local`) could not run because this host has no installed Python runtime.


haibison:
Thanks for the detailed fix and benchmarks, @dotMorten!

Out of curiosity, could we consider adding an inline comment linking upstream (e.g., `// Workaround for https://github.com/microsoft/microsoft-ui-xaml/issues/11865`) in `ItemsViewAnchoring` / `ElementFactory`? It would provide great context right in the code for anyone reading it later without having to run a `git blame`.

Also, just wondering: does the project maintain a tracking list/policy for retiring upstream workarounds once the native WinUI fix lands, or is this expected to remain indefinitely for backward compatibility across older WinAppSDK versions?

azchohfi (approved):

dotMorten:
@haibison I did have a workaround comment in the first pass of the fix, because it was truly a workaround. However the updated fix isn't as much a workaround as it is an actual fix. Even if WinUI fixes the bug, I'm fairly certain there would still be a scroll-issue left, just without the actual crash. The issue was Reactor was collapsing and item and asking WinUI to scroll to that item. That was causing a crash in WinUI, but I don't see how it would be able to consistently scroll to a collapsed item. This fix avoids that all together by collapsing just the inner content.

